Issue #356, opened 2014-11-17
Add Z-Index support to Actors


eonarheim:
To support more robust rendering, Actors should be drawn in the order of their z-index. We sill probably need a new data structure `new SortedList<T>()` to maintain the sortedness of actors if the z-index changes.

- [ ] Implement SortedList data structure that will maintain sortedness on a specified property
- [ ] Update Scene draw logic to use SortedList

PR #430, opened 2015-04-12
Z index

jedeen:
Initial implementation of z-indexing for Actors in Scenes for #356. Will eventually replace with auto-balancing version.

### DIFF
--- a/src/engine/Actor.ts
+++ b/src/engine/Actor.ts
@@ -579,9 +579,8 @@ module ex {
      * @param actor The child actor to remove
      */
     public setZIndex(newIndex: number) {
-       this.scene._sortedDrawingTree.remove(this);
        this._zIndex = newIndex;
-       this.scene._sortedDrawingTree.add(this);
+       this.scene.updateDrawTree(this);
     }
 
     /**
--- a/src/engine/Actor.ts
+++ b/src/engine/Actor.ts
@@ -564,11 +564,20 @@ module ex {
           }
        }
     }
-
+    /**
+     * Gets the z-index of an actor. The z-index determines the relative order an actor is drawn in.
+     * Actors with a higher z-index are drawn on top of actors with a lower z-index
+     */
     public getZIndex(): number {
        return this._zIndex;
     }
 
+    /**
+     * Sets the z-index of an actor and updates it in the drawing list for the scene. 
+     * The z-index determines the relative order an actor is drawn in.
+     * Actors with a higher z-index are drawn on top of actors with a lower z-index
+     * @param actor The child actor to remove
+     */
     public setZIndex(newIndex: number) {
        this.scene._sortedDrawingTree.remove(this);
        this._zIndex = newIndex;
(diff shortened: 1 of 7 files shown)


kamranayub:
So excited!

kamranayub:
Do you mind adding docs for public methods you added to actor?

kamranayub:
I think MockedElement should just be in tests, not in the sorted list file.

src/engine/Util/SortedList.ts
```diff
@@ -0,0 +1,250 @@
+﻿
+module ex {
+
+   // NOTE: this implementation is not self-balancing
```
eonarheim: @jedeen Is this still true?

src/engine/Actor.ts
```diff
@@ -562,6 +564,25 @@ module ex {
...
+    /**
+     * Sets the z-index of an actor and updates it in the drawing list for the scene. 
+     * The z-index determines the relative order an actor is drawn in.
+     * Actors with a higher z-index are drawn on top of actors with a lower z-index
+     * @param actor The child actor to remove
+     */
+    public setZIndex(newIndex: number) {
+       this.scene._sortedDrawingTree.remove(this);
+       this._zIndex = newIndex;
+       this.scene._sortedDrawingTree.add(this);
```
eonarheim: @jedeen should we add a public method off of scen to update an actor's position, so we don't need to access the _sortedDrawingTree? Something like `this.scene.updateDrawTree(actor)`?

jedeen:
@eonarheim Yes, this is just a BST implementation, so no self-balancing yet.

src/engine/Util/SortedList.ts
jedeen: Yes, this is just a BST implementation, so there's no self-balancing yet.

src/engine/Actor.ts
jedeen: Yeah, that's probably a bit cleaner.

jedeen: To do that we'd need to make the z-index property accessible from outside the Actor, and that could lead to a scenario where the z-index has been updated but the change is not reflected in the scene's sorted drawing tree.

src/engine/Util/SortedList.ts
jedeen: @eonarheim Yes, this is just a BST implementation, so no self-balancing yet.

src/engine/Actor.ts
kamranayub: @jedeen I don't think so. You only need to move the lines that modify the tree, you can leave the z-index setting inside `setZIndex` because the order of those operations doesn't appear to affect anything. It could have just as easily been:

```
this._zIndex = newIndex;
this.scene._sortedDrawingTree.remove(this);
this.scene._sortedDrawingTree.add(this);
```

As long as setting the z-index comes before modifying the tree, you're golden.

eonarheim:
@kamranayub @jedeen @alanag13 I really want to use this. If there are no objections I'm going to merge this.

kamranayub:
I'm all good :+1:

PR #434, opened 2015-04-29
Fixes #433 actors not being cleaned up in draw tree

### LINKED ISSUE #433: updating actor z-index does not clean up old actor
public setZIndex(newIndex: number) {
        this._zIndex = newIndex;
        this.scene.updateDrawTree(this);
    }

Because `_zIndex` is set before the actor is removed from the `sortedDrawingTree`, the actor still exists at its previous index value and is never cleaned up.

jedeen:
I believe this is the simplest way to resolve this issue for now. We'll revisit this when the self-balancing implementation is built.

### DIFF
--- a/src/engine/Actor.ts
+++ b/src/engine/Actor.ts
@@ -578,7 +578,8 @@ module ex {
      * Actors with a higher z-index are drawn on top of actors with a lower z-index
      * @param actor The child actor to remove
      */
-    public setZIndex(newIndex: number) {
+     public setZIndex(newIndex: number) {
+       this.scene.cleanupDrawTree(this);
        this._zIndex = newIndex;
        this.scene.updateDrawTree(this);
     }
--- a/src/engine/Scene.ts
+++ b/src/engine/Scene.ts
@@ -564,10 +564,16 @@ module ex {
       }
 
       /**
+       * Removes the given actor from the sorted drawing tree
+       */
+      public cleanupDrawTree(actor: ex.Actor) {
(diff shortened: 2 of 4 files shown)


kamranayub:
So when you call remove(actor) it doesn't remove it by reference?

jedeen:
The actor is removed by reference, but it only looks for that reference in the array with the z-index value that the actor has; otherwise it would have to inspect the entire tree for every removal.

eonarheim:
Will it scan the tree if it can't find the actor at the current z-index list? I would think it would so that remove behaves consistently and guarantees removal of the actor, even if it is not as performant.

kamranayub:
Isn't native array index of performant? Maybe its not, I guess it would
never be as fast as removing by known index. Still, the API is inconsistent
if it says remove and you pass an actor and it doesn't use the actor as
reference. Instead if it removes by zindex it should be `removeByZindex`.
Know what I mean? If you pass in an actor, it should indexOf the actor, not
use its zindex as that is not what is assumed when passing an object to a
`remove` function.

jedeen:
@kamranayub I suppose it would be better named as `removeByComparable()` in its current form.

@eonarheim It doesn't scan the tree. In this case it would end up scanning every time, because the Actor will never be at its original index. I understand the sentiment, though; it's something we can add to the upgraded implementation.

kamranayub:
Yes, I think that makes it clearer (and clear why the old code didn't work)

eonarheim:
@jedeen @kamranayub I'm down to merge if you guys are

PR #436, opened 2015-05-17
Fix #277 cameras are now initialized on scene activate

### LINKED ISSUE #277: Cameras should exist off of the scene not the engine
This has been causing trouble when transitioning to new scenes, because the camera is now offset by a large amount

eonarheim:
This should center cameras on scenes now


eonarheim:
@kamranayub @jedeen Are we happy with this change?

jedeen:
I think so.

kamranayub:
+1

PR #440, opened 2015-05-23
Fix culling

### LINKED ISSUE #287: Render distance needs to be configurable
Actors will disappear when bound boxes are no longer visible. This may not be desired behavior, especially if the drawings in the actor are larger than the bounding box.

jedeen:
(closes #287) Rather than being configurable, `actor.isOffScreen` will only be true if both the actor and its `currentDrawing` are outside of the viewport.

You can see the demo of this behavior at /tests/culling/culling.html


eonarheim:
I think there is a case where the sprite should technically be on screen.

image

If you move the corners just offscreen the sprite is culled, when it should still be visible. But since we have those calculated points, building a composite bounding box would be easy. 

```js
// pseudo-code
var bb = new ex.BoundingBox(Math.min(xCoordsPoint), Math.min(yCoordsPoints), Math.max(xCoordsPoints), Math.max(yCoordsPoints));
// need to calc edges in world coords
if(bb.left > rightEdge || bb.top > bottomEdge || bb.bottom < topEdge || bb.right < leftEdge){
  // offscreen
}

```

That should effectively do this and solve the issue I think

image

eonarheim:
+1 :)

PR #478, opened 2015-07-01
Fix #364 with test

### LINKED ISSUE #364: Setting opacity when actor uses only color doesn't do anything
If I set the opacity in an update loop when an actor is only using a color, their opacity isn't changed.

```
update(...) {
  this.opacity = 0.5;
}
```

eonarheim:
For #364

### DIFF
--- a/src/spec/ActorSpec.ts
+++ b/src/spec/ActorSpec.ts
@@ -919,11 +919,19 @@ describe("A game actor", () => {
 	
 	it('changes opacity on color', () => {
 		actor.color = ex.Color.Black.clone();
+		expect(actor.color.a).toBe(1);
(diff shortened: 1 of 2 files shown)


src/spec/ActorSpec.ts
```diff
@@ -916,6 +916,15 @@ describe("A game actor", () => {
...
 
    });
+	
+	it('changes opacity on color', () => {
+		actor.color = ex.Color.Black.clone();
+		expect(actor.opacity).toBe(1.0);
+		actor.opacity = .5;
+		
+		actor.update(engine, 100);
+		expect(actor.color.a).toBe(.5);
```
kamranayub: I would also expect the other channels to remain the same.

PR #503, opened 2015-08-19
Close #386, Critical Feature detection

### LINKED ISSUE #386: Detect key browser features and fail gracefully on unsupported browsers
There is a bare minimum set of browser features required for Excalibur to function. I propose we do simple feature detection and fail if these critical features don't exist failing with an friendly message.

Examples of core features I can think of right now:

* [x] 2D Canvas Support
* [x] Some features of XHR 
* [x] Blob object support from XHR `request.responseType = 'blob';`
* [x] Object urls `URL.createObjectURL`
* [x] ~~Audio Tag (This may not be critical if we consider do not consider audio critical)~~
* [x] RGBA Support

eonarheim:
This is what it looks like if all the tests fail.
image
Let me know if we should update this

### DIFF
--- a/.settings/settings.json
+++ b/.settings/settings.json
@@ -1,4 +1,5 @@
 // Place your settings in this file to overwrite default and user settings.
 {
-	"editor.tabSize": 3
+	"editor.tabSize": 3,
+	"editor.insertSpaces": true
 }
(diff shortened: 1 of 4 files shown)


eonarheim:
Updated, new messages look like this
image

kamranayub:
Do you think we can provide more details in the docs (and maybe error should include a link to the page).

PR #508, opened 2015-08-22
Fix #507 since effective width is now calc'd in sprites this is causi…

### LINKED ISSUE #507: Centered sprite drawing on actors is broken

eonarheim:
We were having problems because actor was calculating the effective width and height to do sprite centering, but now sprites do that calculation.


eonarheim:
@kamranayub @jedeen Take a look

jedeen:
:+1:

PR #525, opened 2015-09-30
fix #506 animation freezeframe never set

### LINKED ISSUE #506: freezeframe is never set to last animation frame
door_animation

After the animation plays, the current drawing for that actor disappears. The same behavior occurs for `getAnimationForAll()`,  `getAnimationBetween`, and `getAnimationByIndices()`. In fact, it does not appear that `freezeframe` is ever set other than when it is initialized as '-1'. According to the documentation, the default `freezeframe` for an animation should be the last frame of the animation.

The current workaround is to manually set the `freezeframe` for each animation.

jedeen:


src/engine/Drawing/Animation.ts
```diff
@@ -118,7 +118,10 @@ module ex {
             
             this.naturalWidth = images[0] ? images[0].naturalWidth : 0;
             this.naturalHeight = images[0] ? images[0].naturalHeight : 0;
+
+            this.freezeFrame = images.length - 1;
```
eonarheim: @jedeen Do we also want to reset this if this.sprites ever changes length? Adding more sprites to an animation would break this, although I can't remember ever doing something like that. It would be an edge case... thoughts?

jedeen: @eonarheim We should probably have a broader discussion about being able to edit the Sprites on an Animation directly, and the potential pitfall that could become. I'll leave it alone for now.

PR #576, opened 2016-03-12
Fix rotation API to re-include rotation types

### LINKED ISSUE #575: Actor.actions.rotateTo and rotateBy are missing RotationType
Niether of these actions has the `RotationType` parameter that is present on the original (deprecated) `Actor.rotateTo` and `rotateBy`.

original API
image

new API
image

This issue is present in v0.6.0.30-unstable
See also #564

jedeen:
When the rotation api for Actors was changed from Actor.rotateTo /
Actor.rotateBy to Actor.actions.rotateTo / Actor.actions.rotateBy,
the rotation type was forgotten on the new method signatures.

closes #575


eonarheim:
:shipit:

PR #593, opened 2016-05-11
Refactor resources to allow setting data to bypass loading

### LINKED ISSUE #574: Refactor Sound/Texture resources so we can process data
For excalibur-pack, I need a way to load the data from in-memory vs. XHR request, for example:

```js
var sound = new ex.Sound("path/to/sound.mp3");

sound.processDownload(rawData);
```

So I don't pass these to the loader, I call them directly.

kamranayub:
Closes #574 

This change is to help support excalibur-pack by allowing loadables to have data set *before* loading so they resolve the loading promise immediately without performing an XHR request. This also lets users create resources, set data, and use them immediately without requiring a loader.

Tested with the sandbox (audio, textures) and with generic resources, no issues identified.


kamranayub:
I could potentially write some tests to prove immediate loading works without XHR.

src/engine/Resource.ts
```diff
@@ -45,14 +45,16 @@ module ex {
...
+          super();
+      }
 
       /**
        * Returns true if the Resource is completely loaded and is ready
        * to be drawn.
        */
       public isLoaded(): boolean {
-         return !!this.data;
+         return this.data !== null;
```
eonarheim: I agree, `!!this.data` is not clear

src/engine/Sound.ts
```diff
@@ -205,11 +242,10 @@ module ex.Internal {
...
 
       private _logger: Logger = Logger.getInstance();
+      private _data: any;
 
-
-
-      constructor(soundPath: string, volume?: number) {
-         this._path = soundPath;
+      constructor(public path: string, volume?: number) {
+         this._path = path;
```
eonarheim: If `this.path` is public, should we get rid of `this._path`?

src/engine/Resource.ts
kamranayub: More than that, I loaded an empty text file so the value was `""` so this evaluated to false. Whoops! Comparing to null maybe even sometimes could cause an issue if null is a valid value to load...

src/engine/Sound.ts
```diff
@@ -227,6 +263,11 @@ module ex.Internal {
       public onerror: (e: any) => void = () => { return; };
 
       public load() {
+         // Exit early if we already have data
+         if (!!this._data) {
```
eonarheim: Should we be consistent with the above code and go with `this._data !== null` instead?

src/engine/Resource.ts
eonarheim: Ah, that is fair enough

src/engine/Sound.ts
kamranayub: I considered it--sound would never evaluate to false here due to binary/uint data but it's still probably good.

kamranayub:
This is one where we may want to squash the commits into a pretty commit :+1:

eonarheim:
@kamranayub Agreed pretty commits are good, I'm okay with this being merged once it's squashed

kamranayub:
You can squash it when merging the pull request. :+1:

Issue #586, opened 2016-04-18
Actors that have been killed but re-added to Excalibur don't behave correctly


eonarheim:
Currently when actors are re-aded to excalibur after being killed, they don't behave correctly. Notably they don't seem to receive pointer events.

PR #599, opened 2016-05-25
Add PauseAfterLoader to allow pausing before starting a game

### LINKED ISSUE #262: Add an optional iOS "Tap to Play" interstitial button to Loader to accommodate interaction requirements for playing sounds
Add an interstitial step to Loader that is optional (and extensible) to have user interact with a UI element to allow iOS to play sounds immediately after loading.

---

This is for the HTML5 Audio Fallback.

Created a test case here: http://kamranicus.com/wp81-ie11-audio-bug/

Issues:
- iOS will not play audio unless there's user interaction
- Mobile devices have a long initial load time causing a significant delay in playback

Related posts:
- http://codetheory.in/html5-audio-behaviour-and-support-in-ios-and-android/
- http://codetheory.in/fixing-html5-audio-problems-in-ios-and-android-mobile-browsers-to-overcome-the-limitations/
- http://codetheory.in/solve-your-game-audio-problems-on-ios-and-android-with-web-audio-api/
- https://developer.apple.com/library/safari/documentation/AudioVideo/Conceptual/Using_HTML5_Audio_Video/Device-SpecificConsiderations/Device-SpecificConsiderations.html

I think the plan will be to implement the app cache storage with the audio tag fallback for browers that support it. Then in WinJS environment, use Windows Runtime API for sound.

kamranayub:
Closes #262 

Proposed Changes:
- Add new `PausedAfterLoader` loader with "Tap to Play" button
- Refactor Engine and Loader to separate responsibilities of drawing/updating loader
- Fix Engine to use `load()` promise vs. `oncomplete` callback so that Engine respects promise chains
- Move resources into Resources folder


kamranayub:
I need to figure out a way to load the iOS test on iOS to verify it fixes the iOS sound issue.

sandbox/web/tests/loader/pauseafter/ios.ts
```diff
@@ -0,0 +1,30 @@
...
+
+var game = new ex.Engine({
+   canvasElementId: 'game',
+   displayMode: ex.DisplayMode.FullScreen
+});
+
+var jump = new ex.Sound('../../../sounds/jump.mp3', '../../../sounds/jump.wav'); 
+
+// if iOS, use PauseAfterLoader
+var iOS = /iPad|iPhone|iPod/.test(navigator.userAgent) && !(<any>window).MSStream;
```
eonarheim: Should we add an issue for a utility to test platforms/features?

src/engine/Engine.ts
```diff
@@ -1045,19 +1049,20 @@ module ex {
       /**
        * Starts the internal game loop for Excalibur after loading
        * any provided assets. 
-       * @param loader  Optional resources to load before starting the main loop. Some [[ILoadable]] such as a [[Loader]] collection, 
-       * [[Sound]], or [[Texture]].
+       * @param loader  Optional [[ILoader]] to use to load resources. The default loader is [[Loader]], override to provide your own
+       * custom loader.
        */
-      public start(loader?: ILoadable) : Promise<any> {
+      public start(loader?: ILoader) : Promise<any> {
```
kamranayub: This is technically a breaking API change. But it didn't make sense in order to separate drawing/update a Loader. Potentially, one way to allow the previous implementation is to accept an overload which still accepts `ILoadable` and then create the default `Loader`.

src/engine/Interfaces/ILoader.ts
```diff
@@ -0,0 +1,9 @@
+/// <reference path="ILoadable.ts" />
+
+module ex {
+   
+   export interface ILoader extends ILoadable {
```
eonarheim: :+1:

sandbox/web/tests/loader/pauseafter/ios.ts
kamranayub: I could be persuaded to put this into a utilities class--but I'd prefer we not have to maintain it since we'd need to then ensure it works for new browsers/iOS releases. Instead I opted to provide an example in the documentation and reference the [StackOverflow thread](http://stackoverflow.com/questions/9038625/detect-if-device-is-ios).

src/engine/Engine.ts
```diff
@@ -1192,18 +1143,12 @@ module ex {
 
          this._isLoading = true;
 
-         loader.onprogress = (e) => {
-            this._progress = <number>e.loaded;
-            this._total = <number>e.total;
-            this._logger.debug('Loading ' + (100 * this._progress / this._total).toFixed(0));
-         };
-         loader.oncomplete = () => {
+         loader.load().then(() => {
```
kamranayub: I think this makes more sense to allow chaining promises.

src/engine/Engine.ts
```diff
@@ -1192,18 +1143,12 @@ module ex {
...
             setTimeout(() => {
                this._isLoading = false;
                complete.resolve();
             }, 500);
```
kamranayub: Any specific reason to have a 500ms timeout?

src/engine/Engine.ts
```diff
@@ -971,6 +972,11 @@ module ex {
       private _update(delta: number) {
          if (this._isLoading) {
             // suspend updates untill loading is finished
+            this._loader.update(this, delta);
+            // Update input listeners
+            this.input.keyboard.update(delta);
```
kamranayub: This is ugly but because Loader is not part of a Scene (hint hint), the input never gets cleared and results in bad input handling for the tap to play trigger. I didn't want to move the lines below above because that could cause issues (frame input gets cleared before actors can use it).

src/engine/Resources/Texture.ts
```diff
@@ -0,0 +1,117 @@
...
+    *
+    *   var player = new ex.Actor();
+    *
+    *   player.addDrawing(txPlayer);
+    *
+    *   game.add(player);
+    * });
+    * ```
+    */
+   export class Texture extends Resource<HTMLImageElement> {
```
eonarheim: :+1:

src/engine/Engine.ts
eonarheim: Ugh, yeah there really isn't a good solution for this right now. Loaders weren't designed to support all the ex functionality. Now that we are leveraging it it makes sense to maybe make Loaders part of a Scene.

eonarheim:
I'm okay with merging this :shipit:

eonarheim:
Actually, adding this loader caused problems for Hexshaper. For some reason after switching to the PauseAfterLoad causes the game to break on restart.

eonarheim:
There were some unmerged changes from the ludum-35 branch that were causing the issues, not this branch. @kamranayub merged master back once the unkill bug was fixed #586

kamranayub:
OK, latest changes are tested via Hexshaper on iOS and it fixes the issue.

PR #602, opened 2016-05-26
Fix #586 actors are no longer killed when added back to the game

### LINKED ISSUE #586: Actors that have been killed but re-added to Excalibur don't behave correctly

eonarheim:
Closes #586 

Actors that were killed and re-added to the game are now correctly unkilled.


src/engine/Scene.ts
```diff
@@ -378,6 +378,9 @@ module ex {
        */
       public add(uiActor: UIActor): void;
       public add(entity: any): void {
+         if(entity instanceof Actor) {
+            (<any>entity)._isKilled = false;
```
kamranayub: Let's add a `resurrect()` method to Actor and call that baby. :+1:

kamranayub:
:100:

PR #613, opened 2016-07-12
[Closes #511] Fix setting initial opacity to 0 and subsequent opacity updates fail

### LINKED ISSUE #511: Setting initial opacity to 0 and then trying to change to 1 fails
Reproduce:

```
var defendIntro = new ex.Actor(game.width/2, game.height/2, 998, 412);
   defendIntro.addDrawing(Resources.TextureTextDefend);
   defendIntro.opacity = 0;   
   game.add(defendIntro);
   defendIntro.delay(1000).callMethod(() => {
      defendIntro.opacity = 1;
   }).delay(2000).callMethod(() => {
      defendIntro.kill();
   });  
```

to workaround, also set previousOpacity to 0. `previousOpacity` should be initialized to something like `-1` and then on initialize, if `previousOpacity === -1 { previousOpacity === opacity }`

or we should switch to `setOpacity` which can encapsulate initialization logic.

kamranayub:
Closes #511

Proposed Changes:
- Updated Actor to update opacity values in update loop vs. draw.
- Added private opacityFx variable to hold current opacity effect.
- Added _framesDirty flag to mark when drawings are updated to re-apply opacity effect
- Update loop now ensures each drawing frame has the correct opacity effect applied
- Added opacity visual test


sandbox/web/tests/opacity/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
eonarheim: Add this to the toplevel sandbox index.html

src/engine/Actor.ts
```diff
@@ -1131,20 +1134,40 @@ module ex {
           this._isInitialized = true;
        }
        this.emit('preupdate', new PreUpdateEvent(engine, delta, this));
-       
-       var eventDispatcher = this.eventDispatcher;
```
eonarheim: :+1:

PR #617, opened 2016-08-04
[Closes #616] Child actors should be updated

### LINKED ISSUE #616: Regression: Children are not being updated by their parent actors
#### Steps to Reproduce

1. Create an Actor
2. Create another Actor
3. Add the second actor to the first
4. Add logic to the second actor's update method
5. Verify when game starts that the logic in the second actor's update is never called

#### Expected Result

Children actors should participate in the update loop. This is a regression.

#### Actual Result

Child actor's update is never called

#### Proposed Changes

Add this to the end of `Actor.update`:

```ts
// Update child actors
for (var i = 0; i < this.children.length; i++) {          
  this.children[i].update(engine, delta);          
}
```

kamranayub:
Closes #616

Proposed Changes:
- Update children in actor update loop
- Add test coverage that verifies this


eonarheim:
Recompile dists quick

PR #619, opened 2016-08-04
[Closes #618] Fix Actor anchoring and scaling draw

### LINKED ISSUE #618: Center anchored actors are not drawn at correct canvas coords when scaled
#### Steps to Reproduce

1. Create an Actor (0, 0, 10, 10)
2. Scale the Actor by 2x
3. Set `Actor.anchor` to 0.5
4. Add Actor to game

#### Expected Result

When drawn, Actor should be twice as big (20x20) centered on (0, 0)

#### Actual Result

Actor is drawn at proper scale but off-center as shown here:

image

The Green square is a 2x actor centered on (0, 0) and should perfectly contain the red and blue squares.

#### Environment

- browser: Chrome
- browser version: 51
- operating system: Windows 10 x64

kamranayub:
Closes #618 

I added a new visual test for testing anchoring with `fillRect` and Sprite drawing. It's a mess in the current version, as you can see:

Yuck

The anchor is being double-scaled and sprite drawing is super weird even with `setCenterDrawing` to false for everything except the center test.

After making some changes to the draw call logic, I believe I've fixed it properly:

image

I removed all the center drawing cruft, since now it will properly draw anchored correctly.

### Proposed Changes:

- Simplify `Actor.draw` call logic to avoid double scaling
- Remove need for `setCenterDrawing` logic


kamranayub:
There's still some work left for children/sprites.

eonarheim:
I think I've figured this out, let me explain.

The problem: Sprite centering only worked when the sprites were the exact dimensions of the actors bounding box. This is because the actors anchor is set to the middle (.5, .5) which is a good default I believe, so the default drawing are shifted actor.width * .5 and actor.height * .5 and because sprites draw from top left by default this works great. However, if these dimensions don't line up exactly, either the sprite is smaller or larger than the bounding box of the actor, the offsets are not correct.

The solution: If we want drawings to be centered, we need to account for this by developing an additional offset from the actor anchor itself. The formula below I believe accomplishes this in all cases.

The idea is that we want the sprite offset (the relative position in the drawing for example (.5, .5) is the center) to be translated to the center of the actor.

image

eonarheim:
Having a spriteOffset of (.5, .5) has a natural consequence of centering sprite scaling around the center of the actor, so if you want a top left to be the center of scaling you must also set spriteOffset to (0, 0) as seen here:

image

This begs the question, should spriteOffset and anchor always be the same thing?

eonarheim:
@excaliburjs/core-contributers It feels like the answer should be yes, spriteOffset and actor.anchor should always be the same thing

The sandbox platformer looks and behaves like it should this way
image

eonarheim:
@kamranayub spotted an issue in the coordinate test, child actors were not centering on the grid based on their anchors

image

The fix in 62ff021 was to wrap the anchor translations in a `ctx.save()` and `ctx.restore()` to localize the transforms due to anchoring, and let the other transforms propagate to the children
image

eonarheim:
@excaliburjs/core-contributers if we are in agreement this sucker can be merged!

kamranayub:
I'm so ready to merge this, let's do it!

PR #620, opened 2016-08-04
[Closes #60] Streamline and fix Actor world coordinates

### LINKED ISSUE #60: Need a better way to keep track of local and global coordinates for actors
When actors are rotated, translated, or scaled, it becomes difficult to calculate the local/or global coordinates. The engine should reason about these for us, so that we do not need to apply trigonometry to figure out the coordinates.

kamranayub:
Closes #60

This is a change that should fix problems we've been having with children and grandchildren drawing correctly but reporting incorrect world coordinates.

Due to this fix, I also uncovered #618 and fixed that issue as well (see #619)

Here is the visual test that confirms that children and grandchildren can be scaled + rotated and report correct world coords:

image

### Proposed Changes:

- Add `getWorldPos` method off Actor
- Take into account Actor graph when calculating scale and rotation
- Add unit and visual tests to confirm reality


kamranayub:
This is ready to merge when you are.

eonarheim:
I added the visual test to the list of visual tests

PR #631, opened 2016-08-19
Remove Binding and Template classes

### LINKED ISSUE #627: Remove ex.Template and ex.Binding for 0.7
#### Context
This was a poor attempt to make a UI 2-way data binding helper, mostly old crufty, slow, and untested. It would be better if we recommend other established UI frameworks instead of this.

#### Proposal
Delete it

kamranayub:
Closes #627

Proposed Changes:
- Remove `Binding`
- Remove `Template`


eonarheim:
:shipit:

PR #640, opened 2016-08-31
[Closes #512] Fix fading in between 0 and 1 and vice versa

### LINKED ISSUE #512: Fade is broken
## Context

The constructor of the `fade` action was determining what direction to go when updating opacity--however, actions are not instantiated at the time they are executed, they are instantiated immediately before any action is executed, so both `fade` actions had the same multiplier.

## Reproduction

```
actor.addDrawing(foo);
actor.opacity = 0;
actor.fade(1, 200).delay(2000).fade(0, 200);
```

## Expected Result

Actor fades in, then fades back out after 2s.

## Actual Result

Actor never fades out.

kamranayub:
Closes #512

## Context

The constructor of the `fade` action was determining what direction to go when updating opacity--however, actions are not instantiated at the time they are executed, they are instantiated immediately before any action is executed, so both `fade` actions had the same multiplier.

## Proposed Changes:

- Fix multiplier logic in Fade action
- Add visual test
- Add unit tests
- Split other Action tests into separate file (tech debt)


eonarheim:
Looks good, mergin' :shipit:

Issue #380, opened 2014-12-10
Support grouping of audio resources


alanag13:
An associative array could store groups of related sounds. You could then apply existing audio methods to these groups to have those occur on all the soundbytes in the group.
 Possible use cases i've thought of:

Separating music from sound effects and allowing developer to easily mute/adjust volume of all audio in each group

Allowing a simple abstraction for playing several soundbytes at the same time

eonarheim:
See #590

PR #644, opened 2016-09-01
[Closes #515] Refactor sound resources to address inconsistent behaviors

### LINKED ISSUE #515: replaying a long sound will not be muted if game is not visible
When playing a sound with a long duration, if `sound.play()` is called more than once, each successive play of the sound (after the first one) ignores game visibility and does not mute.

kamranayub:
Closes #515

## Context

Before this change, there were multiple subtle issues with how sounds were being paused and played. Both the `AudioTag` and `WebAudio` resources treated simultaneous playing and pausing differently--they weren't consistent with each other. In addition, they both inconsistently fetched the resource via XHR and handled setting data differently as well as volume (web audio was global volume and audio tags could be individual).

I basically standardized a multi-track based approach for both resources, where each "play" track keeps track of its own state. I also refactored enough where we can now share all the base controller logic for an audio resource, delegating any implementation-specific logic to new `AudioTagTrack` and `WebAudioTrack` classes.

## Tasks

- [x] Test sandbox sounds
- [x] Add more proper debug logging
- [x] I think `Resources\Sound.ts` and my new `AbstractAudioResource` are essentially the same, I might be able to consolidate that (or at least fix the weird loading logic in `AbstractAudioResource`)
- [x] Write tests against `AbstractAudioResource` class (create test mock)

## Looking for feedback:

- **(Done)** Does `Track` make sense or would `Instance` make more sense? Or a different word? They essentially represent *individual* plays of a sound. But I don't want it to get confused with real multi-tracking similar to #380

## Proposed Changes:

- Create consistent implementation that handles behavior logic for both audio tags and web audio
- Behavior changes:
  - Calling `play()` when *paused* will resume all unfinished tracks and will not trigger a new track to be played. 
  - Calling `play()` when *not paused* will ensure current tracks play and will trigger a new track to be played.
- Both audio tag and web audio tracks *can* support individual volume per play (but that is not supported in the API interface yet)
- Updated type annotations to take advantage of new `AudioContext` support in TS 1.5+
- Fix audio tag pausing/resuming to properly track offsets for each "play"
- Fix web audio pausing/resuming to properly track offsets for each "play"
- Fix issue with web audio where playing a sound during `hidden` game event would not mute the new sound


src/engine/Interfaces/IAudio.ts
```diff
@@ -0,0 +1,12 @@
+/// <reference path="../Promises.ts" />
+
+module ex {
+   export interface IAudio {
```
kamranayub: Need to add docs

src/engine/Resources/Sound.ts
```diff
@@ -171,41 +339,316 @@ module ex {
...
+            if (request.status !== 200) {
+               this._logger.error('Failed to load audio resource ', this.path, ' server responded with error code', request.status);
+               this.onerror(request.response);
+               complete.resolve(null);
+               return;
+            }
+
+            // load sound
+            this.setData(request.response).then(() => {               
+               this._isLoaded = true;
```
kamranayub: `_isLoaded = true` should be moved to `this.setData`

alanag13:
Per your feedback request and my usual mode of contribution, i will push that Track is not the right term from an audio perspective and is definitely misleading (suggests layering audio resources like in a studio recording). Instance is probably too generic. I would suggest Mix as this aligns to audio terms for the collection if settings on a play and would still fit as a term if we decide to implement multitrack.

eonarheim:
+100 for the awesome increase in test coverage! :)

kamranayub:
This is ready to merge when you all are good with the changes :+1:

eonarheim:
Looks solid, I like the abstraction of the audio instances 

:shipit:

PR #645, opened 2016-09-04
[Closes #505] Fix child actor width and height calculations to take into account parent scaling

### LINKED ISSUE #505: debug draw bounding boxes are not properly aligned for child actors
In the image below, the red square is a child actor of the white square.
image

kamranayub:
Closes #505

image

## Context

Child actors were not properly taking into account parent(s) scale when calculating their world width and height, so the bounding boxes and unit circle were not properly scaled/drawn.

Debug draw was compensating for previous world coord calculation (that has been fixed), so removed the canvas context transformations.

## Proposed Changes:

- Fix `Actor.getWidth()` and `Actor.getHeight()` to take into account global scale
- Fix debug drawing to work properly
- Add unit test coverage


jedeen:
Thanks for fixing this!

The pull request title is currently 96 characters long, almost double the [maximum length of a commit title](https://github.com/excaliburjs/Excalibur/blob/master/.github/CONTRIBUTING.md#commit-messages). Part of this length could be reduced by removing the [Closes #issue_num] prefix, since that information is already conveyed via body of the commit message in the pull request template.

eonarheim:
Looks good :shipit:!

PR #650, opened 2016-09-25
quickstart tutorial : Collision Fixed instead of Active

mseyne:
The paddle is pushed down by the ball when the collision occur.

## Proposed Changes for line 86:

- ex.CollisionType.Fixed


kamranayub:
@eonarheim can you also update the JSFiddle and after that we can merge?

eonarheim:
Thanks @mseyne! :shipit:

eonarheim:
Hi @mseyne thanks again for helping out, we'd love to hear any feedback you have on Excalibur. Let us know what you think of the docs, usability, and/or features. 

/cc @excaliburjs/core-contributers

PR #655, opened 2016-10-01
[#643] Fix Actor updates happening more than once per frame

### LINKED ISSUE #643: Actor updates are being called more than once per frame
#### Context

I noticed while doing some audio testing (/tests/audio) that when I clicked the red square, the sound actually ended up playing 5 times all at once.

When I stepped through, it's because when we capture pointer events, each browser pointer event is captured in an array during a frame (the array is cleared at the end of a frame).

So, even a single "click" can generate *n* pointer events during a single frame.

Is this expected? It means people need to have guards in their event handlers to compensate for being called multiple times in a single frame.

Keep in mind: there *could* be a case where if you are using multi-touch you could trigger pointer events in multiple locations on the same frame--this is where it may make sense to keep an array.

```
  var i: number = 0, 
        len: number = this._pointerUp.length;

         for (i; i < len; i++) {
            if (actor.contains(this._pointerUp[i].x, this._pointerUp[i].y, !isUIActor)) {
               actor.eventDispatcher.emit('pointerup', this._pointerUp[i]);
            }
         }

         i = 0;
         len = this._pointerDown.length;

         for (i; i < len; i++) {
            if (actor.contains(this._pointerDown[i].x, this._pointerDown[i].y, !isUIActor)) {
               actor.eventDispatcher.emit('pointerdown', this._pointerDown[i]);
            }
         }
```

#### Proposal

I propose only emitting a single `pointerup`, `pointercancel` and `pointerdown` event **per actor** during a frame (first valid pointer event wins). I would leave `pointermove` to allow multiple per frame.

This means, the logic would go:

- For each pointer event in array:
  - If actor contains pointer coords trigger event **and** break out of loop

eonarheim:
Closes #643

Pointer events were being triggered multiple times per frame. This was a result of the physics doing multiple passes on each actor by calling the update function multiple times.

## Proposed Changes:

This fixes the issue by refactoring the integration step of actor update out of actor update, so that it can be called multiple times for collision passes and not affect other operations that expect update to be called once per frame.

Additionally, this uncovered a bug in Actor.on('pointerup') opt-in which was subsequently fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Sprite culling was double scaling calculations ([#646](https://github.com/excaliburjs/Excalibur/issues/646))
 - Fix negative zoom sprite culling ([#539](https://github.com/excaliburjs/Excalibur/issues/539))
 - Fix Actor updates happening more than once per frame, causing multiple pointer events to trigger ([#643](https://github.com/excaliburjs/Excalibur/issues/643))
-- Fix `Actor.on('pointerup')` capturePointer events opt-in on event handler. The opt-in was triggering correctly for handlerse on 'pointerdown' and 'pointermove', but not 'pointerup'.
+- Fix `Actor.on('pointerup')` capturePointer events opt-in on event handler. The opt-in was triggering correctly for handlers on 'pointerdown' and 'pointermove', but not 'pointerup'.
 
 ## [0.7.0] - 2016-08-29
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - `Actor.debugDraw` now works properly for child actors ([#505](https://github.com/excaliburjs/Excalibur/issues/505), [#645](https://github.com/excaliburjs/Excalibur/issues/645))
 - Sprite culling was double scaling calculations ([#646](https://github.com/excaliburjs/Excalibur/issues/646))
 - Fix negative zoom sprite culling ([#539](https://github.com/excaliburjs/Excalibur/issues/539))
+- Fix Actor updates happening more than once per frame, causing multiple pointer events to trigger ([#643](https://github.com/excaliburjs/Excalibur/issues/643))
+- Fix `Actor.on('pointerup')` capturePointer opt-in
 
 ## [0.7.0] - 2016-08-29
 ### Breaking Changes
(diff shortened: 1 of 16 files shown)


GruntFile.js
```diff
@@ -151,10 +151,10 @@ module.exports = function (grunt) {
          },
 
          gitBuild: {
-            command: 'git clone https://github.com/excaliburjs/excalibur-dist build',
+            command: 'git clone -q https://github.com/excaliburjs/excalibur-dist build',
             options: {
                stdout: true,
-               failOnError: true
+               failOnError: false
```
jedeen: Are these changes to GruntFile a part of this pull request?

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - `Actor.debugDraw` now works properly for child actors ([#505](https://github.com/excaliburjs/Excalibur/issues/505), [#645](https://github.com/excaliburjs/Excalibur/issues/645))
 - Sprite culling was double scaling calculations ([#646](https://github.com/excaliburjs/Excalibur/issues/646))
 - Fix negative zoom sprite culling ([#539](https://github.com/excaliburjs/Excalibur/issues/539))
+- Fix Actor updates happening more than once per frame, causing multiple pointer events to trigger ([#643](https://github.com/excaliburjs/Excalibur/issues/643))
+- Fix `Actor.on('pointerup')` capturePointer opt-in
```
jedeen: What specifically is being fixed by the capturePointer change?

GruntFile.js
kamranayub: You can leave these changes out as master already has this (it doesn't have `-q` but that is just quiet mode).

src/engine/Scene.ts
```diff
@@ -254,25 +254,32 @@ module ex {
...
             // Run collision resolution strategy
-            if (this._broadphase && Physics.enabled) {
-               this._broadphase.update(this.children, collisionDelta);
-               this._broadphase.findCollisionContacts(this.children, collisionDelta);
+            if (this._broadphase && Physics.enabled) {               
+               this._broadphase.detect(this.children, collisionDelta);
+               for (i = 0, len = this.children.length; i < len; i++) {
+                     // helps move settle collisions
+                     // todo there is a better way to do this
+                     this.children[i].integrate(collisionDelta * .001);
```
kamranayub: Want to document the .001 multiplier?

kamranayub:
You snuck in some refactoring in physics--were those behavioral or organizational changes? Do they need to be documented?

eonarheim:
@kamranayub mostly organization changes, I'm not sure they need any more rigor than they have already.

src/engine/Scene.ts
eonarheim: I've extracted this into the ex.Physics global with some explaination

CHANGELOG.md
eonarheim: Take a look at my newest update to see if this is clear enough

GruntFile.js
eonarheim: Done

PR #656, opened 2016-10-01
[#543] Add time-scaling support to engine

### LINKED ISSUE #543: Add way to do "Time" scaling
This would be a useful feature if you wanted to control the speed of time in excalibur. With something like this you could lie to the underlying engine pieces and create "slow motion" or "super speed" effects.

Perhaps something like:
```typescript
engine.timeScale = 1.0; // 1 ms real time = 1 ms simulation time
engine.timeScale = 0.5; // 1 ms real time = .5 ms simulation time slowing things down
engine.timeScale = 1.5; // 1 ms real time = 1.5 ms simulation time speeding things up
```

kamranayub:
Closes #543

## Proposed Changes:

- Add `engine.timescale` property
- Add visual and unit tests
- Add `IGameLoop` mocked game loop to advance engine update
- Fix issue with unit tests

## Notes

I updated the Mock engine to support advancing the game loop by a duration (and optionally, a targeted FPS) to test the engine loop mechanism. I extracted out the main loop handler creation into a static method, `Engine.createMainLoop`. that takes the dependencies it needs.

There was also a subtle bug with how we were creating mock engines and scenes in the tests:

```ts
engine = mock.engine(0, 0, scene);
scene = new ex.Scene(engine);
```

Because Jasmine executes the `beforeEach` before each test, it does not necessarily clear the `describe`-scoped variable `scene` before each test--so the first line above was passing in **the previous test's scene instance** rather than the actual current scene being acted on. This wasn't a big issue for existing tests because the engine's update was never called--only the scene's update. Now that I have a mocked game loop, the issue was exposed because the engine's references are now accurate within the update loop.


src/engine/Engine.ts
```diff
@@ -1107,37 +1124,8 @@ O|===|* >________________>\n\
...
-   
-                     lastTime = now;
-               
-                  } catch (e) {
-                     window.cancelAnimationFrame(game._requestId);
-                     game.stop();
-                     game.onFatalException(e);
-                  }
-            })();
+            Engine.createMainLoop(this, window.requestAnimationFrame, Date.now)();
```
eonarheim: A+ this is much better

src/spec/CameraSpec.ts
```diff
@@ -6,31 +6,32 @@ describe('A camera', () => {
...
+      engine.currentScene = scene;
 
       sideCamera = new ex.SideCamera();
       lockedCamera = new ex.LockedCamera();
       baseCamera = new ex.BaseCamera();
    });
 
    it('can follow an actor if it is a lockedCamera', () => {
-      engine.camera = lockedCamera;
+      engine.currentScene.camera = lockedCamera;
```
eonarheim: Is there a reason we are being specific about the current scene?

src/spec/Mocks.ts
```diff
@@ -2,6 +2,17 @@
 
 module Mocks {
 
+   export interface ITime {
```
eonarheim: I like this!

src/spec/Mocks.ts
```diff
@@ -87,16 +111,53 @@ module Mocks {
...
+               emit: function () { /* do nothing */  }
+            },
+            _hasStarted: true,
+            _update: (<any>ex.Engine.prototype)._update,
+            _draw: function () { /* do nothing */ }
          };
          return mockEngine;
       };
+
+      loop(game: ex.Engine): IGameLoop {
```
eonarheim: Some comments about how this is working might be good

src/spec/CameraSpec.ts
eonarheim: Mock was out of date, this is correct ignore me

src/engine/Engine.ts
```diff
@@ -568,6 +570,21 @@ O|===|* >________________>\n\
...
+      public get timescale() {
+         return this._timescale;
+      }
+
+      /**
+       * Sets the current engine timescale factor. Useful for creating slow-motion effects or fast-forward effects
+       * when using time-based movement.
+       */
+      public set timescale(value: number) {
+         this._timescale = value;
```
eonarheim: Also I think timescale 0 is also an error

PR #668, opened 2016-10-11
[#663] Add alias `magnitude` to `Vector`

### LINKED ISSUE #663: Add ex.Vector.magnitude()
#### Context

Currently this is possible to calculate in Excalibur using `ex.Vector.distance()` with no argument to distance, however this is mega confusing and not immediately clear as the correct thing to do when translating from vector math.

#### Proposal

Add an alias called `ex.Vector.magnitude()` that calls `ex.Vector.distance()`

kamranayub:
Closes #663

## Proposed Changes:

- Add `Vector.magnitude` alias that calls `Vector.distance()`
- Clarify docs on `Vector.distance()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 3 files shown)


eonarheim:
:shipit:  🚀

PR #669, opened 2016-10-11
[#666] Fix `oldVel` and `oldPos` values being updated

### LINKED ISSUE #666: ex.Actor.body.oldPos not being set
#### Steps to Reproduce

No matter want position an `ex.Actor.body.pos` takes on over time `ex.actor.body.oldPos` stays at (0, 0)
#### Expected Result

oldPos should take on the value of pos from the previous frame

kamranayub:
Closes #666 
## Proposed Changes:
- Fix `oldVel` value being set to the current frame's value due to `addEqual` mutator method
- Fix `oldPos` not being set at all
## Notes

Two issues I discovered:
1. `oldVel` was wrong, it was being set to the current frame's value
2. `oldVel` (and now `oldPos`) are being set in the `integrate()` method of Actor but that method can be called multiple times per update, hence oldVel and oldPos are probably not accurate for use within an update loop especially if they are supposed to represent the previous frame values.

### DIFF
--- a/src/engine/Actor.ts
+++ b/src/engine/Actor.ts
@@ -1173,10 +1173,8 @@ module ex {
           !(this instanceof Label)) {
           totalAcc.addEqual(ex.Physics.acc);
        }
-
-       this.oldVel = this.vel;
-       this.vel.addEqual(totalAcc.scale(seconds));
-
+       
+       this.vel.addEqual(totalAcc.scale(seconds));       
        this.pos.addEqual(this.vel.scale(seconds)).addEqual(totalAcc.scale(0.5 * seconds * seconds));
 
        this.rx += this.torque * (1.0 / this.moi) * seconds;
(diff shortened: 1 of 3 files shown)


src/engine/Actor.ts
```diff
@@ -1229,6 +1227,11 @@ module ex {
           }
        }
 
+       // Capture old values before integration step updates them
+       this.oldVel.setTo(this.vel.x, this.vel.y);
+       this.oldPos.setTo(this.pos.x, this.pos.y);
+
+       // Run Euler integration
        this.integrate(delta);
```
kamranayub: @eonarheim is this a problem? Scene will call this once, then will call it again later multiple times. I don't know enough about how the physics work to know if that's bad.

eonarheim: Yes, I think you're on to something. The `oldVel` and `oldPos` values would not be correct if we set them in the integrate step since integrate can be called multiple times.

eonarheim:
LGTM :shipit:

eonarheim:
Indeed, thanks for the call out. Please update the changelog.md, then this can be merged.

kamranayub:
Done!

PR #670, opened 2016-10-18
[#662] Add Line math methods

### LINKED ISSUE #662: Add some Line math methods
#### Context

The current `ex.Line` is limited in its utility for doing line math. It can provide the basics but I'd like some nice utility methods.

#### Proposal

- [x] `ex.Line.m` - Get the slope `m` component of the line equation
- [x] `ex.Line.b` - Get the y-intercept `b` component of the line equation
- [x] `ex.Line.findPoint({ x?, y? })` - Find a point on a line given a single X or a Y value
- [x] `ex.Line.hasPoint(ex.Vector, threshold? = 0)` - Whether or not a line has the given point (optionally provided a threshold). See [this SO post](http://stackoverflow.com/questions/11907947/how-to-check-if-a-point-lies-on-a-line-between-2-other-points)

```ts
export function getLineSlopeValue(ln: ex.Line) {
        return (ln.end.y - ln.begin.y) / (ln.end.x - ln.begin.x);
    }

    export function getLineIntercept(ln: ex.Line, slope: number) {
         return ln.begin.y - (slope * ln.begin.x);
    }

    export function findPoint(ln: ex.Line, x: number = null, y: number = null): ex.Vector {
        var m = getLineSlopeValue(ln);
        var b = getLineIntercept(ln, m);

        if (x !== null) {
            return new ex.Vector(x, (m * x) + b);
        } else if (y !== null) {
            return new ex.Vector((y - b) / m, y);
        }
    }
```

kamranayub:
Closes #662

## Proposed Changes:

  - `ex.Line.slope` for the raw slope (m) value
  - `ex.Line.intercept` for the Y intercept (b) value
  - `ex.Line.findPoint(x?, y?)` to find a point given an X or a Y value
  - `ex.Line.hasPoint(x, y, threshold)` to determine if given point lies on the line

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - `ex.Vector.magnitude` alias that calls `ex.Vector.distance()` to get magnitude of Vector ([#663](https://github.com/excaliburjs/Excalibur/issues/663))
+- Added new `ex.Line` utilities ([#662](https://github.com/excaliburjs/Excalibur/issues/662)):
+  - `ex.Line.slope` for the raw slope (m) value
+  - `ex.Line.intercept` for the Y intercept (b) value
+  - `ex.Line.findPoint(x?, y?)` to find a point given an X or a Y value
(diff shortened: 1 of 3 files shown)


eonarheim:
Looks good to me 🚀

PR #673, opened 2016-10-26
[#536] Fix ex.Utils.clamp min/max handling

### LINKED ISSUE #536: Update ex.Util.clamp to be smarter
Here is the code :P

``` typescript
function clamp(val, min, max) {
   return Math.min(Math.max(min, val), max);
}
```

FerociousQuasar:
Closes #536 

## Changes:
- Update ex.Util.clamp to use Math libraries Min and Max
- Add unit tests for ex.Util.clamp


kamranayub:
Also, I'm thinking to be good citizens we should ensure this function has proper tests around it.

FerociousQuasar:
Roger.
Will add unit tests to make sure this behaves as expected.
Will update CHANGELOG.md with the changes.
Will update commit message to meet the appropriate standard.

eonarheim:
@jedeen @FerociousQuasar Let's not worry about the commit messages on this one since we'll squash the messages.

Any issues @excaliburjs/core-contributors? LGTM :shipit:

kamranayub:
Only two thoughts, could you add a couple tests that also test boundary numbers (0, Infinity, -Infinity) to ensure it still works as expected?

src/spec/UtilSpec.ts
```diff
@@ -16,4 +16,10 @@ describe('Utility functions', () => {
       expect(ex.Util.getSideFromVector(ex.Vector.Up)).toBe(ex.Side.Top);
       expect(ex.Util.getSideFromVector(ex.Vector.Down)).toBe(ex.Side.Bottom);
    });
+
+   it('can clamp a number to a maximum and minimum', () => {
+      expect(ex.Util.clamp(0, 10, 20)).toBe(10);
+      expect(ex.Util.clamp(15, 10, 20)).toBe(15);
+      expect(ex.Util.clamp(30, 10, 20)).toBe(20);
```
kamranayub: Could you add another couple tests for Infinity and -Infinity to test those boundaries?

src/engine/Util/Util.ts
```diff
@@ -104,7 +104,7 @@ module ex.Util {
    }
 
    export function clamp(val, min, max) {
```
kamranayub: Could you add the `number` types to each parameter to make sure people don't accidentally pass in other types?

PR #676, opened 2016-11-03
Fix converting same string to lower case 3-4 times

hogart:
## Proposed Changes:

- change one of 'pointerdown' in Actor._checkForPointerOptIn to 'pointerup' since this is clearly a typo;
- also don't covert same string to lowercase 3-4 times in this method

### DIFF
--- a/src/engine/Actor.ts
+++ b/src/engine/Actor.ts
@@ -689,14 +689,15 @@ module ex {
     }
 
     private _checkForPointerOptIn(eventName: string) {
-       if (eventName && (eventName.toLowerCase() === 'pointerdown' ||
-          eventName.toLowerCase() === 'pointerup' ||
-          eventName.toLowerCase() === 'pointermove')) {
+      if (eventName) {
+        const normalized = eventName.toLowerCase();
+        if (normalized === 'pointerup' || normalized === 'pointerdown' || normalized === 'pointermove') {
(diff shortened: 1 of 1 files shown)


eonarheim:
Hi @hogart! Thanks for the PR.

I don't see the typo in the original based on the git diff, but I do like the refactoring.

image

This was definitely a bug in the past that was fixed , perhaps you have an older version of master?
https://github.com/excaliburjs/Excalibur/commit/cbee62b2442051fef072f11d64deb8c37ee4ae44#diff-20f78184845d86cb2a750f0e519151d9L692 

If I don't hear anything from @excaliburjs/core-contributers I'll accept the PR.

Thanks for contributing!

hogart:
I'm sorry for the mess, I followed link from 0.7 documentation and it looked like this: https://github.com/excaliburjs/Excalibur/blob/f94b38e/src/engine/Actor.ts#L691
So I jumped into fixing without checking twice:)

hogart:
@jedeen please check if this commit message will do?

> Fix converting same string to lower case 3-4 times

jedeen:
@hogart That message looks good, but is 51 characters long. Luckily, "lowercase" can be written as one word, putting it under our 50 character limit; just that one small change and you're good to go!

eonarheim:
lgtm :shipit:

PR #680, opened 2016-11-05
Fix Label#getTextWidth returns incorrect result

### LINKED ISSUE #679: Label#getTextWidth() returns incorrect result
#### Steps to Reproduce
Let me just demonstrate the code instead of putting it as STR:
```ts
const engine = new ex.Engine({
    width: 800,
    height: 600,
});
const title = new Label('hello world!', 100, 100);
engine.add(title)
console.log(title.getTextWidth(engine.ctx)); // line A, 52 or so
title.fontSize = 40;        
console.log(title.getTextWidth(engine.ctx)); // should be around 200 at least
```        

Problem is this method doesn't take into account fontSize and fontUnit properties (unlike private `_fontDraw`).

#### Context
May be it's not the issue with sprite fonts, which are preferable way, but it's bug anyway.

#### Proposal
Make `getTextWidth` use same approach as `_fontDraw`. Actually, I'm preparing PR to fix that.

hogart:
Closes #679 

## Proposed Changes:

- fix `Label#getTextWidth` so it returns proper number
- introduce protected readonly `_fontString` property, which properly combines fontFamily with fontSize and is used in `_fontDraw` and `getTextWidth`.

Basically, this PR changes method so it takes into account not only fontFamily, but fontSize too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fix `Actor.oldPos` and `Actor.oldVel` values on update ([#666](https://github.com/excaliburjs/Excalibur/issues/666))
+- Fix `Label.getTextWidth` returns incorrect result ([#679](https://github.com/excaliburjs/Excalibur/issues/679))
 
 ## [0.7.1]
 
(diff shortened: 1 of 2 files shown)


kamranayub:
@hogart @eonarheim I think we need to add a Changelog entry for this fix.

eonarheim:
@kamranayub there is a changelog entry

Issue #653, opened 2016-09-25
UseXCollision methods should require collision type


kamranayub:
### Context

It seems like for each physics body, you need to set collision type anyway, the shortcut methods should require CollisionType as the first parameter.
### Proposal

``` ts
this.body.useCircleCollision(ex.CollisionType.Active, ...)
```

eonarheim:
Collision type is now a property of bodies

PR #686, opened 2016-11-08
Fix typo in Physics docs (egde -> ground)

hogart:
## Proposed Changes:

- Fix typo in Physics docs, there's no `edge` variable around, definitely meant `ground`


hogart:
BTW, is this example supposed to do nothing? Blue actor started to move only after I added `block .collisionType = ex.CollisionType.Active;`.

kamranayub:
@eonarheim I believe `collisionType` has to be set to Active for things to move, correct? I had an issue open to auto-set this (or let people specify) in those shortcut methods. #653

hogart:
@kamranayub that proposition makes a lot of sense.

Issue #481, opened 2015-07-07
Unnecessary calls to _loadPixels() and _applyEffects() in Sprites


eonarheim:
This would make sprites way more efficient on draw. `_loadPixels()` and `_applyEffects` seem to be called even when they are completely loaded and effects have been applied. This can be seen in the console network tab as the base64 images are reloaded into the in memory image tag src attributes.

``` javascript
// we should avoid setting src as much as possible
 this.internalImage.src = this._spriteCanvas.toDataURL('image/png');
```

![image](https://cloud.githubusercontent.com/assets/612071/8537900/92c047c6-242a-11e5-8e97-3fda8d1db386.png)

![image](https://cloud.githubusercontent.com/assets/612071/8537920/cc169106-242a-11e5-9dc0-571a0aa9e456.png)

![image](https://cloud.githubusercontent.com/assets/612071/8537927/eba74844-242a-11e5-971e-fceee8117cd5.png)

eonarheim:
We should measure the impact of this issue, perhaps with a visual test or a timing test in the chrome profiling tools

eonarheim:
Closed by #1730

PR #689, opened 2016-11-11
Optimize Loader#draw

hogart:
## Proposed Changes:

- Move image creating to cached protected property
- Extract logo src, it's dimensions and background color to public properties for easier customizing

Actually I was looking into #481, when I noticed that logo gets loaded huge amount of times, so I went to investigate it and managed to slightly optimize Loader and made it bit more flexible.


eonarheim:
LGTM :shipit:

hogart:
@eonarheim I'm glad you liked it:)

PR #690, opened 2016-11-11
Minimize Sprite#_applyEffects calls count

hogart:
## Proposed Changes:

- In `Actor`, don't reapply effects to drawing in `update` method, do it only in `draw` and if `frames` or `opacity` was changed
- In `Sprite`, always set `_effectsDirty` flag after effects were applied
- Few minor changes, like don't calculate same value twice in same method

It's my attempt to improve things on issue #481. It doesn't close issue completely because source of unnecessary calls to `Sprite#_loadPixels` remains mystery to me, but I was able to reduce amount of `_applyEffects` calls.
I can miss some effects' use cases (although sandbox tests seemed working), so please review this carefully.

### DIFF
--- a/src/engine/Drawing/Sprite.ts
+++ b/src/engine/Drawing/Sprite.ts
@@ -1,5 +1,7 @@
 module ex {
 
+   var clamp = Util.clamp;
+   
    /**
     * Sprites
     *
@@ -140,7 +142,6 @@ module ex {
 
       private _loadPixels() {
          if (this._texture.isLoaded() && !this._pixelsLoaded) {
-            var clamp = ex.Util.clamp;
             var naturalWidth = this._texture.image.naturalWidth || 0;
             var naturalHeight = this._texture.image.naturalHeight || 0;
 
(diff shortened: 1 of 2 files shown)


src/engine/Drawing/Sprite.ts
```diff
@@ -1,5 +1,7 @@
 module ex {
 
+   var clamp = ex.Util.clamp;
```
kamranayub: Do we need this?

hogart: It's used further below. I moved this assignment out of method.

kamranayub: Sure, but I mean do we **really** need it? We're already in the `ex` module, why not just do `Util.clamp`. I know you didn't do it, but while we're here maybe we should just clean it up ;)

hogart: Updated as per suggestion:)

hogart:
Awesome, thanks, guys:)

PR #693, opened 2016-11-21
[#692] Actor Bounding Boxes Include Rotation and Scale

### LINKED ISSUE #692: Actor.getBounds() does not return an accurate bounding box when actors are scaled/rotated
#### Steps to Reproduce
Experimenting with scale vectors on Actor objects seems to indicate that scaling is being done after rotation, which leads to unexpected results. 

Consider the following scenario:
I have a square actor of size (10, 10) set to scale (1, 2). This results in a stretched-out actor 20 high and 10 wide. I rotate the actor by 90 degrees, expecting it now to be 20 wide and 10 high, but instead it remains 20 high and 10 wide because the rotation is applied before the scale.

See the following codepen with evidence and workarounds
https://codepen.io/eonarheim/pen/yVMBQM?editors=0011

#### Expected Result

I would expect that calling `actor.getBounds().getWidth()` to yield 20 in the case above with the rotated/scaled actor, but it does not.

#### Actual Result
In the rotated case `actor.getBounds().getWidth()` incorrectly returns 10.

Perhaps there also needs to be a shorthand for this like? `actor.getBoundingWidth()` or `actor.getCalculatedWidth()` Or should `actor.getWidth()` just return this value?

eonarheim:
This pull request fixes an issue that was brought up on the google groups where rotated and scaled width and height could not be easily calculated, and the `actor.getWidth()` and `actor.getHeight()` were misleading.

@excaliburjs/core-contributers Before we merge this PR, I think we should have a discussion about `actor.getWidth()` and `actor.getHeight()` should they continue to operate without knowledge of rotation? is there a compelling reason to not just report the actual width and height of the bouding box? Please comment.

Closes #692 

## Proposed Changes:

- Add a static method for creating bounding boxes `ex.BoundingBox.fromPoints`
- Add a method for rotating bounding boxes `ex.BoundingBox.rotate`
- `actor.getBounds()` and `actor.getRelativeBounds()` now always return the correct bounds regardless of scale or rotation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Fix `Label.getTextWidth` returns incorrect result ([#679](https://github.com/excaliburjs/Excalibur/issues/679))
 - Fix semi-transparent PNGs appear garbled ([#687](https://github.com/excaliburjs/Excalibur/issues/687))
 - Fix incorrect code coverage metrics, previously our test process was reporting higher than actual code coverage ([#521](https://github.com/excaliburjs/Excalibur/issues/521))
+- Fix `Actor.getBounds()` and `Actor.getRelativeBounds()` to return accurate bounding boxes based on the scale and rotation of actors. ([#692](https://github.com/excaliburjs/Excalibur/issues/692))
 
 ## [0.7.1]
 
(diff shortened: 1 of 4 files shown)


eonarheim:
@excaliburjs/core-contributers Can this be merged?

Issue #97, opened 2013-11-19
Add Basic Excalibur Diagnostics tool


eonarheim:
Something like

```
ex.Diagnostics.getFps()
ex.Diagnostics.getActorCount()
ex.Diagnostics.getActorPoolUtilization()
ex.Diagnostics.getLogs()
ex.Diagnostics.getMemoryUsage()
ex.Diagnostics.getTimeline()
ex.Diagnostics.showToolbar() //use the dom to show diagnostics window;

```

Also we should include Excalibur build information in the diagnostics. 

Resources:

- https://docs.webplatform.org/wiki/apis/timing/properties/memory

eonarheim:
Perhaps use `ex.debug` insteead of `ex.Diagnostics` like suggested in #390

kamranayub:
Part of this was done as part of 0.8.0 release, moving to 0.9.0 for more features. It might be better to split this up into separate defined chunks.

jedeen:
I recommend just pulling the features that didn't get added yet into their own new issue, referencing this issue, and throwing that new issue in 0.9, so we can close this for 0.8.

kamranayub:
Agreed

jedeen:
closed via #694

PR #694, opened 2016-11-21
[#97] Add basic frame statistics

kamranayub:
Part of #97

To make our way towards closing #97 we need basic statistics to gather and display. This adds some basic frame statistics that are collected during a frame and is designed to be extended later on with more stats. Since all stats are encapsulated in a class, users can add historical tracking and logging of past statistics using the `preframe` and `postframe` hooks on engine.

## Proposed Changes:

- Add `Engine.debug` to access current frame stats and (later) provide debugging flags/functions
- Add `Engine.stats.currFrame` and `Engine.stats.prevFrame` to hold current and previous frame statistics
- Add new `FrameStats` class to hold frame statistics
- Add new `postframe` and `preframe` events which fire before and after the frame is drawn/updated to aid in stat display/collection


kamranayub:
Thoughts on the names for `currentFrameStats` and `prevFrameStats`? I am leaning towards renaming to `currFrameStats` and `lastFrameStats` (Current and Last, vs. Next and Previous). I would also be willing to entertain something like `Engine.stats.current` and `Engine.stats.last`.

Also I think I need to fix the visual test as now that I think about it, the engine reuses the two instances to keep memory overhead low--so to do historical, I have to "clone" them otherwise the instances will change.

alanag13:
"previous" is more clear than "last" (one could argue that the current frame is last chronologically). I would go curr/prev for brevity's sake and do something like `Engine.stats.currFrame` and `Engine.stats.prevFrame` to allow for possibility of extending `Engine.stats` to include things beyond just frame statistics (e.g. perhaps game or scene).

kamranayub:
@alanag13 I wonder since I already have a `debug` property on Engine and it shortcuts to the current frame stats, if I should simply add current and previous frame stats to that object vs. creating a new stats holder on Engine?

alanag13:
@kamranayub  I agree that you probably don't want a whole new holder. Could you do that and just have `Engine.stats` be an alias for `Engine.debug.stats`? Basically I'm making an argument that the term 'debug' doesn't always apply when trying to get at these (although sometimes it does). Small usability/discoverability thing. If we do expand stats we could just do it on the `Engine.debug` object then.

kamranayub:
Alright, fixed some bugs, verified with tests, we should be good.

PR #696, opened 2016-11-30
[#691] Add shrinkpacked dependencies and docs

### LINKED ISSUE #691: Implement shrinkpack to harden and speed up dev/CI builds
#### Context

Excerpted from "[target problem](https://github.com/JamieMason/shrinkpack#target-problem)":

> We were all very happy with this process and the convenience of npm in particular, but the phase of our builds where npm install listed a huge amount of network traffic would always raise the same concerns;
> 
> - This seems slow, wasteful, and inefficient.
> - We really depend on registry.npmjs.org, what do we do if it goes down?
> 
> The first suggestion was always to check in our dependencies, but the idea of some large and chatty commits whenever we chose to upgrade or change them would put us off.
> 
> Some teams went a little further and decided that pain was acceptable and decided to proceed, only to find that some packages such as phantomjs and node-sass helpfully install the appropriate binary for you depending on what system you're running.
> 
> This meant that if Chris added phantomjs or node-sass to the project on his Mac and checked it into the repository, Helen wouldn't be able to use it on her Windows Machine.

https://github.com/JamieMason/shrinkpack

#### Proposal

Update npm workflow to use `npm shrinkwrap` and `shrinkpack` together. 

> Note: if you're developing an npm package and want to use shrinkpack to speed up and harden your development and CI environments, adding npm-shrinkwrap.json and node_shrinkwrap to your .npmignore file will allow you to do that, without publishing your shrinkpacked dependencies to the registry.

kamranayub:
Closes #691

## Proposed Changes:

- Add shrinkpacked dependencies

## Notes

The only change this will add to our workflow is anytime the npm dependencies change, we have to shrinkwrap + shrinkpack before merging to master. I updated the README with the two steps.


eonarheim:
@excaliburjs/core-contributors Will need to take note of this

PR #721, opened 2016-12-06
[#711] Removing reference to setting deprecated value

### LINKED ISSUE #711: Remove webkitImageSmoothingEnabled
#### Context

We don't need to set `webKitImageSmoothingEnabled` anymore--it'd be best to have a condition around each of those properties:

```
'CanvasRenderingContext2D.webkitImageSmoothingEnabled' is deprecated and will be removed in M55, around November 2016. Please use 'CanvasRenderingContext2D.imageSmoothingEnabled' instead. See https://www.chromestatus.com/features/5639849247768576 for more details.
```

```ts
(<any>this.ctx).imageSmoothingEnabled = isSmooth;
(<any>this.ctx).webkitImageSmoothingEnabled = isSmooth;
(<any>this.ctx).mozImageSmoothingEnabled = isSmooth;
(<any>this.ctx).msImageSmoothingEnabled = isSmooth;
```

This is generating a warning in Chrome currently.

#### Proposal

Can be:

```ts
var ctx: any = this.ctx;
ctx.imageSmoothingEnabled = isSmooth;
("webkitImageSmoothingEnabled" in ctx) && ctx.webkitImageSmoothingEnabled = isSmooth;
("mozImageSmoothingEnabled" in ctx) && ctx.mozImageSmoothingEnabled = isSmooth;
("msImageSmoothingEnabled" in ctx) && ctx.msImageSmoothingEnabled = isSmooth;
```

guahanweb:
Closes #711

## Changes:

- Remove direct references to deprecated context smoothing attributes
- Added dynamic setting of context smoothing attributes by existence

### DIFF
--- a/src/engine/Engine.ts
+++ b/src/engine/Engine.ts
@@ -788,10 +788,14 @@ O|===|* >________________>\n\
        */
       public setAntialiasing(isSmooth: boolean) {
          this._isSmoothingEnabled = isSmooth;
-         (<any>this.ctx).imageSmoothingEnabled = isSmooth;
-         (<any>this.ctx).webkitImageSmoothingEnabled = isSmooth;
-         (<any>this.ctx).mozImageSmoothingEnabled = isSmooth;
-         (<any>this.ctx).msImageSmoothingEnabled = isSmooth;
+
+         var ctx: any = this.ctx;
+         ctx.imageSmoothingEnabled = isSmooth;
+         ['webkitImageSmoothingEnabled', 'mozImageSmoothingEnabled', 'msImageSmoothingEnabled'].forEach((smoothing) => {
+            if (smoothing in ctx) {
+               ctx[smoothing] = isSmooth;
+            }
+         });
       }
 
       /**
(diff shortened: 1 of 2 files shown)


src/engine/Engine.ts
```diff
@@ -788,10 +788,14 @@ O|===|* >________________>\n\
...
       public setAntialiasing(isSmooth: boolean) {
          this._isSmoothingEnabled = isSmooth;
-         (<any>this.ctx).imageSmoothingEnabled = isSmooth;
-         (<any>this.ctx).webkitImageSmoothingEnabled = isSmooth;
-         (<any>this.ctx).mozImageSmoothingEnabled = isSmooth;
-         (<any>this.ctx).msImageSmoothingEnabled = isSmooth;
+
+         var ctx: any = this.ctx;
+         ctx.imageSmoothingEnabled = isSmooth;
+         ['webkitImageSmoothingEnabled', 'mozImageSmoothingEnabled', 'msImageSmoothingEnabled'].forEach((smoothing) => {
```
kamranayub: It's tiny, but using `forEach` does incur anonymous function creation overhead, we could do:

```
for (var smoothing of [arr]) {
}
```

instead for speed.

guahanweb: Is TS configured to polyfill the if...of syntax? That would be my only concern with it (browser support). Otherwise, I'll make the change.

eonarheim:
Thanks @guahanweb!

PR #722, opened 2016-12-07
[#410] Implement Option bag constructor

### LINKED ISSUE #410: Configure option bag constructor overloads for Actor, Scene, Engine, ParticleEmitters, etc
### Context
This would make configuration more terse, and easier to pass configuration around.

### Proposal
``` javascript
var game = new ex.Engine({
   width: 800,
   height: 600,
   canvasElementId: "game",
   pointerScope: ex.Input.PointerScope.Document
});

var actor = new ex.Actor({
   x: 50,
   y: 60,
   width: 10,
   height: 10,
   scale: new ex.Point(2,2),
   ay; 10,
   texture: new ex.Texture('myimage.png')   
});
```

alanag13:
Closes #410 (eventually)

## Changes:
Added  option bag constructors to the following classes:
- Vector

### DIFF
--- a/src/engine/Algebra.ts
+++ b/src/engine/Algebra.ts
@@ -6,8 +6,8 @@ module ex {
     * A 2D vector on a plane. 
     */
    export class Vector implements IVectorOptions {
-      public x: number;
-      public y: number;
+      public x: number = 0;
+      public y: number = 0;
       /**
        * A (0, 0) vector
        */
--- a/src/engine/Algebra.ts
+++ b/src/engine/Algebra.ts
@@ -1,10 +1,13 @@
+/// <reference path="Interfaces/ConstructorBags/AlgebraOptions.ts" />
+
 module ex {
 
    /**
     * A 2D vector on a plane. 
     */
    export class Vector {
-
+      public x: number;
+      public y: number;
       /**
        * A (0, 0) vector
        */
@@ -68,11 +71,25 @@ module ex {
          return true;
       }
 
+      constructor();
+      /**
+       * @param vectorOptions coordindates of the Vector
+       */
+      constructor(vectorOptions: IVectorOptions);
       /**
        * @param x  X component of the Vector
        * @param y  Y component of the Vector
        */
-      constructor(public x: number, public y: number) { }
+      constructor(x: number, y: number);
+      constructor(optionsOrX?: number | IVectorOptions | any, y?: number) {
+            if (typeof optionsOrX === 'number') {
+               this.x = optionsOrX;
+               this.y = y || 0;
+            }else if (typeof optionsOrX === 'object') {
+               this.x = optionsOrX.x || 0;
+               this.y = optionsOrX.y || 0;
+            }
+       }
 
       /**
        * Sets the x and y components at once
--- a/src/engine/Algebra.ts
+++ b/src/engine/Algebra.ts
@@ -5,7 +5,7 @@ module ex {
    /**
     * A 2D vector on a plane. 
     */
-   export class Vector {
+   export class Vector implements IVectorOptions {
       public x: number;
       public y: number;
       /**
@@ -81,7 +81,7 @@ module ex {
        * @param y  Y component of the Vector
        */
       constructor(x: number, y: number);
-      constructor(optionsOrX?: number | IVectorOptions | any, y?: number) {
+      constructor(optionsOrX?: number | IVectorOptions, y?: number) {
             if (typeof optionsOrX === 'number') {
                this.x = optionsOrX;
                this.y = y || 0;
--- a/src/engine/Interfaces/ConstructorBags/AlgebraOptions.ts
+++ b/src/engine/Interfaces/ConstructorBags/AlgebraOptions.ts
@@ -1,0 +1,7 @@
+module ex {
+   export interface IVectorOptions {
+      x?: number;
+      y: number;
+   }
+
+}
--- a/src/engine/Interfaces/ConstructorBags/AlgebraOptions.ts
+++ b/src/engine/Interfaces/ConstructorBags/AlgebraOptions.ts
@@ -1,7 +1,7 @@
 module ex {
    export interface IVectorOptions {
       x?: number;
-      y: number;
+      y?: number;
    }
 
 }
--- a/src/spec/AlgebraSpec.ts
+++ b/src/spec/AlgebraSpec.ts
@@ -11,10 +11,18 @@ describe('Vectors', () => {
       var v = new ex.Vector(20, 200);
       expect(v).not.toBeNull();
    });
-   
+
    it('can have values set', () => {
       var v = new ex.Vector(20, 200);
       
+      expect(v.x).toEqual(20);
+      expect(v.y).toEqual(200);
+
+      var v2 = new ex.Vector({x: 20, y: 200});
+      expect(v2.x).toEqual(20);
+      expect(v2.y).toEqual(200);
+
+      v = new ex.Vector(v2);
       expect(v.x).toEqual(20);
       expect(v.y).toEqual(200);
       
@@ -28,6 +36,11 @@ describe('Vectors', () => {
       
       expect(v.x).toEqual(0);
       expect(v.y).toEqual(0);
+
+      v = new ex.Vector();
+      expect(v.x).toEqual(0);
+      expect(v.y).toEqual(0);
+
    });
    
    it('has a Zero constant', () => {


alanag13:
Super exciting Vector property bags! now you can do the following:
```
new Vector(1,1); //the old classic still works
new Vector({x: 1, y: 1}); //via IVectorOptions interface
new Vector(); //yields x: 0, y: 0
````
As I'm writing this I'm realizing I forgot to Vector implement IVectorOptions so that you can pass a Vector the contructor. Will do that and add.

**Comment by eonarheim** [comment]
@alanag13 One thing I might change right now is to move the interfaces from the separate files, to be next to the class that they are for in the same file. I really don't see a good reason to split them into separate files, Especially if you want to make a change to a class and are forced to open 2 files minimum to do it.  For example the IVectorOptions interface delcaration could just be right before the class declaration for Vector.

File: src/engine/Algebra.ts [hunk 1]
```diff
@@ -68,11 +71,25 @@ module ex {
...
+      /**
+       * @param vectorOptions coordindates of the Vector
+       */
+      constructor(vectorOptions: IVectorOptions);
       /**
        * @param x  X component of the Vector
        * @param y  Y component of the Vector
        */
-      constructor(public x: number, public y: number) { }
+      constructor(x: number, y: number);
```
kamranayub: Should these all be marked as obsolete so we can deprecate them in a future version?

src/engine/Interfaces/ConstructorBags/AlgebraOptions.ts
```diff
@@ -0,0 +1,7 @@
+module ex {
+   export interface IVectorOptions {
```
kamranayub: Agree with Erik, I would prefer to keep these side-by-side as they are directly related.

src/engine/Interfaces/ConstructorBags/AlgebraOptions.ts
```diff
@@ -0,0 +1,7 @@
...
+      x?: number;
```
kamranayub: Missing public JSDoc

src/engine/Algebra.ts
```diff
@@ -68,11 +71,25 @@ module ex {
          return true;
       }
 
+      constructor();
+      /**
+       * @param vectorOptions coordindates of the Vector
```
kamranayub: Misspelling of coordinates

alanag13:
@excaliburjs/core-contributors So last night I tried to implement this on another (seemingly) simple class, `Ray`. I quickly discovered that maintaining backwards compatibility is going to be really gross for any class whose first constructor argument is currently an object, because any logic in the constructor won't be able to properly coerce the type (we can only detect which primitive it is, which is why everything's fine if the first arg is not an object). We could do some shady stuff like make assumptions based on the type of arguments passed to the constructor, but I'm not sure how I feel about that. So I have a couple clarifying questions.

 - is the expectation that this will be a breaking change on all constructors that we implement it on / is that okay?
- should I allow for backwards compatabilty where possible like I did here with `Vector`?
- are there an constructors that we would not want to implement this on? For example, `Vector` is currently instantiated with `new Vector(x,y)`. Is `new Vector({x: 1, y: 1})` overkill? I'm thinking yes, probably and would like to propose that this change be implemented on classes with numPublicProperties > 2.

Let me know  your thoughts.

kamranayub:
- It would be nice to avoid a breaking change. Remember all we need to do to see if it's an options bag is to check if `arguments.length === 1`, because due to your second point, this will only be for complex objects.
- I think it's totally possible in all cases but let me know if you see the opposite
- I agree, >2 props seems like good candidates, also ones where everything is optional

src/engine/Algebra.ts
```diff
@@ -68,11 +71,25 @@ module ex {
...
+       */
+      constructor(vectorOptions: IVectorOptions);
       /**
        * @param x  X component of the Vector
        * @param y  Y component of the Vector
        */
-      constructor(public x: number, public y: number) { }
+      constructor(x: number, y: number);
+      constructor(optionsOrX?: number | IVectorOptions, y?: number) {
+            if (typeof optionsOrX === 'number') {
```
kamranayub: Could simplify this logic for all ctors:

```
if (arguments.length === 1 && typeof arguments[0] === 'object') {
  // options bag
} else {
  // overload
}
```

alanag13: Think you meant 'classes' not 'actors', but this makes a lot of sense and I will plan on this. I must have been tired because I had my in head that this would be an issue for existing constructors where there was just one arguments that was an object, but we would of course just not change those.

alanag13: e.g. if a constructor currently takes an Actor as it's only parameter we wouldn't make an options bag constructors consisting of just `{ actor: new Actor() }`. This also adheres to my previous suggestion of just doing this for Params > 2. Cool.

alanag13:
P.S. Does everyone like the idea have an empty constructor on all classes that would just create an instance with all the defaults? I did this on `Vector` and think it would probably be handy everywhere.

kamranayub:
Yeah I think that's a good idea, you could do that automatically using the `?` operator on the parameter.

src/engine/Algebra.ts
```diff
@@ -1,10 +1,13 @@
+/// <reference path="Interfaces/ConstructorBags/AlgebraOptions.ts" />
+
 module ex {
 
    /**
     * A 2D vector on a plane. 
     */
-   export class Vector {
-
+   export class Vector implements IVectorOptions {
```
eugirdor: The `implements IVectorOptions` seems unnecessary.  You can already pass a `Vector` to the constructor because it implicitly implements the `IVectorOptions` interface.

eugirdor: @kamranayub I didn't mean changing the constructor to take a `Vector`, I agree that should be `IVectorOptions`.

I was referring to having `implements IVectorOptions` on the `Vector` class.  With the constructor taking `IVectorOptions`, you can already in any object that satisfies the interface without having to explicitly implement `IVectorOptions`.

kamranayub: Sure, but we would like to take advantage of the compiler errors/warnings that would be generated when adding new options. Removing the implementing would make it harder to ensure every class correctly implements the properties we expect.

eugirdor: Fair enough

kamranayub: Thanks for jumping in with feedback!

kamranayub:
Couple things we've talked about in person, just want to note them down:

- Probably change the convention to be `[Class]Props` instead of the word Options to better represent what the properties are and will be used for
- Properties should all be implemented as getters and setters, to provide flexibility with implementation
- We may want a Props base/mixin, to create a `set` method that can provide a strongly-typed way of setting multiple prop values at once (e.g. `Vector.set({ x: 0, y: 0 })`)
- To aid in generic methods that can work against props-based entities (like pools), we may want a Props interface (to go along with previous point) -- might also be useful to force implementing `defaults` props to use for extending (e.g. `Vector.defaults`)
- Anything that we probably want to be eventually poolable should have a Props (including Vector?)

This would allow us to do something like:

```ts
var vectorPool = Pool.create(Vector); // typed as Pool<T> with constraint T implements Props
var vector1 = vectorPool.create({ x: 1, y: 1}); // TS will type this as Vector & Pooled
vector1.set({ x: 1, y: 1 }); // strongly-typed way to set multiple props at once
vector1.x = 1; // still can get/set individual prop
vector1.reset(); // reset the state
```

Props are basically the state of the class, and if we enumerate them properly in Props, it can be strongly typed all the way down via TS.

kamranayub:
Here is a full working example using props, mapped types, and pools:

```ts
interface VectorProps {
    x: number;
    y: number;
}

function extend(...o) {
    return o.reduce((p, c, i) => {
        return { ...p, ...c };
    });
}

abstract class Class<TProps> {

    set(props: Partial<TProps>) {
        for (var k in props) {            
            (<any>this)[k] = props[k];            
        }
    }
}

class Vector extends Class<VectorProps> implements VectorProps {
    static Defaults = { x: 0, y: 0 };

    x: number;
    y: number;

    constructor(props?: Partial<VectorProps>) {
        super();
        this.set(extend(Vector.Defaults, props));
    }
}

interface Poolable<T extends TProps, TProps> {
    create<T extends TProps>(props?: Partial<TProps>): T & PooledInstance;
}

interface PooledInstance {
    reset();
}

class Pool {
    static create<T extends TProps, TProps>(ctor: { new (props?: Partial<TProps>): T }): Pooled<T, TProps>  {        
        var pooled = new Pooled<T, TProps>(ctor);
        
        return pooled;
    }
} 

class Pooled<T extends TProps, TProps> implements Poolable<T, TProps> {

    constructor(private _ctor: { new (props?: Partial<TProps>): T }) {

    }

    create(props?: Partial<TProps>): T & PooledInstance {
        var ins = new this._ctor(props);

        return this.mixin(ins);
    };

    mixin(instance: T): T & PooledInstance {
        instance['reset'] = function () {

        }

        return <T & PooledInstance>instance;
    }

}

var v = new Vector();
var vp = Pool.create<Vector, VectorProps>(Vector);
var v2 = vp.create({ x: 1, y: 1 });
console.log(v, v.x, v.y);
console.log(vp, v2, v2.x, v2.y);
```

You can play with it by using the [Playground](http://www.typescriptlang.org/play).

alanag13:
Thanks for all the notes. Have we confirmed that typedoc support ts 2.1, which is what we would need in order to use mapped types?

kamranayub:
It doesn't yet but I was gonna work on it this weekend probably

PR #723, opened 2016-12-07
[#707] Add report for browser feature support

### LINKED ISSUE #707: On bootup, log enabled/detected features
#### Context

To aid in browser/device testing, it might be good to have a formal mechanism to register detected/enabled browser features and log them at bootup time so a developer can see what Excalibur sees as enabled or available.

#### Proposal

- [ ] WebGL support
- [ ] Web Audio vs. HTML5 Audio support
- [ ] Gamepad API

guahanweb:
Closes #707 

## Changes:

- Added singleton browser feature support matrix into Detector.ts
- Added debug step in Engine.ts to log feature support matrix
- Added a getter to access the current support matrix from consuming modules: `Detector.getBrowserFeatures()`

For future reference, any additional features that are desired to show up in this log statement need two things:
1.  Add the IIFE to assign support value in `Detector._loadBrowserFeatures()`
2.  Add the human readable label to the features map in `Detector.logBrowserFeatures()`

image

As it stands, if debugging is enabled, here is the output to the console:

image

Feedback encouraged!

### DIFF
--- a/src/engine/Util/Detector.ts
+++ b/src/engine/Util/Detector.ts
@@ -54,7 +54,7 @@ module ex {
                args.push('font-weight: bold; color: green');
                args.push('font-weight: normal; color: inherit');
             } else {
-               msg += '(%c\u2717%c)'
+               msg += '(%c\u2717%c)';
                args.push('font-weight: bold; color: red');
                args.push('font-weight: normal; color: inherit'); 
             };
--- a/src/engine/Util/Detector.ts
+++ b/src/engine/Util/Detector.ts
@@ -15,7 +15,7 @@ module ex {
    /**
     * Interface for detected browser features matrix
     */
-   export interface DetectedFeatures {
+   export interface IDetectedFeatures {
       canvas: boolean;
       arraybuffer: boolean;
       dataurl: boolean;
@@ -31,7 +31,7 @@ module ex {
     */
    export class Detector {
 
-      private _features: DetectedFeatures = null;
+      private _features: IDetectedFeatures = null;
 
       public failedTests: string[] = [];
 
@@ -85,7 +85,7 @@ module ex {
        * Executes several IIFE's to get a constant reference to supported
        * features within the current execution context.
        */
-      private _loadBrowserFeatures(): DetectedFeatures {
+      private _loadBrowserFeatures(): IDetectedFeatures {
          return {
             // IIFE to check canvas support
             canvas: (() => {
--- a/src/engine/Util/Detector.ts
+++ b/src/engine/Util/Detector.ts
@@ -1,12 +1,37 @@
 /// <reference path="Log.ts" />
+
+/**
+ * This is the list of features that will be used to log the supported
+ * features to the console when Detector.logBrowserFeatures() is called.
+ */
+const REPORTED_FEATURES: Object = {
+   webgl: 'WebGL',
+   webaudio: 'WebAudio',
+   gamepadapi: 'Gamepad API'
+};
+
 module ex {
+
+   /**
+    * Interface for detected browser features matrix
+    */
+   export interface DetectedFeatures {
+      canvas: boolean;
+      arraybuffer: boolean;
+      dataurl: boolean;
+      objecturl: boolean;
+      rgba: boolean;
+      webaudio: boolean;
+      webgl: boolean;
+      gamepadapi: boolean;
+   }
 
    /**
     * Excalibur internal feature detection helper class
     */
    export class Detector {
 
-      private _features: Object = null;
+      private _features: DetectedFeatures = null;
 
       public failedTests: string[] = [];
 
@@ -18,8 +43,6 @@ module ex {
        * Returns a map of currently supported browser features. This method
        * treats the features as a singleton and will only calculate feature
        * support if it has not previously been done.
-       * 
-       * @return {Object} the list of supported features
        */
       public getBrowserFeatures(): Object {
          if (this._features === null) {
@@ -31,16 +54,8 @@ module ex {
       /**
        * Report on non-critical browser support for debugging purposes.
        * Use native browser console colors for visibility.
-       * 
-       * @return {void}
        */
       public logBrowserFeatures(): void {
-         let features = {
-            webgl: 'WebGL',
-            webaudio: 'WebAudio',
-            gamepadapi: 'Gamepad API'
-         };
-
          let msg = '%cSUPPORTED BROWSER FEATURES\n==========================%c\n';
          let args = [
             'font-weight: bold; color: navy',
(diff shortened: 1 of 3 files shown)


src/engine/Util/Detector.ts
```diff
@@ -5,9 +5,117 @@ module ex {
     * Excalibur internal feature detection helper class
     */
    export class Detector {
-      
+
+      private _features: Object = null;
```
kamranayub: Because this object is exposed publicly, it would be preferable to expose the type information to consumers. Can we create an interface to use for this?

```ts
export interface DetectedFeatures {
    canvas: boolean;
    // etc...
}

private _features: DetectedFeatures = null;
```

src/engine/Util/Detector.ts
```diff
@@ -5,9 +5,117 @@ module ex {
...
+
+         args.unshift(msg);
+         console.log.apply(console, args);
+      }
+
+      /**
+       * Executes several IIFE's to get a constant reference to supported
+       * features within the current execution context.
+       * 
+       * @return {Object} a map of features
```
kamranayub: Same as above, can leave off return type.

src/engine/Util/Detector.ts
```diff
@@ -5,9 +5,117 @@ module ex {
...
+      public constructor() {
+         this._features = this._loadBrowserFeatures();
+      }
+
+      /**
+       * Returns a map of currently supported browser features. This method
+       * treats the features as a singleton and will only calculate feature
+       * support if it has not previously been done.
+       * 
+       * @return {Object} the list of supported features
```
kamranayub: I know this automatically generated by VSCode, but you can technically leave off the return type info--TypeDoc (our doc generator) can infer it via TypeScript compiler.

src/engine/Util/Detector.ts
```diff
@@ -5,9 +5,117 @@ module ex {
...
+            this._features = this._loadBrowserFeatures();
+         }
+         return this._features;
+      }
       
+      /**
+       * Report on non-critical browser support for debugging purposes.
+       * Use native browser console colors for visibility.
+       * 
+       * @return {void}
```
kamranayub: Can leave this off, TypeDoc can infer this.

src/engine/Util/Detector.ts
```diff
@@ -5,9 +5,117 @@ module ex {
...
+       */
+      public logBrowserFeatures(): void {
+         let features = {
```
kamranayub: Would it make sense to pull this up out of the class towards the top of the file so we can see this as a reminder to update if we add new features later on?

src/engine/Util/Detector.ts
```diff
@@ -5,9 +5,117 @@ module ex {
...
+         let msg = '%cSUPPORTED BROWSER FEATURES\n==========================%c\n';
+         let args = [
+            'font-weight: bold; color: navy',
+            'font-weight: normal; color: inherit'
+         ];
+
+         let supported = this.getBrowserFeatures();
+         for (let feature of Object.keys(features)) {
+            if (supported[feature]) {
+               msg += '(%c\u2713%c)';
```
kamranayub: A comment with the unicode character pasted in might be helpful

src/engine/Util/Detector.ts
```diff
@@ -5,9 +5,117 @@ module ex {
...
+               args.push('font-weight: bold; color: green');
+               args.push('font-weight: normal; color: inherit');
+            } else {
+               msg += '(%c\u2717%c)';
```
kamranayub: A comment with the unicode character pasted in might be helpful

guahanweb:
Great feedback, @kamranayub! I've made the requested changes. Let me know how that looks.

guahanweb:
@jedeen, just updated the CHANGELOG. Please review and be sure that I accurately represented your understanding of this feature. I only documented the public facing surfaces, as I assume this is a CHANGELOG for consumption, not necessarily to document all internal modifications. Let me know if I have misunderstood the desire here.

jedeen:
@guahanweb You're correct, that's its primary purpose!

src/engine/Util/Detector.ts
```diff
@@ -1,13 +1,134 @@
...
+      public constructor() {
+         this._features = this._loadBrowserFeatures();
+      }
+
+      /**
+       * Returns a map of currently supported browser features. This method
+       * treats the features as a singleton and will only calculate feature
+       * support if it has not previously been done.
+       */
+      public getBrowserFeatures(): Object {
```
kamranayub: This function should return the new interface you made, `IDetectedFeatures` instead of `Object`.

kamranayub: Also, optionally, we could use a ES5 getter here since the result is cached once when the game starts and only returns the cached response from then on. I'm cool either way since this is mostly an internal function.

src/engine/Util/Detector.ts
```diff
@@ -1,13 +1,134 @@
...
+   gamepadapi: 'Gamepad API'
+};
+
 module ex {
 
    /**
+    * Interface for detected browser features matrix
+    */
+   export interface IDetectedFeatures {
+      canvas: boolean;
```
kamranayub: Optional: could add the new `readonly` modifier to each of the properties so TS will throw if someone accidentally assigns to these props.

src/engine/Util/Detector.ts
guahanweb: Updated the type, good catch. I'm torn on using `get` here, even if only for clarity. Referencing `this.getBrowserFeatures` as a property seems dirty to me, somehow. Since this is an exposed method that we're documenting, **AND** it doesn't actually get invoked unless you're in DEBUG mode, I think I'd prefer the optimization of implicit caching. 

If you're not in DEBUG mode, `this._loadBrowserFeatures()` doesn't execute until you call `Detector.logBrowserFeature()` or `Detector.getBrowserFeatures()` explicitly.

kamranayub:
Looks great! I think this is ready

src/engine/Util/Detector.ts
kamranayub: Interesting, we updated our TS version to 2.0, if you're using VS Code what does it say in the bottom right? It should say TypeScript 2.0.10 or something... did you do an `npm install`?

src/engine/Util/Detector.ts
kamranayub: That's fair, we can keep it as-is! 👍

kamranayub:
Optionally can add `readonly` but if you can't get TS to compile it, I'm cool with merging.

guahanweb:
Man, @kamranayub, I feel like an idiot. Everything was working, but I hadn't done an `npm install`, so VSCode was still using my old TS version. Just added the `readonly` flag and confirmed compile and tests running.

kamranayub:
No problem, it's all done! Thanks so much for the contribution!

PR #725, opened 2016-12-08
[#661] Fix Scene/Actor activation and initialization order

### LINKED ISSUE #661: Scene initialize should initialize any children added before calling onActivate
There is an issue with how actor's are initialized by scenes the first time.

Actor's are initialized within their `update` method, which means to initialize an actor you **must** call update at least once on it.

However, this will cause a problem if you expect an actor to be initialized immediately when a Scene is initialized. Because a Scene is initialized by Engine when `goToScene` is called:

``` ts
if (!this.currentScene.isInitialized) {
  this.currentScene.onInitialize.call(this.currentScene, this);
  this.currentScene.eventDispatcher.emit('initialize', new InitializeEvent(this));
  this.currentScene.isInitialized = true;
}

this.currentScene.onActivate.call(this.currentScene);
```

In-between `onInitialize` and `onActivate`, one might expect any actors that are in the scene to be initialized but that is not the case as you can see--actors can only be initialized once the `update` loop is run (which is _after_ `onActivate`).
#### Steps to Reproduce

``` ts
class Level extends ex.Scene {

  private _actor: ex.Actor;

  onInitialize() {
    this._actor = new ex.Actor(0, 0);
    this.add(this._actor);
  }

  onActivate() {
    // on the first activation, this will be false,
    // on subsequent activations, it will be true (because update has run)
    console.log(this._actor._isInitialized);
  }
}

game.addScene('level', new Level(game));

game.start(() => {
  game.goToScene('level');
});
```
#### Expected Result

Console should log `true` because scene was initialized and actor should have been initialized.
#### Actual Result

When a Scene is initialized, actors are not initialized, so the console prints `false`.

eonarheim:
Closes #661

## Changes:

- Refactor initialization logic in scene and actor into `initialize(engine: Engine)` methods
- Create tests demonstrating that the documented initialize/activate flow works as expected
   - Deactivate old scene
   - Initialize new scene (if uninitialized)
   - Initialize actors and any children (if uninitialized)
   - Activate new scene
   - Begin mainloop


src/engine/Actor.ts
```diff
@@ -426,6 +426,27 @@ module ex {
...
+     * Gets wether the actor is Initialized 
+     */
+    public get isInitialized(): boolean {
+       return this._isInitialized;
+    }
+
+    /**
+     * Initializes this actor and all it's child actors, meant to be called by the Scene before first update.
+     */
+    public initialize(engine: Engine) {
```
kamranayub: This is a bigger discussion, but I think we'll need to make a decision soon about whether or not we should tell people to override these methods but end up leaving a landmine that blows up if you forget to call `super.whatever`. It's happened with draw, update, and now initialize.

I think we should consider moving all lifecycle events to internal methods and using the `onWhatever` convention to denote methods that can safely be overridden without any side-effects.

src/engine/Scene.ts
```diff
@@ -122,6 +120,34 @@ module ex {
          // will be overridden
          this._logger.debug('Scene.onDeactivate', this);
       }
+
+      /**
+       * Initializes actors in the scene
+       */
+      public initializeChildren(): void {
```
kamranayub: Add `@internal` flag?

src/engine/Scene.ts
```diff
@@ -122,6 +120,34 @@ module ex {
...
+       * Gets whether or not the [[Scene]] has been initialized
+       */
+      public get isInitialized(): boolean {
+         return this._isInitialized;
+      }
+
+      /**
+       * Initializes the scene before the first update, meant to be called by engine
+       */
+      public initialize(engine: Engine) {
```
kamranayub: Add `@internal` flag?

src/engine/Actor.ts
kamranayub: I updated our external Trello discussion, so no changes immediately needed here until we decide.

src/engine/Scene.ts
kamranayub: As we discussed, probably do `public _initializeChildren()` and mark as `@internal` for good measure. Hopefully that is enough warning to people.

eonarheim:
@kamranayub Let me know if you'd like any more changes

kamranayub:
LGTM

PR #730, opened 2016-12-18
[#565] Fix loader with no resource bug

### LINKED ISSUE #565: Excalibur will choke on null assets
image

image

eonarheim:
Closes #565 

## Changes:

- Fixed a bug where Excalibur would not load a game with an empty loader
- Added a test to confirm and fix

### DIFF
--- a/src/spec/ResourceSpec.ts
+++ b/src/spec/ResourceSpec.ts
@@ -1,6 +1,8 @@
 /// <reference path="jasmine.d.ts" />
 /// <reference path="require.d.ts" />
 /// <reference path="Mocks.ts" />
+/// <reference path="TestUtils.ts" />
+
 
 describe('A generic Resource', () => {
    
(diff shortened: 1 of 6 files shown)


src/spec/ResourceSpec.ts
```diff
@@ -23,6 +23,23 @@ describe('A generic Resource', () => {
       expect(resource.isLoaded()).toBe(false);
       
    });
+
+   describe('without data', () => {
+      it('should not fail on load', (done) => {
+         var emptyLoader = new ex.Loader();
+         var game = new ex.Engine({
```
kamranayub: This is the second place where this exact code is being used to create an engine--now might be a good time to create a TestUtils helper to create a testable engine (maybe default width/height to 500 but allow override).

```ts
var game = TestUtils.engine()
```

PR #731, opened 2016-12-22
[#538] Implement a better seeded Random 

### LINKED ISSUE #538: Add better random utilities for Excalibur
It is pretty well documented that V8 and other browser default random utilities are pretty lacking
https://medium.com/@betable/tifu-by-using-math-random-f1c308c4fd9d#.y6mxmug2a

With this in mind, we should improve the existing random utilities in Excalibur. I propose that we make a `ex.Random` module for this purpose.
- [ ] Update random int to use rejection sampling (outlined in the link above) over the current "multiply-and-floor" method
- [x] Update random in range to do the same
- [x] ~~Update random strategies to use a better PRNG generator than `Math.random()` maybe using `window.crypto` that has reasonable browser compatability http://caniuse.com/#feat=cryptography~~ 
- [x] Add random boolean helper (ex.Random.bool()) that will return true or false (coin flip)
- [x] Add Mersenne Twister "seeded" random generation: https://github.com/pigulla/mersennetwister

Related to #396 and #491 we should include these in the `ex.Random` module when completed

eonarheim:
Closes #538 

## Changes:

- Implements a seeded random utility in Excalibur called `ex.Random` using the popular Mersenne-Twister algorithm documented [here](http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/MT2002/emt19937ar.html) and [here](http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/emt.html)
- Added new utilities for random
  * Random in range (`ex.Random.floating()`/`ex.Random.integer()`), 
  * Coin-tosses with variable outcomes `ex.Random.bool()`
  * Pick-one `ex.Random.pickOne(array)`
  * Pick-set `ex.Random.pickSet(array, numPicks)`
  * Dice roll helpers for d4, d6, d8, d10, d12, and d20
- Tests for the random features included

Additionally I have compared this implementation to the native C implementation done by the people who originally proposed the paper and it returns the same values for a seeded sequence as our implementation does. (downloaded and compiled locally using windows subsystem for linux).

### DIFF
--- a/src/engine/Math/Random.ts
+++ b/src/engine/Math/Random.ts
@@ -135,10 +135,25 @@ module ex {
       }
 
       /**
+       * Returns a new array random picking elements from the original
+       * @param array Original array to pick from
+       * @param numPicks can be any positive number
+       * @param allowDuplicates indicates whether the returned set is allowed duplicates (it does not mean there will always be duplicates 
+       * just that it is possible)
+       */
+      public pickSet<T>(array: Array<T>, numPicks: number, allowDuplicates: boolean = false): Array<T> {
+         if (allowDuplicates) {
+            return this._pickSetWithDuplicates(array, numPicks);
+         } else {
+            return this._pickSetWithoutDuplicates(array, numPicks);
+         }
+      }
+
+      /**
        * Returns a new array randomly picking elements in the original (not reused)
        * @param numPicks must be less than or equal to the number of elements in the array.
        */
-      public pickSet<T>(array: Array<T>, numPicks: number): Array<T> {
+      private _pickSetWithoutDuplicates<T>(array: Array<T>, numPicks: number): Array<T> {
          if (numPicks > array.length || numPicks < 0) {
             throw new Error('Invalid number of elements to pick, must pick a value 0 < n <= length');
          }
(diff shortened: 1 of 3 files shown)


eonarheim:
I mentioned in #538 that I'm not sure we need to implement rejection sampling since it is a subtle problem to close this issue.

kamranayub:
We also talked about random array? Length, min, max.

alanag13:
Have you considered pick set with duplicates possible vs duplicates not possible?

eonarheim:
@alanag13 @kamranayub Take a look

alanag13:
I might have implemented pickSetWithDuplicates as just a boolean flag of allowDuplicates(default: false) passed to the original pickSet function.

src/engine/Math/Random.ts
```diff
@@ -0,0 +1,253 @@
...
+
+         return result;
+      }
+
+      /**
+       * Returns a new array random picking elements from the original allowing duplicates
+       * @param numPicks can be any positive number
+       */
+      public pickSetWithDuplicates<T>(array: Array<T>, numPicks): Array<T> {
+         if (numPicks < 0) {
```
kamranayub: Missing check for `numPicks > array.length`

alanag13:
The name pickSetWithDuplicates kind of suggests that the resulting set will always contain duplicates.

src/engine/Math/Random.ts
eonarheim: This is actually okay, if there are duplicates allowed. Since you can now have an array larger than the source if you allow dups

eonarheim:
@alanag13 I agree, I'll make that change

src/engine/Math/Random.ts
kamranayub: Ahhh, I get it

Issue #733, opened 2016-12-28
Add Angular2/WebPack/SystemJS getting started helper


eonarheim:
#### Context
There are more and more people trying to use Angular/WebPack/SystemJS with the new ES6 style module syntax and there are a few pitfalls that folks are running into since it requires modifying the Excalibur dist for it to work correctly.

#### Proposal
We should add this documentation to our README and/or have a quickstart for such users in our docs site.

See https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/excaliburjs/z_8HThXxkaY

kamranayub:
Done in #734

Issue #717, opened 2016-12-04
Fix sourcemaps not being included in dists


kamranayub:
#### Context

Noticed dists don't include the generated source maps.

#### Proposal

- [ ] Fix gitignore in dist repo
- [ ] Fix nuspec License path

PR #734, opened 2016-12-31
[#606] Convert codebase to external modules

### LINKED ISSUE #606: Switch to new ES6 style modules in Excalibur
#### Context

Currently, we are using the old `/// <reference path="./path/to/my/module"/>` syntax. This is old and clunky and it seems like VSCode has a hard time interpreting modules and dependencies. 
image
#### Proposal

I propose we move to the new ES6 style module syntax (commonjs) that is now available in the TypeScript compiler so that dependencies look like this

image

We just need to set the module style in the compiler settings to `commonjs`

This will be a pretty large task since it will need to touch most all files in Excalibur

kamranayub:
Closes #606, #733, #717 

## Changes:

- Changed internal Excalibur code into modules (ES6 style)
- Excalibur uses UMD module style for distribution (supports CommonJS, AMD, and window global)
- Bundled AMD modules using Almond.js micro loader
- Added two declaration files to dist, `excalibur.d.ts` and `excalibur.amd.d.ts`
- Added AMD bundle to distribution `excalibur.amd.js`
- Added inlined source maps for browser debugging
- Cleaned up Gruntfile and various compilation tasks

## TODO

- [x] Test visual tests
- [x] Test declaration file
- [x] Get tests compiling/working
- [x] Get debugging working
- [x] Update docs regarding module loading/referencing
- [x] Test documentation generation
- [x] Verify public API surface matches current to maintain full backwards-compatibility
- [x] Test AMD module loading
- [x] Test CJS module loading
- [x] Test importing
- [x] Update referencing in docs

## Notes

- I'm a wizard and am amazed I got this to even work.
- The best way to review this code is commit-by-commit. I split the work up into chunks.
- **NO FUNCTIONALITY WAS MODIFIED** unless noted in my review comments, these are cosmetic/import/formatting changes only in the core files.
- The public API definition lives in `src/engine/Index.ts`. This is the entrypoint to bootstrap ex in the browser. Anything exported here will show up on the `ex` global object.
- Almond.js is super tiny, <1KB overhead, it allows us to maintain the same "embed and go" attitude.
- Alternatively, users can reference **excalibur.amd.js** to get the AMD-ready bundle where they can use their own AMD-compatible module loader (e.g. require.js)
- I had to remove a circular dependency inside `Actor`. @eonarheim please review, I will comment on the line.


src/engine/Actor.ts
```diff
@@ -1,1128 +1,1126 @@
...
+   /**
+    * Perform euler integration at the specified time step
+    */
+   public integrate(delta: number) {
+      // Update placements based on linear algebra
+      var seconds = delta / 1000;
+
+      var totalAcc = this.acc.clone();
+      // Only active vanilla actors are affected by global acceleration
+      if (this.collisionType === CollisionType.Active) {
```
kamranayub: @eonarheim I removed this check. Is that OK? Is there a different/better way to check this without introducing a circular reference to Actor's sub-classes?

src/engine/Actor.ts
```diff
@@ -1,1128 +1,1126 @@
...
+      this.integrate(delta);
+
+      // Update actor pipeline (movement, collision detection, event propagation, offscreen culling)
+      for (var trait of this.traits) {
+         trait.update(this, engine, delta);
+      }
+
+      // Update child actors
+      for (var i = 0; i < this.children.length; i++) {         
+         this.children[i].update(engine, delta);
```
kamranayub: I moved the stats incrementing code to `Scene` instead, to remove the reference to `UIActor`.

src/engine/Actor.ts
eonarheim: I don't think that would be okay, because if you have global accel `ex.Physics.acc` it will affect all actors including UIActors, Triggers, and Labels if they are active collision types which I don't think that would be expected. There local acc can be set if that's what people want.

Having an `affectedByPhysics` property may be good, but that was what `CollisonType.Active` was kinda meant for making it unclear which properties you need to set for things to work...

I saw in the Traits module you are using `ActorUtils.isVanillaActor()` is there are reason we can't also use that here? Or does it also introduce a circular dependency?

src/engine/Actor.ts
eonarheim: I think that is fine, a simpler actor is a better actor ;)

src/engine/Actor.ts
kamranayub: Nah because it is still an indirect dependency on its sub classes through the dependency resolution. I'm just wondering a) when would someone ever make a trigger/label/UIActor active collision? and b) if they did, they'd do it knowingly, so maybe they *should* be affected by physics--if they weren't supposed to, they shouldn't inherit from Actor ;) It points to Actor being too complicated and we should probably split it up more.

eonarheim: I'm definitely feeling that they shouldn't inherit from actor... a lot of bagage comes with it. Really the only thing they share are a position, bounds, update, and draw. Perhaps actor and these other actor adjacent things need to inherit from some base object with these properties? `BaseObject` or `Component`?

eonarheim: Also fair point, if UIActor, Label, or Trigger are active they did do it on purpose. So probably okay for now.

My previous point is definitely out of scope for this issue and is a major change that needs more discussion ;)

kamranayub: 👍

kamranayub:
This is ready, I've verified the API surface comparing 0.8.0 and the latest side-by-side in the browser.

kamranayub:
I've updated the Installation documentation, once this is merged I will push two new repositories:

- example-ts-angular2
- example-ts-webpack

Both I have working locally against this branch 👍

docs/source/installation.rst
```diff
@@ -69,25 +67,92 @@ If you're using TypeScript, be sure to reference the declaration file ``excalibu
...
+   
+   npm install excalibur -D
 
-This will allow you to reference Excalibur using TypeScript with the import style syntax:
+In a TypeScript project, you can reference Excalibur with the ES6 import style syntax:
 
 .. code-block:: typescript
 
-   import ex from 'excalibur'
+   import * as ex from 'excalibur'
```
eonarheim: Is it not possible to use `import ex from 'excalibur'`

kamranayub:
Done @eonarheim 👍

docs/source/installation.rst
kamranayub: The old import style was because we had to use "legacy" `export = ex` syntax in declaration files. Now we can use proper ES6 style importing 👍 

This supports `import { Engine } from 'excalibur'` style too.

PR #741, opened 2017-01-07
[#509] Add world and screen coordinates on Pointer

### LINKED ISSUE #509: Add a way to poll the last cursor position in excalibur

kamranayub:
Closes #509

This allows you to poll the last pointer position (for whatever pointer you are targeting). It will be `null` if the Pointer wasn't active last frame.

## Changes:

- Add `Pointer.lastPagePos`, `Pointer.lastScreenPos` and `Pointer.lastWorldPos` properties on Pointer class
- Add visual test to confirm coordinates are correct

## Notes

Cleaned up non-ignored files in sandbox directory as well.


src/engine/Input/Pointer.ts
```diff
@@ -236,7 +240,7 @@ export class Pointers extends Class {
          var x: number = e.pageX - Util.getPosition(this._engine.canvas).x;
          var y: number = e.pageY - Util.getPosition(this._engine.canvas).y;
          var transformedPoint = this._engine.screenToWorldCoordinates(new Vector(x, y));
-         var pe = new PointerEvent(transformedPoint.x, transformedPoint.y, 0, PointerType.Mouse, e.button, e);
+         var pe = new PointerEvent(transformedPoint.x, transformedPoint.y, e.pageX, e.pageY, 0, PointerType.Mouse, e.button, e);
```
eonarheim: Is pageX and pageY the right value to use? What if the canvas for the game is centered in the middle of the screen? Don't we want the screen coordinates where top left of the canvas element is (0, 0) regardless of where the canvas is on the screen?

https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/pageX

kamranayub: There's not a standard event that would expose that, we would need to calculate it manually (find offset of Canvas element)--which might be a good feature to add. The current functionality is what I would expect screenX and screenY to return, I would not expect it to be relative to the canvas element.

eonarheim: I would always expect screen coordinates (0, 0) to be the top left of the canvas no matter where it is, that's how UI actors and events work right now

kamranayub:
Let me verify what coords worldtoscree coordinates returns and ensure it
matches, you might be right and if so it should be consistent. I'd like to
keep the page coordinates too, maybe under pagePos

src/engine/Input/Pointer.ts
kamranayub: I will verify the screen coords match what we expect and maybe rename the current `lastScreenPos` to `lastPagePos`.

kamranayub:
@eonarheim Fixed!

image

The box screen coordinate is 600, 500.

image

PR #743, opened 2017-01-07
[#588] Actor can be added/removed from Scene in same frame

### LINKED ISSUE #588: Adding and removing actor within same update frame shouldn't cause problems
Either warn or throw an error, if this will cause problems--otherwise find a way to avoid issues.

kamranayub:
Closes #588

## Changes:

- Added verification tests

## Notes

I couldn't reproduce the error when adding/removing same actor.


kamranayub:
Added other tests, ready for review. 👍

PR #753, opened 2017-02-05
[#410] Add option bag constructors on commonly used classes

### LINKED ISSUE #410: Configure option bag constructor overloads for Actor, Scene, Engine, ParticleEmitters, etc

alanag13:
(eventually) closes #410. This will add option bag constructors to several classes, and will additionally provide default values to any properties not passed in to each objects' constructor.


alanag13:
I mostly borrowed heavily from @kamranayub 's original implementation suggestion on my previous PR with one notable difference: The use of interfaces to maintain options for the property bag has been totally removed. Instead, extending `Defaultable<T>` where T is the same type that you are extending will automatically require a `defaults` property of type `Partial<T>`. e.g. `Actor` would extend `Defaultable<Actor>`, which would require a `defaults` property of type `Partial<Actor>`.

The main advantage this has is that we no longer would have to main the interfaces for each option bag -- if we added a property to a class that extends `Defaultable<T>`, the new property would automatically be allowed to be passed to the constructor and be used in instantiation.

The two (in my opinion, minor) disadvantages to this approach are:

- You need to manually maintain default property values as new properties get added. Without this, new properties that get passed to the constructor would still be properly set, but if they are omitted, they would be undefined.

- When using a type definition instead of an interface that contains only properties (no functions), ts / intellisense will allow you to pass redefined functions as part of the optionbag constructor. My code explicitly prevents replacing the functions on the instance, but this could still be slightly misleading.

Let me know your thoughts.

alanag13:
Here's a thought I just had for the reset functionality -- I think `reset` should set the properties back not to the "parameter is missing so set it to value x" values, but to what the properties were _when the object was first instantiated._

Example:
````
var vector = new Vector(); // x = 0, y = 0.
vector.x = 20;
vector.y = 100;
vector.resetProperties(); //x = 0, y = 0.

//However....

var vector = new Vector({x: 20, y:100}); // x = 0, y = 0.
vector.x = 30;
vector.y = 200;

//current implementation has this set x = 0, y = 0, but I might want/expect x = 20, y = 100
//since that it has how this vector came into this world and it has never been x = 0, y = 0
vector.resetProperties(); 
```` 
Thoughts?

**Comment by jedeen** [comment]
I think resetting it back to its original values is very useful. In fact, it's what I would expect `resetProperties()` to do. Perhaps we could use `clearProperties()` or something similar if we still want to maintain the ability to zero out the values on an object.

**Comment by kamranayub** [comment]
I like what you did. I am not sure I'd prefer a shorter `props` vs. `properties` syntax though.

```
vector.props({ x: 0, y: 0 });
vector.reset(); // use initial
vector.reset(true); // use defaults
```

I like the conciseness of that vs. the heavier method names. You could also hang reset off props.

```
interface Defaultable<T> {
  props: {
    (p: Partial<T>); // set
    reset(default?: boolean);
  }
}
```

File: src/engine/Interfaces/Defaultable.ts [hunk 1]
```diff
@@ -0,0 +1,21 @@
+import { extend } from '.././Util/Util';
+
+export abstract class Defaultable<T> {
```
kamranayub: I think it may be possible to replace `T` with `this` type to refer back to the class itself, reducing need for generic arguments. Not sure though.

eonarheim:
@alanag13 Any updates on this?

alanag13:
I'm still trying to think of a good way around the inheritance problem @eonarheim and I discussed a while ago. Basically, my current approach won't work on classes that are already inheriting from a base class (like, uh, `Actor`). Is there some sort of decorator pattern /wrapper object approach that would be better here?

eonarheim:
@alanag13 Looking at the TS 2.2 release notes https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html

A good way to handle the problem you're talking about with a type already extending a base class.

Now there is a way! I think this is close

```typescript
type Constructor<T> = new(...args: any[]) => T;
function Defaultable<T extends Constructor<{}>, TProps>(Base: T) {
	return class extends Base {
		defaults: Partial<T>;

		public setProperties(instance: Defaultable<T>, props?: Partial<T>): Defaultable<T> {

			var mergedProperties = extend(instance.defaults, props);

			for (var k in mergedProperties) {
				if (typeof <any>instance[k] !== 'function') {
					(<any>instance)[k] = mergedProperties[k];
				}
			}
			return <Defaultable<T>>instance;
		}

		public resetProperties() {
			this.setProperties(this);
		}
		
        constructor(...args: any[]) {
			super(...args);
			this.setProperties(this.defaults);
        }
    }
}

class Actor extends Defaultable(Class) {
}
```

Plus TypeDoc supports 2.2 syntax already

kamranayub:
So, I made a sample that compiles and runs successfully in the TS playground using Mixins that supports inheritance:

```ts
class Class {
	classProp = 'foo'
}

type Constructor<T> = new (...args: any[]) => T;

function Defaultable<TProps, T extends Constructor<{}>>(Base: T, defaults: TProps) {
	return (class extends Base {
		defaults: Partial<TProps>;

		public setProperties(props?: Partial<TProps>): this {

			var mergedProperties = Object.assign({}, this.defaults, props);

			for (var k in mergedProperties) {
				if (typeof <any>this[k] !== 'function') {
					(<any>this)[k] = mergedProperties[k];
				}
			}
			return this;
		}

		public resetProperties() {
			this.setProperties(this.defaults);
		}
		
        constructor(...args: any[]) {
			super(...args);

			this.defaults = defaults;
			this.setProperties(args[0]);
        }
    })
}

interface ActorProps {
	width: number;
	height: number;
}

const DefaultActorProps: ActorProps = {
	width: 0,
	height: 0
}

class Actor extends Defaultable(Class, DefaultActorProps) implements ActorProps {
	width: number;
	height: number;

	constructor(props: Partial<ActorProps>) {
		super();		
	}
}

var actor = new Actor({ height: 100 });

console.log('width', actor.width);
console.log('height', actor.height);
console.log('classprop', actor.classProp);

interface PlayerProps extends ActorProps {
	playerName: string
}

const DefaultPlayerProps: PlayerProps = { playerName: '', ...DefaultActorProps }

class Player extends Defaultable(Actor, DefaultPlayerProps) implements PlayerProps {
	playerName: string

	constructor(props: Partial<PlayerProps>) {
		super(props);
	}
}

var player = new Player({ width: 100, height: 100, playerName: 'Kamranicus' });

console.log('player width', player.width);
console.log('player height', player.height);
console.log('player classprop', player.classProp);
console.log('playerName', player.playerName);
```

The issue is, it requires some tricks to get the props to cascade--such as making sure to reference the extended object's Props interface in the extending Props interface (`PlayerProps extends ActorProps`). Still, it *does* work which is kind of compelling.

alanag13:
I forgot to include a implementation example with my latest push. Here's basically how this would look:

#### Changing existing Ex classes to support this feature

```typescript
//1. Rename the existing class to something else
//2. remove its export so that is no longer available on the ex namespace
//3. add a constructor that takes an argument of Partial<T>
export class InternalActor extends Class {
    ...
    
    constructor(options?: Partial<InternalActor>)
     //... resolve ambiguities with existing constructors...
 }

//create a new class named the same as the existing class (before it was renamed)
//that extends the new `Configurable` type
export class Actor extends Configurable(InternalActor){
    constructor(options?: Partial<Actor>){
        super(options);
    }
 } 
```

#### Using this feature in custom classes 
Simply continue to use the classes as you normally would, only now the constructor option bag is available to you. If you want to enable this on your custom class (to have the typing suggest all of your classes custom properties), all you have to do is define the constructor yourself:

```typescript
export class MyCustomActor extends Actor {
    constructor(options?: Partial<MyCustomActor>){
        super();
    }
}
```

Here's a list of commonly used classes I'm thinking would benefit most from implementing this on. Let me know if any of you think something's missing.

- Actor
- Animation
- Label
- Scene
- Sprite
- SpriteSheet
- Sound
- Texture
- UIActor

alanag13:
I've removed `Trigger` from the implementation list since @eonarheim is working on that.
I also discovered we can't do this to `Sound` unless we change something there since its ctor takes a rest parameter.

alanag13:
This should be good to go, at long last! Give it a review whenever you're able to. A couple notes:
- Parameter requirements are the same. If an existing ctor had a required variable, that same variable needs to be included in the option bag. Optional variables and all public properties not used in ctors are optional. 
- You also get an `assign` function that provides the same functionality for post-initialization.
- I got rid of the "reset to default" functionality. It would have required a lot of maintenance and any new classes that would have wanted to use it would have required a lot of setup. You can achieve a similar effect by just doing something like `actor.assign(new Actor())`, but private variables would not change.
- I updated the sandbox game to use the option bag implementations in several cases. Some "old" implementations are also left in to demonstrate that both ways still work. I also left any "old" versions that were changed commented so that they can be easily toggled back to demonstrate this.

Questions:
- **Tests**: :man_shrugging: I can't think of a super efficient way to do this other than to copy the existing tests and replace the implementations with option bags. Is that what I should do, or should I just make some generic tests for `Configurable`?
- **Docs**: I'd like to review once #919 gets merged to make sure this all looks good. Also, should we add jsdocs headers for the new constructors? If so, what I'm guessing we can probably use similar verbiage for all of them. What do we want that to be?

I also need to update the changelog. I'm thinking I would just list the classes where this available and provide a short example. Let me know your thoughts.

eonarheim:
@excaliburjs/core-contributors Please review, this PR is huge! 

Things to check:
* [ ] - Visual tests
* [ ] - Additional documentation for new constructors?
* [ ] - Documentation generation (looks okay?)
* [ ] - Code style

src/engine/Docs/Constructors.md
```diff
@@ -0,0 +1,61 @@
...
+```typescript
+var actor = new ex.Actor({
+   pos: new ex.Vector(1,2)
+})
+var actorClone = new ex.Actor(actor);
+
+expect(actor.pos).toBe(actorClone.pos); // true;
+```
+
+Types that support option bags can be mass assigned.
```
alanag13: "can have their properties mass assigned using the `assign` method"

PR #759, opened 2017-02-24
#591

### LINKED ISSUE #591: potential confusion between different engine dimension parameters
The difference between "the height of the engine's drawing surface" (`engine.getHeight()`) and "the height of the game canvas in pixels" (`engine.height`) is not immediately obvious, both in their names and their descriptions in the API .

(This issue also applies to `engine.width` and `engine.getWidth()`, but height is used in the examples below)

These dimensions are usually the same, unless the camera has been zoomed in or out, whereby `engine.getHeight()` is reduced/increased based on the zoom amount.
##### Proposed Changes
- Change `engine.height` to `engine.canvasHeight` and `engine.width` to `engine.canvasWidth`
- Update the documentation to more clearly explain the differences between each pair of parameters

rehlert95:
Closes #591 
## Changes:

- renames engine.width to engine.canvasWidth and engine.height to engine.canvasHeight

### DIFF
--- a/src/engine/Engine.ts
+++ b/src/engine/Engine.ts
@@ -48,12 +48,12 @@ export interface IEngineOptions {
    /**
     * Optionally configure the native canvas width of the game
     */
-   width?: number;
+   canvasWidth?: number;
 
    /**
     * Optionally configure the native canvas height of the game
     */
-   height?: number;
+   canvasHeight?: number;
 
    /**
     * Optionally specify the target canvas DOM element to render the game in
@@ -113,11 +113,11 @@ export class Engine extends Class {
    /**
     * The width of the game canvas in pixels
     */
-   public width: number;
+   public canvasWidth: number;
    /**
     * The height of the game canvas in pixels
     */
-   public height: number;
+   public canvasHeight: number;
 
    /**
     * Access engine input like pointer, keyboard, or gamepad
@@ -232,8 +232,8 @@ export class Engine extends Class {
     * Default [[IEngineOptions]]
     */
    private static _DefaultEngineOptions: IEngineOptions = {
-      width: 0,
-      height: 0,
+      canvasWidth: 0,
+      canvasHeight: 0,
       canvasElementId: '',
       pointerScope: Input.PointerScope.Document,
       suppressConsoleBootMessage: null,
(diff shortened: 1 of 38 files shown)


src/engine/Engine.ts
```diff
@@ -232,8 +232,8 @@ export class Engine extends Class {
     * Default [[IEngineOptions]]
     */
    private static _DefaultEngineOptions: IEngineOptions = {
-      width: 0,
-      height: 0,
+      canvasWidth: 0,
```
eonarheim: Let's avoid changing the IEngineOption property. It keeps the constructor option back terse and we don't need to change every file.

I think it's more important when accessing `ex.Engine.canvasWidth` and `ex.Engine.canvasHeight` which gives us the canvas pixel width/height  vs  `ex.Engine.getWidth()` and `ex.Engine.getHeight()` which get's the current drawing surface in pixels taking zoom into account.

Perhaps it would also be good to further disambiguate the different heights in #591, and refactor `ex.Engine.getWidth()` and `ex.Engine.getHeight()` to be something like `getScreenWidth()`,`getPixelWidth()`. @jedeen What do you think?

Other than that I like the PR.

src/engine/Engine.ts
```diff
@@ -321,15 +321,15 @@ O|===|* >________________>\n\
...
          if (options.displayMode === undefined) {
             this.displayMode = DisplayMode.Fixed;
          }
-         this._logger.debug('Engine viewport is size ' + options.width + ' x ' + options.height);
-         this.width = options.width;
-         this.canvas.width = options.width;
-         this.height = options.height;
-         this.canvas.height = options.height;
+         this._logger.debug('Engine viewport is size ' + options.canvasWidth + ' x ' + options.canvasHeight);
+         this.canvasWidth = options.canvasWidth;
```
eonarheim: 👍

src/engine/Engine.ts
jedeen: I agree we should avoid changing the IEngineOption. 

I wouldn't recommend `getScreenWidth()`, as that could potentially be mistaken for something else (e.g. the width of the browser window or actual screen dimensions). What about`getCanvasWidth()` for "the width of the game canvas in pixels" and `getPixelWidth()` for "the width of the engine's drawing surface"?

kamranayub:
Or getDrawWidth?

src/engine/Engine.ts
jedeen: @kamranayub's comment didn't thread, but I like his suggestion of `getDrawWidth()` over `getPixelWidth()`.

eonarheim: @rehlert95 @jedeen I think we have consensus on renaming `getWidth()` and `getHeight()` `getDrawWidth()` and `getDrawHeight()` as well as not changing `IEngineOptions`

eonarheim:
@rehlert95 Thanks!

PR #760, opened 2017-03-03
[#750] Add Engine's Running Status Check

### LINKED ISSUE #750: Add utility method to determine if Engine is paused or not
### Context
Currently, you can stop/start (i.e. pause/resume) a game with `Engine.start()` and `Engine.stop()`. However, in order to implement pause logic, you need to manage your own boolean:

```typescript
var paused = false;
game.on('p', () => {
   if (!paused) {
      game.stop();
   } else {
      game.start();
   }
   paused != paused;
});
```

### Proposal
The `Engine` already has a private `_hasStarted` variable that appears to be maintaining, but not exposing, this status.

It would be helpful to have an `isStopped()` or `isPaused()` method to return the current state of the game, so the above code could be simplified to:

```typescript
game.on('p', () => {
   if (!game.isPaused()) {
      game.stop();
   } else {
      game.start();
   }
});
```

SurajGoel:
Closes #750 

## Changes:

- Added a function isPaused to check the running status of Engine into Engine.ts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [0.9.0] 2017-02-09
 
 ### Added
+- Added `Engine.isPaused` to retrieve the running status of Engine ([#750](https://github.com/excaliburjs/Excalibur/issues/750))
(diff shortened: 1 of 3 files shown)


eonarheim:
@SurajGoel Thanks!

PR #769, opened 2017-03-28
[#762] Upgrade project to typescript 2.2

### LINKED ISSUE #762: Update TypeScript to version 2.2
### Context
[TypeScript 2.2](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html) has been released with support for mix-in classes, which would simplify the work needed for #410.

### Proposal
- Update TypeScript to 2.2
- Update TypeDoc

alanag13:
Closes #762 

## Changes:

- upgrade ex to typescript 2.2.2


alanag13:
`Color` had to be modified due to what appears to be a new compilation rule -- methods must be defined before they are used.

alanag13:
is there anything re: docs or anything like that that I might not be privy to that has to also be done as a result of this change?

eonarheim:
@alanag13 We probably should do a doc site push with ts 2.2 support as well

kamranayub:
Yeah just need to upgrade the site to latest typedoc and make sure it
compiles this branch, I usually do it by editing the docs.js file and
hardcode the branch for testing locally.

kamranayub:
Site works with 2.2 👍

PR #772, opened 2017-03-31
[#770] Fix drawings on rotated and scaled actors

### LINKED ISSUE #770: Actors when scaled don't appear to rotate as expected
### Steps to Reproduce
When using sprites inside of actors while setting actor scale, the actor's current image does not appear to rotate as expected.

See full repro code below:
```typescript
var game = new ex.Engine({
    width: 800,
    height: 600
});


// create an asset loader
var loader = new ex.Loader();
var resources = {
    txPlayer: new ex.Texture("assets/logo.png")
};

// queue resources for loading
for (var r in resources) {
    loader.addResource(resources[r]);
}

var target = new ex.Actor(500, 500, 20, 20, ex.Color.Red);
game.add(target);

var aim = new ex.Actor(200, 200, 100, 100, ex.Color.Black);
aim.addDrawing(resources.txPlayer);
aim.setHeight(10);
aim.scale.setTo(1, .5);
game.add(aim);

// uncomment loader after adding resources
game.start(loader).then(() => {

    game.input.pointers.primary.on('move', (ev: ex.Input.PointerEvent) => {
        target.pos.setTo(ev.x, ev.y);

        var aimVec = target.pos.sub(aim.pos);
        aim.rotation =  aimVec.toAngle();
    });

});
```

The current work around is to scale the sprite separately instead of the actor.

`actor.currentDrawing.scale.setTo(1, .5)`

### Expected Result
The excalibur logo while scaled to .5 height should point at the cursor

works

### Actual Result
The excalibur logo kind of points at the cursor, but is scaled oddly as it rotates around

notworks

### Proposal

Fix rotation with scaled actors to behave as expected.

eonarheim:
@excaliburjs/core-contributors I've reviewed the visual tests. Please review on your machines against your games before merging, since this is a core draw loop change I want to make sure that more than just the visual tests have been checked

Closes #770 

## Changes:

- Update the order of the affine transformations, if you scale before rotating the scale will not be applied along the new rotated coordinates. Rotations must occur before scaling so that the x,y scale are applied in the correct direction. Unfortunately this was a very subtle bug thanks to @HSAR for bringing this to our attention
- Added visual test to the sandbox/tests/scale
- Added ScaleSpec to our test suite to do image diff testing


src/spec/ScaleSpec.ts
```diff
@@ -0,0 +1,58 @@
...
+   });
+
+   afterEach(() => {
+      engine.stop();
+   });
+	
+   
+   it('is drawn coorectly scaled at 90 degrees', (done) => {            
+
+      let game: ex.Engine = TestUtils.engine({ width: 800, height: 600 });
```
kamranayub: The engine is already availabile above, either remove the one that gets created in `beforeEach` or use that one so we don't have two running at once.

kamranayub:
I will test with [redacted] Golf.

eonarheim:
@kamranayub updated per feedback

Issue #777, opened 2017-04-08
100% automated release process


kamranayub:
### Context
We are very close to automating the release process for Excalibur.

### Proposal
**Goal:** To cut a new release, push new GitHub release tag.

- [x] Excalibur repo should pull GH releases for latest stable (env var?)
- [x] Excalibur repo should build API docs
- [x] Docs should build at latest release tag AND edge
- [x] Docs should query GH for latest release information for homepage (env var?) on every build
  - [x] BONUS: At build time, calculate file size and gzip size for minified release for homepage
- [x] Excalibur repo should inject GH releases into documentation TOC somehow (env var?)
- [x] On release tag, should trigger excalibur site build
- [ ] On release tag, automation should publish a corresponding GH release in excalibur-dist repo. It should use the tagged version number.
- [ ] On excalibur-dist release tag, CI should publish Nuget package
- [ ] On excalibur-dist release tag, CI should publish npm package
- [ ] On excalibur-dist release (alpha or release) dependent plugins/repos should automatically be updated
   - [ ] Excalibur Tiled
   - [ ] Excalibur Electron
   - [ ] Documentation
   - [ ] Samples

### Notes

- Assumes manual work has been done (pre-validation, merging PRs, etc.)

kamranayub:
For Typedoc, it would be best to generate docs inside this repository but have the site repo publish the docs. That way, we always ensure the docs can build in CI and use the right version, but the site repo can slurp them up and publish them to the appropriate location (tag or edge)

Tankobot:
@kamranayub Question having to do with 8eec7052717bf1b9385752ddc3c90edf0a0f63dc : how do you,

> use local package version placeholder for quick builds

I'm fairly new to TypeScript, Node, etc. and I was hoping that I could use the git repository for one of my dependencies so I could make changes to the code base as I work on a game if necessary. When I try to install/add the local copy as a dependency, npm complains about no version being present in the package.json. 

I ended up adding an arbitrary version to package.json and it seemed to install correctly. Is this the preferred method, or did I misinterpret the instructions? Also, this might be good information to include in the docs unless I just missed something.

*I'm not too familiar with the etiquette around here, so I'll end with "Thank You".*

kamranayub:
@Tankobot Good question! 

## Using npm

With NPM, I *believe* you can use a [feature](https://docs.npmjs.com/cli/install) of using the Git URL as the package name:

    npm install <git repo url>

[...]

PR #783, opened 2017-04-15
[#777] Use GH latest release version

kamranayub:
Related #777

## Changes:

- On local builds, use local package version placeholder for quick builds
- On CI builds, use GitHub release version. Fails the build if API request fails.

This change allows us to remove all specific versioning from the repository, relying on GitHub as the source of truth for all releases. As a side effect, dist files get a better version anyway--previously they only had the latest release (v0.10.0) and date. Now they will get the commit and Travis build number.


kamranayub:
Merge #782 and then I can merge that into this PR to fix build.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "excalibur",
   "private": true,
   "description": "Excalibur.js is a simple JavaScript game engine with TypeScript bindings for making 2D games in HTML5 Canvas. Our mission is to make web game development as simple as possible.",
-  "version": "0.10.0",
+  "version": "0.0.1",
```
eonarheim: Why did we back the verion up to 0.0.1?

kamranayub:
Because this isn't a real version and I wanted to make that clear. Local
builds will use this only, production builds use GH release.

kamranayub:
The only place this will be confusing is for people browsing the repo. If
there's a clearer semver that communicates this is a fake version, I'm up
for it. I think npm requires a version, hence why I didn't delete the
property otherwise I would've.

eonarheim:
@kamranayub can we do a piece of non-valid semver then? like `0.0.local` or something?

kamranayub:
Maybe, I will check. I thought npm would fail if it couldn't parse
version...

kamranayub:
This is ready. Appveyor PR build fails due to missing token, but will be there once merged.

    Error: Missing environment variable GH_TOKEN

PR #785, opened 2017-04-18
Update deprecated fps method call

jedeen:
Debug mode was using a deprecated method when drawing FPS to the screen


eonarheim:
@jedeen @kamranayub Should we add convenience alias for fps somewhere? It's much harder to get at that value now

kamranayub:
I'd say not really. It's not often you need to draw FPS yourself, usually we do it for people... and if you are doing debug stuff, all the info is off of current/previous frame stats which makes it easy.

eonarheim:
@kamranayub Cool cool, you've convinced me ;) 

🚢 IT!

PR #786, opened 2017-04-18
Fix API docs build

kamranayub:
Related #777 

## Changes:

- Install SASS gem
- Run API docs build at end of build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ cache:
   - node_cache
   - node_shrinkwrap
   - $HOME/.npm
+before_install:
+- gem install sass
 script:
 - node_modules/.bin/grunt travis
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -8,6 +8,8 @@ cache:
   - node_cache
   - node_shrinkwrap
   - $HOME/.npm
+before_install:
+- gem install sass
```
eonarheim: Do people need ruby installed to build Excalibur?

eonarheim: Or is this travis only

kamranayub:
Only if they want to build the API docs, it's part of the default Typedoc
theme. It's not something everyone will do I don't think.

PR #792, opened 2017-04-27
Add window positioning as an Engine option

### LINKED ISSUE #781: Add engine option to allow positioning the canvas
### Context

Right now you can position the game canvas using CSS, however not everyone may know how to do this effectively, especially if they want to vertically center the canvas.

### Proposal

Add an engine option `canvasPosition` that will add some styles to position the canvas:
- top / bottom / middle
- left / right / center
- absolute value (CSS object, `{ top: '4px', left: '50%', right: 6, bottom: '4em' }`)

Can use an enumeration for the positional values. The CSS object should treat strings as raw values to apply to the styles. Numbers should be interpreted in pixels.

Flexbox can be used to position the canvas according to the layout needed, while `position: absolute` can be used for the absolute positioning. Flexbox may require a parent wrapper element to be generated OR we can just apply the styles to whatever the parent element is for the game canvas.

This is entirely **optional** and advanced users will be expected to use their own CSS to customize the game canvas.

andrewmbyrd:
Closes #781 

## Changes:

- Added another DisplayMode option: `DisplayMode.Position`. When this is selected as the displayMode type, the user must specify a new `position` option
- Added a new Interface `AbsolutePosition` which can describe the `position` option. A `string` can also describe `position`

##  Note
I've written some Jasmine tests for this issue, as well as made sure that this didn't break any tests. (there were 8 tests that were failing before I implemented my changes, looks like HTML property changing tests).

I've updated the docs in the repo, but it may be good to document on [the external docs site](https://excaliburjs.com/docs/api/v0.10.0/index.html). I'm happy to do that if need be!

Note also that I don't believe the Engine display mode was reliably being set before. There were references to `options.displayMode`, but `this.displayMode` was only set as a default. So, I've added lines 773-775 as you can see below.

### DIFF
--- a/sandbox/tests/scene/display-mode-position.html
+++ b/sandbox/tests/scene/display-mode-position.html
@@ -1,0 +1,42 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>Bloc Breaker</title>
+    <link rel="stylesheet" type="text/css" href="display-mode-position.css">
+  </head>
+  <body>
+    
+    <section class="buttons">
+      <div class ="btn top-left">Put game at top left</div>
+      <div class ="btn mid-right">Put game at middle right</div>
+      <div class ="btn mid-center">Put game at middle center</div>
+      <div class ="btn bottom-left">Put game at bottom left</div>
+      <div class ="btn top-right">Put game at top right</div>
+      <div class ="btn top-center">Put game at top center</div>
+    </section>
+    
+    <div class="absolute-input">
+      <p>You can set exactly where the game will be presented by giving absolute positioning in the boxes below</p>
+      <p> Use <strong>Strings</strong> like 4px and 50%</p>
+      <div>
+          Top: <input type="text" id="top" name="top">
+      </div>
+      <div>
+          Right: <input type="text" id="right" name="right">
+      </div>
+      <div>
+          Bottom: <input type="text" id="bottom" name="bottom">
+      </div>
+      <div>
+          Left: <input type="text" id="left" name="left">
+      </div>
+      <div class="button">
+        <button type="submit">Draw the game at these coordinates</button>
+      </div>
+    </div>
+    
+    
+    <script src="../../../build/dist/excalibur.min.js"></script>
+    <script src="display-mode-position.js"></script>
+  </body>
+</html>
--- a/sandbox/tests/scene/display-mode-position.html
+++ b/sandbox/tests/scene/display-mode-position.html
@@ -36,7 +36,7 @@
     </div>
     
     
-    <script src="../../../build/dist/excalibur.min.js"></script>
+    <script src="../../excalibur.js"></script>
     <script src="display-mode-position.js"></script>
   </body>
 </html>
--- a/src/engine/Engine.ts
+++ b/src/engine/Engine.ts
@@ -50,6 +50,7 @@ export enum DisplayMode {
 /*
 * Interface describing the absolute CSS position of the game window. For use when DisplayMode.Position
 * is specified and when the user wants to define exact pixel spacing of the window.
+* When a number is given, the value is interpreted as pixels
 */
 export interface IAbsolutePosition {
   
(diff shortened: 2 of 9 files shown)


kamranayub:
I need to patch the version script to support community PR builds

eonarheim:
@andrewmbyrd Apologies our community PRs were busted, after this next build we should know more :)

andrewmbyrd:
no problem!

andrewmbyrd:
I'll look into adding some visual tests! How do I determine how TSLint is failing? I didn't get any notifications on my machine (I'm using Atom on Windows)

kamranayub:
You can use "npm run all" to run all linting and tests locally. The README
has some other commands you can run for different purposes.

andrewmbyrd:
Ah, I see. I ran each individually, wasn't aware that npm run all would do the Linting tests, sorry about that! Linting is all passing now. I added an html and js file to `sandbox/tests/scene` that allows you to click buttons to place the game window in various positions.

sandbox/tests/scene/display-mode-position.html
```diff
@@ -0,0 +1,42 @@
...
+          Left: <input type="text" id="left" name="left">
+      </div>
+      <div class="button">
+        <button type="submit">Draw the game at these coordinates</button>
+      </div>
+    </div>
+    
+    
+    <script src="../../../build/dist/excalibur.min.js"></script>
+    <script src="display-mode-position.js"></script>
```
eonarheim: The source file `display-mode-position.ts` is missing from the PR

eonarheim: Let's add a helper link to this test in the index.html located in the root of sandbox as well

image

src/engine/Engine.ts
```diff
@@ -83,6 +101,14 @@ export interface IEngineOptions {
...
     */
    suppressMinimumBrowserFeatureDetection?: boolean;
+   
+   /*
+   *Specify how the game window is to be positioned when the DisplayMode.Position is chosen. This option MUST be specified
+   *if the DisplayMode is set as DisplayMode.Position. The position can be either a string or an AbsolutePosition. String must be in the
+   *format of css style background-position. 
+   */
+   
+   position?: string | IAbsolutePosition;
```
eonarheim: Let's add a few examples to the documentation of valid positions, since diverse types are valid it may be good to give some examples.

src/engine/Engine.ts
```diff
@@ -754,8 +788,80 @@ O|===|* >________________>\n\
...
+              
+              this.canvas.style.display = 'block';
+              this.canvas.style.position = 'absolute';
+              
+              if (typeof options.position === 'string') {
+                var specifiedPosition = options.position.split(' ');
+                
+                switch (specifiedPosition[0]) {
+                  case 'top':
+                    this.canvas.style.top = '1%';
```
eonarheim: What is the reason for the 1%? Would `this.canvas.style.top = '0px';` be simpler?

sandbox/tests/scene/display-mode-position.html
andrewmbyrd: Ah, I had created a .js file, but it looks like that didn't get pushed up. I've made a .ts file now!

src/engine/Engine.ts
andrewmbyrd: Examples are in there now

src/engine/Engine.ts
andrewmbyrd: No particular reason, I just personally though having a little border around the edge looked better :) I've set it to 0px now!

andrewmbyrd:
Yep, resizing works fine. It's based on absolute positioning so it's always looking at the window size

eonarheim:
@andrewmbyrd Looks good to me! Thanks for helping out! 👍 

@kamranayub Any updates on your end?

src/engine/Engine.ts
```diff
@@ -83,6 +102,16 @@ export interface IEngineOptions {
     * browsers or if there is a bug in excalibur preventing execution.
     */
    suppressMinimumBrowserFeatureDetection?: boolean;
+   
+   /*
+   *Specify how the game window is to be positioned when the DisplayMode.Position is chosen. This option MUST be specified
```
kamranayub: Can you add a space after each `*` on each line?

sandbox/tests/scene/display-mode-position.html
```diff
@@ -0,0 +1,42 @@
...
+      <div>
+          Left: <input type="text" id="left" name="left">
+      </div>
+      <div class="button">
+        <button type="submit">Draw the game at these coordinates</button>
+      </div>
+    </div>
+    
+    
+    <script src="../../../build/dist/excalibur.min.js"></script>
```
kamranayub: Can you change this to `../../excalibur.js`? It should reference a file that gets copied to `sandbox`.

kamranayub:
Once this merges and works, we can finish the PR,

kamranayub:
Thanks @andrewmbyrd! Merged 👍

PR #802, opened 2017-05-04
[#667] Add detection for multi-actor body contact

### LINKED ISSUE #667: Add helper properties on body for touching and wasTouching
#### Context

I would like an easy way to determine whether something was touching something else in the previous frame and is now not touching. Phaser has this functionality.

This could be enabled by adding more collision events: before, during (current one we have), after.
#### Proposal

Add `this.body.touching(actor)` and `this.body.wasTouching(actor)`

andrewmbyrd:
Closes #667 

## Changes:

- Add `touching(Actor)` method to `Body`
- Add `wasTouching(Actor, Engine)` method to `Body`

`wasTouching` works when two bodies bounce off of each other, not when they remain in stationary contact (even if one body is then killed).

Previous frame data was acquired by adding a collidersHash to `Engine.stats.prevFrame.physics`. If the pair is in that hash, and is not currently `touching`, then `wasTouching` will return `true`.


src/engine/Debug.ts
```diff
@@ -9,6 +10,14 @@ export interface IDebugStats {
 }
 
 /**
+ * Hash containing the [[Pair.id]]s of pairs that collided in a frame
+ */
+export interface ICollidersHash {
+   [pairID: string]: Pair;
```
eonarheim: Let's stick to pascal case "pairId" to remain consistent with the codebase

src/engine/Debug.ts
```diff
@@ -292,6 +306,7 @@ export class FrameStats implements IFrameStats {
 export class PhysicsStats implements IPhysicsStats {
    private _pairs: number = 0;
    private _collisions: number = 0;
+   private _collidersHash: ICollidersHash = {};
```
eonarheim: I like typing the hash with an interface 👍 probably going to start doing this in the future especially on intereseting type signatures :)

src/spec/CollisionSpec.ts
```diff
@@ -36,5 +36,19 @@ describe('A Collision', () => {
...
+       if (actor1.body.touching(actor2)) {
+         touching = true;
+       }
+     });
+     
+     scene.update(engine, 20);
+     expect(touching).toBe(true);
+   });
+   
+
```
eonarheim: Can we also add a test for wasTouching?

sandbox/sounds/html/healthbar.tmpl
```diff
@@ -0,0 +1,8 @@
+<div style="position:absolute; top: 400px; left: 300px">
```
eonarheim: This file may have been copied and included by mistake in the sounds folder

sandbox/sounds/html/index.html
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html>
```
eonarheim: This file may have been copied and included by mistake in the sounds folder as well

sandbox/tests/collision/touching.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
```
eonarheim: Visual test looks great! Let's add a link to it in sandbox/index.html 👍

src/engine/Debug.ts
eonarheim: Err, I mean camel case 🐙

andrewmbyrd:
I've updated the small changes as requested.

I was having the hardest time figuring out why the `wasTouching` spec wouldn't pass. Our `wasTouching` method requires a collision to have happened and that the engine updates its `prevFrame` debug stats with that information. If you run the test I've set up for `wasTouching` (which is failing intentionally), you'll note that in the test environment, prevFrame collisions are always 0, even though currFrame collisions vary. I'm not sure why this is happening...

eonarheim:
@andrewmbyrd Odd, I wonder if there is something wrong with the engine stats functionality in the test context. Or perhaps it's a bug with the engine stats entirely!

sandbox/index.html
```diff
@@ -26,6 +26,7 @@
       <li><a href="tests/label/label.html">Label</a></li>
       <li><a href="tests/label/fonts.html">Label - Fonts</a></li>
       <li><a href="tests/collision/index.html">Collision</a></li>
+      <li><a href="tests/collision/touching.html">Touching</a><li>
```
eonarheim: The <li> at the end of the line should be a </li>

eonarheim:
Downloaded your fork, definitely feels like a bug in stats, been going through this and _it should_ be working as you expected. Let me see if I can narrow it down...

sandbox/index.html
andrewmbyrd: ah, sorry about that! got it updated

eonarheim:
@andrewmbyrd I found the issue, you were right the test context was the problem. The `mock.engine` doesn't have the implementation that stats relies on to be correct. The stats recycler only exists in the `createMainLoop` method.  The correct way to fix this would be to extract this stats stuff out of this section and into a separate test friendly place. I'll create a separate issue for this.

```typescript
public static createMainLoop(game: Engine, raf: (func: Function) => number, nowFn: () => number) {
....
            // reset frame stats (reuse existing instances)
            var frameId = game.stats.prevFrame.id + 1;
            game.stats.prevFrame.reset(game.stats.currFrame);
            game.stats.currFrame.reset();
            game.stats.currFrame.id = frameId;
            game.stats.currFrame.delta = delta;
            game.stats.currFrame.fps = 1.0 / (delta / 1000);
....
}
```

This hits a deeper issue that I think is out of scope for this PR, we really should re-evaluate the `mock.engine` in the broader test landscape.

Let's keep your test for later, but deactivate it for now, the way to do that is by putting an `x` in front, like so

```typescript
xit('should recognize when bodies were touching last frame', () => {
  var wasTouching = 0;
...
```

src/spec/CollisionSpec.ts
```diff
@@ -36,5 +36,57 @@ describe('A Collision', () => {
       scene.update(engine, 20);
       expect(numCollisions).toBe(2);
    });
+   
+   it('should recognize when actor bodies are touching', () => {
```
eonarheim: Let's `x` out the test for now and I'll approve the PR 👍

src/spec/CollisionSpec.ts
```diff
@@ -36,5 +36,57 @@ describe('A Collision', () => {
...
+         wasTouching++;
+       }
+       
+     });
+     
+     var ms = 0;
+     while (wasTouching < 2) {
+       scene.update(engine, 10);
+       console.log(`current frame collisions: ${engine.stats.currFrame.physics.collisions}`);
+       console.log(`previous frame collisions: ${engine.stats.prevFrame.physics.collisions}`);
```
eonarheim: Also let's comment out the `console.log` ;)

src/spec/CollisionSpec.ts
eonarheim: Everything looks like it's working as I expect, thanks for tackling this one!

kamranayub:
Mock engine should not be used in favor of using the actual engine which is now testable... only old tests should use the mock, I think. See [ScaleSpec](https://github.com/excaliburjs/Excalibur/blob/master/src/spec/ScaleSpec.ts) for an example of how.

andrewmbyrd:
So here's an interesting situation: I've updated the `wasTouching` test to look like this: 

```
it('should recognize when bodies were touching last frame', () => {
     var wasTouching = 0;
     var engine1 = TestUtils.engine({width: 200, height: 200});
     
     var actor3 = new ex.Actor(0, 0, 10, 10);
     var actor4 = new ex.Actor(30, 0, 10, 10);
     
     engine1.start().then(() => {
       actor3.vel.setTo(200, 0);
       actor3.collisionType = ex.CollisionType.Elastic;
       actor4.collisionType = ex.CollisionType.Elastic;
       engine1.add(actor3);
       engine1.add(actor4);
       
       actor3.on('update', function() {
         console.log(`current frame collisions: ${engine1.stats.currFrame.physics.collisions}`);
         console.log(`previous frame collisions: ${engine1.stats.prevFrame.physics.collisions}`);
         if (actor3.body.wasTouching(actor4, engine1)) {
           wasTouching++;
           engine1.stop();
         }
         
       });
       
     });
  
     expect(wasTouching).toBe(1);
     
   });
```

I know that the `wasTouching` gets set to true at some point because if it didn't, `engine1.stop()` would never be called and we would continue to see random log messages throughout the test suite run. The problem is that the tests continue to run while the game is running, it just takes a couple of seconds for the collision to happen. Without fine-tuning the collision to occur at just the right time, is there a way to halt the testing process until the collision is detected?

eonarheim:
@andrewmbyrd Sorry about the delay I was out on vacation 🏖, we can do an asynchronous test to accomplish this.

If you take a `done` argument to the spec, Jasmine will pass a function that it will wait for before continuing the rest of the test suite.
```typescript
it('should recognize when bodies were touching last frame', (done) => {

...
        if (actor3.body.wasTouching(actor4, engine1)) {
           wasTouching++;
           engine1.stop();
           done(); // jasmine will wait to continue until this is called, or fail the test after a timeout
         }
...

});
```

kamranayub:
@eonarheim Ready to merge now with fix?

src/engine/Collision/Body.ts
```diff
@@ -228,4 +231,44 @@ export class Body {
...
+     if (pair.collision) {
+       return true;
+     }
+     
+     return false;
+   }
+   
+   /**
+    * Returns a boolean indicating true if this body COLLIDED with 
+    * the body of the other Actor in the last frame, && they are no longer touching
```
kamranayub: Change `, &&` to `and`

PR #806, opened 2017-05-09
[#801] Add per-play sound volume option

### LINKED ISSUE #801: Add ex.Sound per-play volume setting
### Context
It is slightly clumsy to play sounds at different volumes dynamically. It would be nice to have a new way to play a sound on demand at a specific volume. Currently you need to set the `ex.Sound.volume` property then call `.play()` then set the `.volume` property for every play that you want at a different volume. We should also be able to dynamically change volume while playing.

### Proposal

Add a `ex.Sound.play(volume)` or a `ex.Sound.playAt(volume)` method for per play volume settings and should return a promise that resolves with the sound play is done..

```typescript
var sound = new ex.Sound();
sound.playAt(.5); // play once at a specific maximum volume
```

andrewmbyrd:
Closes #801 

## Changes:

- Added optional volume argument to `Sound.play(volume?: number)`, which will play the Audio file
+at anywhere from mute (`volume` is 0.0) to full volume (`volume` is 1.0).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Breaking Changes
 ### Added
+- Added optional volume argument to `Sound.play(volume?: number)`, which will play the Audio file
+at anywhere from mute (`volume` is 0.0) to full volume (`volume` is 1.0).
 ### Changed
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Breaking Changes
 ### Added
+- Added optional volume argument to `Sound.play(volume?: number)`, which will play the Audio file
```
kamranayub: Can you add a link to the GH issue?

eonarheim:
@andrewmbyrd Thanks again for helping out!

PR #809, opened 2017-05-12
[#808] Add PointerWheel event support

### LINKED ISSUE #808: Add support for wheel events
### Context
Currently, Excalibur does not directly support wheel events (such as mouse scrolling) for input.

### Proposal
Add support for [wheel events](https://developer.mozilla.org/en-US/docs/Web/Events/wheel). 

Additionally, work will need to be done for the scenario where a game is using wheel events, but the page has a scroll bar.

mrkmg:
Closes #808 

## Changes

- Added event "wheel"
- Added sandbox pointer wheel test
- Added docs

### DIFF
--- a/src/engine/Docs/Pointers.md
+++ b/src/engine/Docs/Pointers.md
@@ -26,6 +26,17 @@ engine.input.pointers.primary.on("move", function (evt) { });
 engine.input.pointers.primary.on("cancel", function (evt) { });
 ```
 
+### Wheel Event
+
+You can also subscribe to the mouse wheel event through `engine.input.points.on`. A [[WheelEvent]]
+object is passed to your handler which offers information about the wheel event being received.
+
+- `wheel` - When a mousewheel is activated (trackpad scroll or mouse wheel)
+
+```js
+engine.input.pointers.primary.on("wheel", function (evt) { });
+```
+
 ## Last position querying
 
 If you don't wish to subscribe to events, you can also access the [[Pointer.lastPagePos]], [[Pointer.lastScreenPos]]
--- a/src/engine/Docs/Pointers.md
+++ b/src/engine/Docs/Pointers.md
@@ -28,13 +28,13 @@ engine.input.pointers.primary.on("cancel", function (evt) { });
 
 ### Wheel Event
 
-You can also subscribe to the mouse wheel event through `engine.input.points.on`. A [[WheelEvent]]
+You can also subscribe to the mouse wheel event through `engine.input.pointers.on`. A [[WheelEvent]]
 object is passed to your handler which offers information about the wheel event being received.
 
 - `wheel` - When a mousewheel is activated (trackpad scroll or mouse wheel)
 
 ```js
-engine.input.pointers.primary.on("wheel", function (evt) { });
+engine.input.pointers.on("wheel", function (evt) { });
 ```
 
 ## Last position querying
--- a/src/engine/Engine.ts
+++ b/src/engine/Engine.ts
@@ -47,6 +47,24 @@ export enum DisplayMode {
    Position
 }
 
+/**
+ * Enum representing the different mousewheel event bubble prevention
+ */
+export enum PageScrollPreventionMode {
+   /**
+    * Do not prevent any page scrolling
+    */
+   None,
+   /**
+    * Prevent page scroll if mouse is over the game canvas
+    */
+   Canvas,
+   /**
+    * Prevent all page scrolling via mouse wheel
+    */
+   Page
+}
+
 /*
 * Interface describing the absolute CSS position of the game window. For use when DisplayMode.Position
 * is specified and when the user wants to define exact pixel spacing of the window.
@@ -112,6 +130,11 @@ export interface IEngineOptions {
    */
    
    position?: string | IAbsolutePosition;
+
+   /**
+    * Scroll prevention method.
+    */
+   pageScrollPreventionMode?: PageScrollPreventionMode;
 }
 
 /**
@@ -234,6 +257,11 @@ export class Engine extends Class {
     */
    public onFatalException = (e: any) => { Logger.getInstance().fatal(e); };
 
+   /**
+    * The mouse wheel scroll prevention mode
+    */
+   public pageScrollPreventionMode: PageScrollPreventionMode;
+
    private _logger: Logger;
    private _isSmoothingEnabled: boolean = true;
 
@@ -273,7 +301,8 @@ export class Engine extends Class {
       canvasElementId: '',
       pointerScope: Input.PointerScope.Document,
       suppressConsoleBootMessage: null,
-      suppressMinimumBrowserFeatureDetection: null
+      suppressMinimumBrowserFeatureDetection: null,
+      pageScrollPreventionMode: PageScrollPreventionMode.Canvas
    };
 
    /**
@@ -874,6 +903,8 @@ O|===|* >________________>\n\
       this.input.keyboard.init();
       this.input.pointers.init(options ? options.pointerScope : Input.PointerScope.Document);
       this.input.gamepads.init();
+
+      this.pageScrollPreventionMode = options.pageScrollPreventionMode;
 
       // Issue #385 make use of the visibility api
       // https://developer.mozilla.org/en-US/docs/Web/Guide/User_experience/Using_the_Page_Visibility_API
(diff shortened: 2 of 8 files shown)


mrkmg:
Not ready for merging yet. Look for feedback.

src/engine/Docs/Pointers.md
```diff
@@ -26,6 +26,17 @@ engine.input.pointers.primary.on("move", function (evt) { });
 engine.input.pointers.primary.on("cancel", function (evt) { });
 ```
 
+### Wheel Event
+
+You can also subscribe to the mouse wheel event through `engine.input.points.on`. A [[WheelEvent]]
```
kamranayub: Typo I think, should be `engine.input.pointers.on`

src/engine/Input/Pointer.ts
```diff
@@ -80,6 +86,44 @@ export class PointerEvent extends GameEvent<any> {
 };
 
 /**
+ * Wheel Events
+ *
+ * Represents a mouse wheel event. See [[Pointers]] for more information on
+ * handling point input.
+ */
+export class WheelEvent extends GameEvent<any> {
```
kamranayub: Do you think it'd be helpful to provide a `direction` property for developers to quickly determine direction? Or is comparing positive to negative enough?

kamranayub:
Do we need to support multiple ways of preventing scrolling on the page? I was thinking it'd probably be an `IEngineOptions` parameter. I think it would pass through to a property on the Pointers class.

The modes I can think of:

- **None** - Excalibur won't prevent default
- **Canvas** - Prevents scroll if mouse is over game canvas, so page doesn't scroll (**default?**)
- **Document** - Prevents scroll at all times, users have to use scrollbars

```
new ex.Engine({ preventPageScrollMode: ex.Input.PageScrollPreventMode.Canvas })
game.input.pointers.preventPageScrollMode= ex.Input.PageScrollPreventMode.None;
```

I'm up for a better name of the property.

It should definitely keep inspecting that flag every event, in case you need to dynamically manipulate it.

src/engine/Docs/Pointers.md
mrkmg: Agreed. Will update shortly.

src/engine/Input/Pointer.ts
mrkmg: Maybe? Would we use an Enum for that? WheelDirection { Up, Down, Left, Right, In, Out }?

Theoretically, you could have *+ or -* in *x, y, or z*. Although in most circumstances, only *y* will be used (as thats the typical scroll wheel).

mrkmg:
@kamranayub the scroll prevent modes is a great idea. I think the name's you have chosen are clear and should be very apparent to the user what they mean. I will implement this shortly.

src/engine/Input/Pointer.ts
mrkmg: Thinking about it, maybe the Enum should be:

```ts
enum WheelDirection {
 Forward,
 Backward,
 Left,
 Right,
 Inward,
 Outward
}
```

How does that sound @kamranayub ?

mrkmg: We will have to take into account devices which can scroll in multiple directions at once, do we examine the magnitude and choose the highest for the "direction" enum?

Then, what do we do if deltaX and deltaY are equal, do we always favor Y movement over X movement?

mrkmg: Upon further thought, I think this will not work in an way which is obvious.

According to spec, the result is essentially a 3D Vector. It describes a physical user action to a direction and magnitude in a 3D space.  

Since Excaliburjs does not have an internal 3D Vector model I don't think there is a reason to implement this. Giving the delta of X, Y, and Z is easy enough for any developer. Doing all the checks to figure out the direction seems a little overkill when most people will just a simple check of deltaY.

kamranayub: Yeah, sounds good!

src/engine/Input/Pointer.ts
```diff
@@ -309,6 +376,59 @@ export class Pointers extends Class {
...
+
+         var deltaX = 0;
+         var deltaY = 0;
+         var deltaZ = 0;
+         var deltaMode = WheelDeltaMode.Pixel;
+
+         if (e.deltaX) {
+            deltaX = e.deltaX;
+         } else if (e.wheelDeltaX) {
+            deltaX = e.wheelDeltaX * (-1 / 40);
```
eonarheim: @mrkmg What does the (-1 / 40) stand for? Can we pull this into a constant?

src/engine/Input/Pointer.ts
```diff
@@ -309,6 +376,59 @@ export class Pointers extends Class {
...
+         }
+
+         if (e.deltaY) {
+            deltaY = e.deltaY;
+         } else if (e.wheelDeltaY) {
+            deltaY = e.wheelDeltaY * (-1 / 40);
+         } else if (e.wheelDelta) {
```
eonarheim: Are the different properties for different browser types? Can we add some comments for which browsers?

src/engine/Engine.ts
```diff
@@ -47,6 +47,24 @@ export enum DisplayMode {
    Position
 }
 
+/**
+ * Enum representing the different mousewheel event bubble prevention
+ */
+export enum PageScrollPreventionMode {
```
eonarheim: Let's omit the word`Page` from `PageScrollPreventionMode` to make it simpler. `ScrollPreventionMode` is pretty clear

src/engine/Engine.ts
```diff
@@ -47,6 +47,24 @@ export enum DisplayMode {
...
+    */
+   None,
+   /**
+    * Prevent page scroll if mouse is over the game canvas
+    */
+   Canvas,
+   /**
+    * Prevent all page scrolling via mouse wheel
+    */
+   Page
```
eonarheim: Let's change this option to `All` matches the comment and behavior better

eonarheim:
I noticed on the browser test, the value for scrolling is either 100 or -100 but nothing in between. Is this expected?

image

image

src/engine/Input/Pointer.ts
```diff
@@ -309,6 +376,59 @@ export class Pointers extends Class {
...
+         var deltaZ = 0;
+         var deltaMode = WheelDeltaMode.Pixel;
+
+         if (e.deltaX) {
+            deltaX = e.deltaX;
+         } else if (e.wheelDeltaX) {
+            deltaX = e.wheelDeltaX * (-1 / 40);
+         }
+
+         if (e.deltaY) {
```
eonarheim: For everything being set to `deltaY` can we use a null coalesce trick to save some lines?

```typescript
deltaY = e.deltaY || (e.wheelDeltaY * CONSTANT) || (e.wheelDelta * CONSTANT) || e.detail
```

mrkmg:
I just pushed updates based on your reviews

src/engine/Input/Pointer.ts
```diff
@@ -50,6 +50,12 @@ export enum PointerScope {
 }
 
 /**
+ * A constant used to normal wheel events across different browsers
+ * @type {number}
```
kamranayub: Don't actually need `@type` annotation, TypeScript knows what type it is 👍

src/engine/Input/Pointer.ts
```diff
@@ -379,40 +385,22 @@ export class Pointers extends Class {
...
          }
 
          var x: number = e.pageX - Util.getPosition(this._engine.canvas).x;
          var y: number = e.pageY - Util.getPosition(this._engine.canvas).y;
          var transformedPoint = this._engine.screenToWorldCoordinates(new Vector(x, y));
 
-         var deltaX = 0;
-         var deltaY = 0;
-         var deltaZ = 0;
+         // Support
```
kamranayub: Can we add some comments that explain what the normalization factor is for? For posterity.

mrkmg:
Just pushed another update.

@kamranayub and @eonarheim should we name the "wheel" event for actors to "poitnerwheel"? I feel like that coincides with the rest of the pointer events.

kamranayub:
@mrkmg Yeah, I like that! Good idea.

mrkmg:
@kamranayub and @eonarheim I have renamed the wheel event.

I went through all the rest of the changes requested. I think they are all covered.

kamranayub:
@mrkmg As far as I know this looks complete. Once @eonarheim approves, can we merge?

Issue #804, opened 2017-05-09
Sound profile feature for dynamic sounds


eonarheim:
### Context
It would be super cool to change the volume and other properties of a sound over time while it is playing. Perhaps other properties could also be controlled, like pitch, or anything else web audio supports.

### Proposal

Potentially add a `ex.Sound.playWith(profile: ISoundProfile)` that takes a ISoundProfile to describe how the sound should change over time.

```typescript
// all of these properties are optional
var profile: ISoundProfile = {
   delay: 20, // delay in milliseconds to wait before playing any sound
   volume: [ {time: .2, volume: .8}, { time: 1, volume: .5 }], // time is a percentage of the whole track 1 being the end, 0 being teh beginning
   interpolator: ex.EasingFunctions =  EasingFunctions.EaseInOutCubic
}
````

andrewmbyrd:
I believe I've got a solid understanding of how to implement a bezier curve in general, but where would it be best to put the code? Since `play` is a `Promise`, it seems hard to have any aspect of the sound change _while_ the sound is playing. 

`HTMLAudioElement` seems to at least have an `ontimeupdate` event listener that could be useful, but I couldn't find any such function for `AudioBufferSourceNode`

eonarheim:
SoundManager solves this

PR #810, opened 2017-05-12
[#804] Audio profile first draft

andrewmbyrd:
Begins to deal with #804 

##NOT A COMPLETE REQUEST
- I've made a shot at implementing the sound profiles. The build so far was just to test that I could control volume while the song was playing - no actual user input is configured yet, and I've only got it working for the loop audio file in the sandbox.

I've used the [cubic bezier curve equation](https://en.wikipedia.org/wiki/B%C3%A9zier_curve#Cubic_B.C3.A9zier_curves) in lines 610-640 to change volume as the song is playing, based on 4 possible "points". (points usually require two components, but since there is no x and y with respect to volume - only time and volume - I've omitted the independent variable of time) 

Doing it this way seems really computation hungry - maybe it's best to only update the volume 4 or 5 times throughout the course of the sound? Of course this will kill a lot of the smoothness of volume transition.

Does the `playWithProfile` function idea work well in this context? It's a bit of a repeat of the `play` function, but it is doing a very different thing.

Please let me know your thoughts

### DIFF
--- a/src/engine/Interfaces/IAudio.ts
+++ b/src/engine/Interfaces/IAudio.ts
@@ -49,4 +49,29 @@ export interface IAudio {
     * Stop playing the sound and reset
     */
    stop(): void;
+   
+   
+   /**
+    * File of the HTMLAudioElement
+    */
+    audioElement?: HTMLAudioElement;
+    
+   /**
+    * File of the WebAudio Element
+    */
+   bufferSource?: AudioBufferSourceNode;
+   
+   buffer?: AudioBuffer;
+   /**
+    * System time when the audio file begins playing
+    */
+   absoluteStartTime?: number;
+   
+   /**
+    * Bezier points for adjusting volume through the engine
+    */
+   point1?: number;
+   point2?: number;
+   point3?: number;
+   point4?: number;
 }
--- a/src/engine/Interfaces/IAudio.ts
+++ b/src/engine/Interfaces/IAudio.ts
@@ -24,6 +24,11 @@ export interface IAudio {
     * Will play the sound or resume if paused
     */
    play(): Promise<any>;
+   
+   /**
+    * Will play the sound with a Bezier-type volume profile
+    */
+   playWithProfile(): Promise<any>;
 
    /**
     * Pause the sound
--- a/src/engine/Interfaces/IAudio.ts
+++ b/src/engine/Interfaces/IAudio.ts
@@ -36,7 +36,7 @@ export interface IAudio {
    play(): Promise<any>;
    
    /**
-    * Will play the sound with a Bezier-type volume profile
+    * Will play the sound with a cubic Bezier-type volume profile
     */
    playWithProfile(profile: string | IBezierPoints): Promise<any>;
 
--- a/src/engine/Interfaces/IAudio.ts
+++ b/src/engine/Interfaces/IAudio.ts
@@ -1,6 +1,16 @@
 import { Promise } from './../Promises';
 
 /**
+ * Defines the points which describe a cubic bezier curve for a sound profile
+ */
+export interface IBezierPoints {
+  point1: number;
+  point2: number;
+  point3: number;
+  point4: number;
+}
+
+/**
  * Represents an audio control implementation
  */
 export interface IAudio {
@@ -28,7 +38,7 @@ export interface IAudio {
    /**
     * Will play the sound with a Bezier-type volume profile
     */
-   playWithProfile(): Promise<any>;
+   playWithProfile(profile: string | IBezierPoints): Promise<any>;
 
    /**
     * Pause the sound
--- a/src/engine/Resources/Sound.ts
+++ b/src/engine/Resources/Sound.ts
@@ -276,6 +276,13 @@ export class Sound implements ILoadable, IAudio {
 
    /**
     * Play the sound, returns a promise that resolves when the sound is done playing
+    * Valid string inputs include:
+    * -'linearUp'
+    * -'linearDown'
+    * -'ease'
+    * -'ease-in'
+    * -'ease-out'
+    * -'ease-in-out'
     */
    public play(data?: string | number | IBezierPoints): Promise<boolean> {
       if (this._isLoaded) {
--- a/src/engine/Resources/Sound.ts
+++ b/src/engine/Resources/Sound.ts
@@ -1,6 +1,8 @@
 import { ILoadable } from '../Interfaces/ILoadable';
 import { IAudioImplementation } from '../Interfaces/IAudioImplementation';
 import { IAudio } from '../Interfaces/IAudio';
+import { IBezierPoints} from '../Interfaces/IAudio';
+import { EasingFunctions } from '../Util/EasingFunctions';
 import { Logger } from '../Util/Log';
 import * as Util from '../Util/Util';
 import { Engine } from '../Engine';
@@ -275,7 +277,7 @@ export class Sound implements ILoadable, IAudio {
    /**
     * Play the sound, returns a promise that resolves when the sound is done playing
     */
-   public play(volume?: string): Promise<boolean> {
+   public play(data?: string | number | IBezierPoints): Promise<boolean> {
       if (this._isLoaded) {
          var resumed = [];
 
@@ -302,8 +304,8 @@ export class Sound implements ILoadable, IAudio {
          this._tracks.push(newTrack);
 
          this._logger.debug('Playing new instance for sound', this.path);
-         if (volume) {
-           return newTrack.playWithProfile().then(() => {
+         if (data && typeof data !== 'number') {
+           return newTrack.playWithProfile(data).then(() => {
   
               // when done, remove track
               this._tracks.splice(this._tracks.indexOf(newTrack), 1);
@@ -488,14 +490,73 @@ class AudioTagInstance implements IAudio {
       return this._playingPromise;
    }
    
-   public playWithProfile(){
+   public playWithProfile(profile: string | IBezierPoints) {
       if (this._isPaused) {
         this._resume();
       } else if (!this._isPlaying) {
         this._start();
       }
-      this._audioElement.ontimeupdate = function() {
-        console.log("Playing")
+      
+      var point1;
+      var point2;
+      var point3;
+      var point4;
+      
+      if (typeof profile === 'string') {
+        switch (profile) {
+          case 'linearUp':
+            point1 = 0.0;
+            point2 = 0.33;
+            point3 = 0.67;
+            point4 = 1.0;
+            break;
+          case 'linearDown':
+            point1 = 1.0;
+            point2 = 0.67;
+            point3 = 0.33;
+            point4 = 0.0;
+            break;
+          case 'ease':
+            point1 = 0.0;
+            point2 = 0.1;
+            point3 = 0.9;
+            point4 = 1.0;
+            break;
+          case 'ease-in':
+            point1 = 0.0;
+            point2 = 0.0;
+            point3 = 0.75;
+            point4 = 1.0;
+            break;
+          case 'ease-out':
+            point1 = 0.0;
+            point2 = 0.33;
+            point3 = 0.9;
+            point4 = 1.0;
+            break;
+          case 'ease-in-out':
+            point1 = 0.0;
+            point2 = 0.0;
+            point3 = 1.0;
+            point4 = 1.0;
+            break;
+          default:
+            throw new Error('Invalid profile given');
+        }
+      } else {
+        point1 = profile.point1;
+        point2 = profile.point2;
+        point3 = profile.point3;
+        point4 = profile.point4;
+      }
+      
+      if (point1 > 1 || point1 < 0 || point2 > 1 || point2 < 0 || point3 > 1 || point3 < 0 || point4 < 0 || point4 > 1) {
+        throw new Error('All points for the play profile must be between 0 and 1');
+      }
+      
+      while (this._audioElement.currentTime < this._audioElement.duration) {
+        var currentPlaythroughTimePercentage = this._audioElement.currentTime / this._audioElement.duration;
+        this.setVolume(EasingFunctions.CubicBezier(currentPlaythroughTimePercentage, point1, point2, point3, point4));
       }
       return this._playingPromise;
    }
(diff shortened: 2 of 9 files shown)


src/engine/Resources/Sound.ts
```diff
@@ -582,6 +607,25 @@ class WebAudioInstance implements IAudio {
...
+     } else if (!this._isPlaying) {
+       this._start();
+     }
+     var point1 = 0.8;
+     var point2 = 0.1;
+     var point3 = 0.1;
+     var point4 = 0.4;
+     
+     while (new Date().getTime() < this._startTime + this._buffer.duration*1000) {
+       var cpt = (new Date().getTime() - this._startTime) / (this._buffer.duration*1000) ;
```
eonarheim: `new Date()` is really hard to test because it can't be mocked easily. Could we use a different method of counting time perhaps similar to how `ex.Timer` works? It would mean that we'd have to add a bit more plumbing into sound so that it can be "updated" over time maybe by adding a `update(delta: number)` method to sound, this does pose a challenge of wiring it into the `ex.Engine` update cycle without making sounds too painful to use. 

This API also looks promising and is designed for working with Audio time https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/currentTime

src/engine/Resources/Sound.ts
```diff
@@ -582,6 +607,25 @@ class WebAudioInstance implements IAudio {
...
+       this.setVolume(((1-cpt)**3)*point1 + 3*((1-cpt)**2)*cpt*point2 + 3*(1-cpt)*(cpt**2)*point3 + (cpt**3)*point4);
```
eonarheim: We also have some pre-defined easing functions in `Util/EasingFunctions.ts` that are defined on the domain `[0, duration]` which might be useful here.

eonarheim: Also the TypeScript linter will fail on the math operators not being separated by a space.

src/engine/Resources/Sound.ts
andrewmbyrd: I'd taken a look at the `currentTime` property on `AudioContext`, but the problem seems to be that it is ever-increasing. So the moment the page is loaded, a counter starts going up, as far as I could tell. So I'm not sure if that would support a button press, that could occur at anytime, and then adjust the sound profile based off of the `currentTime`.

I'll take a look at `ex.Timer`

andrewmbyrd: The more I think about it, it seems `currentTime` may be useful to track current playthrough time in a song, although `this._startTime` may not be used then, which I suppose is no problem

andrewmbyrd:
So I've taken a look at the easing functions in `Util/EasingFunctions`, and it looks like those produce a graph that is a direct function of time. So like on `EaseInCubic` the return value is `endValue * currentTime * currentTime * currentTime + startValue`, which, for a series of calls with different `currentTime`, should produce a graph like the ones seen [here](http://jwilson.coe.uga.edu/EMT668/EMT668.Folders.F97/Wynne/Cubic/Cubic%20Functions.html) correct?

I was under the impression that we'd want the graph of volume to follow a pattern more like the one shown [here](http://cubic-bezier.com/), where the points provided by the user are (usually) never actually touched, but serve as guidance to where the bezier curve should be drawn.

Should the function implement the graph shown in the first link? The second link? Am I missing the intent altogether?

andrewmbyrd:
In this most recent commit, I've got it set up so that we're using the currentTime API. I've also moved the cubic bezier calculation into `Util/EasingFunctions`. My only thought is that by doing it this way, volume is being updated several hundred times per second; if you could provide some guidance on how to tie this into the 60 fps engine update, it would be appreciated!

Also I've yet to add spec tests... I'm not sure how the mock audio will work with no real file to play.

eonarheim:
@andrewmbyrd Actually I think you have a point about the volume over time problem, the easing functions, as is, are not designed for this type of continuous curve given a bunch of points. The cubic bezier calculation sounds like the right thing the more I think about it. Also `Util/EasingFunctions` sounds like the best place for the implementation. The intent originally was to specify a list of points (points of volume) and do dynamic volume interpolation, which a bezier curve would work perfectly.

As for recalculating the volume, I agree this would be computationally prohibitive. We could hook into the `ex.Sound.load` routines and front load the calcuation and cache it at loader time inside the sound object. That way you don't pay the penalty when you play the sound, the users expect to wait when they see the loading bar.

This way we would only need to ever recalculate the points of the curve if a new profile was provided to the `ex.Sound`

We could assume cached points at 16ms timestep when being precalculated (which would be the fastest it could change in a 60fps based update anyway). If that turns out to be to much memory we could just sample the bezier curve at larger time steps and save those points and during play time linear interpolate between sampled points where appropriate. I bet the human ear wont be able to tell the difference... maybe...

Something like this? Where the red lines are interpolated between sample points of the "true" bezier
image

For this spec, I think just confirming that the logical volume value changes over time as we would expect. If we want to do a visual test to confirm with our ears like we did with the `wasTouching`

src/engine/Resources/Sound.ts
```diff
@@ -274,8 +276,15 @@ export class Sound implements ILoadable, IAudio {
...
+    * Valid string inputs include:
+    * -'linearUp'
+    * -'linearDown'
+    * -'ease'
+    * -'ease-in'
+    * -'ease-out'
+    * -'ease-in-out'
     */
-   public play(volume?: string): Promise<boolean> {
+   public play(data?: string | number | IBezierPoints): Promise<boolean> {
```
eonarheim: I like idea of simply saying `play(linearUp)` in sound. I probably would want to control the maximum volume somehow, I could imagine this getting loud accidentally.

Using a type union we can strongly type these options, take a look at events
```typescript
// Events.ts
export type kill = 'kill';

export type predraw = 'predraw';
export type postdraw = 'postdraw';

export type predebugdraw = 'predebugdraw';
export type postdebugdraw = 'postdebugdraw';

export type preupdate = 'preupdate';
export type postupdate = 'postupdate';

export type preframe = 'preframe';
export type postframe = 'postframe';

export type collision = 'collision';

export type initialize = 'initialize';
export type activate = 'activate';
...
// Actor.ts
public on(eventName: Events.kill, handler: (event?: KillEvent) => void): void;
public on(eventName: Events.initialize, handler: (event?: InitializeEvent) => void): void;
public on(eventName: Events.preupdate, handler: (event?: PreUpdateEvent) => void): void;
public on(eventName: Events.postupdate, handler: (event?: PostUpdateEvent) => void): void;
public on(eventName: Events.predraw, handler: (event?: PreDrawEvent) => void): void;
public on(eventName: Events.postdraw, handler: (event?: PostDrawEvent) => void): void;
public on(eventName: Events.predebugdraw, handler: (event?: PreDebugDrawEvent) => void): void;
public on(eventName: Events.postdebugdraw, handler: (event?: PostDebugDrawEvent) => void): void;
public on(eventName: Events.pointerup, handler: (event?: PointerEvent) => void): void;
public on(eventName: Events.pointerdown, handler: (event?: PointerEvent) => void): void;
public on(eventName: Events.pointermove, handler: (event?: PointerEvent) => void): void;
public on(eventName: Events.pointercancel, handler: (event?: PointerEvent) => void): void;
public on(eventName: string, handler: (event?: GameEvent<any>) => void): void;
public on(eventName: string, handler: (event?: GameEvent<any>) => void): void {
    this._checkForPointerOptIn(eventName);
    this.eventDispatcher.on(eventName, handler);
}
```

src/engine/Interfaces/IAudio.ts
```diff
@@ -1,6 +1,16 @@
...
 
 /**
+ * Defines the points which describe a cubic bezier curve for a sound profile
+ */
+export interface IBezierPoints {
+  point1: number;
+  point2: number;
+  point3: number;
+  point4: number;
+}
```
andrewmbyrd: It can. The general definition of a bezier curve is given [here](https://en.wikipedia.org/wiki/B%C3%A9zier_curve#Cubic_B.C3.A9zier_curves). I like the cubic because it allows for up to two changes in direction, and can do everything the quadratic and linear equations can do. Getting much more than 4 bezier points starts to become maybe more confusing than necessary in my opinion, but it can be done

src/engine/Resources/Sound.ts
andrewmbyrd: Right now, I've got it so that the bezier points will be set automatically if the type of the argument passed into `play` is `string`. If we make it so that 'linearUp', 'ease', etc. are all their own types, is there a way to group those types together at runtime, so that we know the input wasn't an `IBezierPoints`?

Essentially I'm saying the syntax `typeof variable == 'string'` works, but `typeof variable == "IBezierPoints"` and `variable instanceof IBezierPoints` throw errors, so I'm not sure how to identify custom types/interfaces at runtime.

kamranayub: In this case, interface will just be an object, so you can determine if it's a string it's an easing constant or if it's an object, it's (probably) the interface.

```ts
if (typeof arg === 'string') {
// easing string
} else if (typeof arg === 'object') {
// IBezierPoints implementation
}
```

If you wanted to strongly type the IBezierPoints check, you can use [user-defined type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards).

```ts
function isBezierPoints(arg: object): arg is IBezierPoints {
   return (<IBezierPoints>arg).point1;
}

Does that help?

Comment (andrewmbyrd) [inline src/engine/Resources/Sound.ts]: I guess I'm confused on whether it's better to just pass in a string:

```
sound.playWithProfile('linearUp')
```

or to create Types and then assign those to Strings. I'm picturing that the method call would look identical except for the quotes:
```
sound.playWithProfile(linearUp)
```

Comment (kamranayub) [inline src/engine/Resources/Sound.ts]: Well, they can't be "types" per se, they'd have to be constants. And then you'd have to do something like `ex.Audio.EASE_LINEAR_UP` or something. I feel like regular strings is probably OK, folks with TS will get compile errors and JS users will reference the docs to know what are valid strings to pass.

**Comment by andrewmbyrd** [comment]
I've gotten it so that the audio will now change volume in line with when the engine updates! (Right now it does this every 10th frame but that is easily changed to every frame or every 5th frame or whatever
`if (this.sounds.length > 0 && this.stats.currFrame.id % 10 === 0 ) `)

This way doesn't seem too expensive because it's just one arithmetic calculation being done when the update is called.

This is creating an error in the Frame Spec tests, but it seems this is only affecting tests; "visual" tests with the new feature are working just fine.

The problem for the specs comes from this bit of code;
```
if (this.sounds.length > 0 && this.stats.currFrame.id % 10 === 0 ) {
        for (var sound of this.sounds){
           sound.update();
        };
      }
```

but the problem persists even if I change the code to this
```
if (this.sounds.length > 0 ) {
     console.log(this.sounds);
  }
```
So it doesn't appear to be an issue with expensive computation or anything...
Maybe I'm missing something with how the FrameStats spec is performed?

**Comment by kamranayub** [comment]
@andrewmbyrd when you modified those lines to the last example, did you run both the compile and test steps? It's easy to forget--I can't see how that change would affect anything. Are you saying if you remove that block, everything works OK? I usually run `grunt compile tests` to run both steps, but `npm run all` makes sure to run everything.

**Comment by andrewmbyrd** [comment]
@kamranayub  yeah I'm confused by it as well! But I can confirm that commenting out lines 880 - 885 in `Engine.ts`: 
```
  //update volume on profiled sounds
      /*if (this.sounds.length > 0 && this.stats.currFrame.id % 10 === 0 ) {
        for (var sound of this.sounds){
           sound.update();
        };
      }*/
```
and then `npm run all` makes the normal tests pass (there's been 8 tests that have failed since I've been working on this project, but those don't fail the Travis CI build when I push up to github).

When I un-comment those lines, not changing anything else, and do `npm run all` , I get 6 new errors, 4 here:
```
 The engine
   √ should have current and previous frame stats defined
   after frame is ended
     × should collect frame delta
       TypeError: undefined is not a constructor (evaluating 'game.stop()') in file:///C:/Users/Byrd/Documents/Bloc/Excalibur/.grunt/grunt-contrib-jasmine/build/dist/excalibur.js (line 9) (1)
       TypeError: undefined is not an object (evaluating 'stats.delta') in file:///C:/Users/Byrd/Documents/Bloc/Excalibur/src/spec/FrameStatsSpec.js (line 26) (2)
     × should collect frame fps
       TypeError: undefined is not a constructor (evaluating 'game.stop()') in file:///C:/Users/Byrd/Documents/Bloc/Excalibur/.grunt/grunt-contrib-jasmine/build/dist/excalibur.js (line 9) (1)
       TypeError: undefined is not an object (evaluating 'stats.fps') in file:///C:/Users/Byrd/Documents/Bloc/Excalibur/src/spec/FrameStatsSpec.js (line 29) (2)
     × should collect frame actor stats
       TypeError: undefined is not a constructor (evaluating 'game.stop()') in file:///C:/Users/Byrd/Documents/Bloc/Excalibur/.grunt/grunt-contrib-jasmine/build/dist/excalibur.js (line 9) (1)
       TypeError: undefined is not an object (evaluating 'stats.actors') in file:///C:/Users/Byrd/Documents/Bloc/Excalibur/src/spec/FrameStatsSpec.js (line 32) (2)
     × should collect frame duration stats
       TypeError: undefined is not a constructor (evaluating 'game.stop()') in file:///C:/Users/Byrd/Documents/Bloc/Excalibur/.grunt/grunt-contrib-jasmine/build/dist/excalibur.js (line 9) (1)
       TypeError: undefined is not an object (evaluating 'stats.duration') in file:///C:/Users/Byrd/Documents/Bloc/Excalibur/src/spec/FrameStatsSpec.js (line 39) (2)
```
and 2 here
```
 The engine
   √ should default to a timescale of 1.0
   × should run at 2x speed when timescale is 2.0
     TypeError: undefined is not a constructor (evaluating 'game.stop()') in file:///C:/Users/Byrd/Documents/Bloc/Excalibur/.grunt/grunt-contrib-jasmine/build/dist/excalibur.js (line 9) (1)
   × should run at 1/2 speed when timescale is 0.5
     TypeError: undefined is not a constructor (evaluating 'game.stop()') in file:///C:/Users/Byrd/Documents/Bloc/Excalibur/.grunt/grunt-contrib-jasmine/build/dist/excalibur.js (line 9) (1)
```

kamranayub:
@andrewmbyrd we can help debug this issue, but do you mind update this branch with the latest master code? It's a bit out-of-date and updating it will affect the code you wrote. Once that's ready, we can bring it down and try to step through in the debugger to figure this out.

jedeen:
Thanks for working on this! We're going to close this PR for now, but feel free to revisit it in the future!

PR #812, opened 2017-05-22
[#798] Rename Utils.removeItemToArray

Crizzooo:
Closes #798

## Changes:

- Utils.removeItemToArray -> Utils.removeItemFromArray
- Created test specs to test this function in UtilSpec.ts

PR #813, opened 2017-05-22
Pointer event vector property

### LINKED ISSUE #612: Add position vector to ex.Input.PointerEvent
### Context

`ex.Input.PointerEvent` contains `x` and `y` properties because that is what the W3C spec says MouseEvent should have but to be consistent within Excalibur's API, it should just be a `pos` Vector with the same values.

Look at other events to ensure it's consistent.
### Proposal
1. `@deprecate` `x` and `y`
2. Add `pos` vector instead

This will be a breaking change when `x` and `y` are removed.

Crizzooo:
Closes #612 

I see that a large majority of the diff are due to whitespacing issues. Do I just need to download TSLint and run with your settings or something to fix these? My apologies...

## Changes:

- PointerEvent now initializes property `pos` with a Vector using X and Y arguments from the constructor. X and Y are still public properties


kamranayub:
It looks like you may have accidentally brought along the changes from #812, do you mind removing that commit from this PR so we can keep them separated?

Also, for TSLint, you can just run the full suite `npm run all` to catch any linting errors, but we'd prefer to keep whitespace modifications separated from code change commits.

kamranayub:
Closing since you opened #814

PR #815, opened 2017-05-24
Changed pointer initialization to match same initialization as keyboard in order to make mocking for tests possible.

mrkmg:
I started to write tests for Pointer, but quickly realized that the init function for the Pointers class was not able to easily be mocked.

I changed it to match the same method that Keyboard uses, which is to pass in the "target" instead of an option to define what it is binding to.

### DIFF
--- a/src/engine/Input/Pointer.ts
+++ b/src/engine/Input/Pointer.ts
@@ -121,8 +121,8 @@ export class Pointers extends Class {
    /**
     * Initializes pointer event listeners
     */
-   public init(target?: any): void {
-      target = target || <any>this._engine.canvas;
+   public init(target?: GlobalEventHandlers): void {
+      target = target || this._engine.canvas;
 
       // Touch Events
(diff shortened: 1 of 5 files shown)


src/engine/Input/Pointer.ts
```diff
@@ -121,13 +121,8 @@ export class Pointers extends Class {
...
     * Initializes pointer event listeners
     */
-   public init(scope: PointerScope = PointerScope.Document): void {
-      var target = <any>document;
-      if (scope === PointerScope.Document) {
-         target = document;
-      } else {
-         target = <any>this._engine.canvas;
-      }
+   public init(target?: any): void {
```
kamranayub: It may make sense to type target as `GlobalEventHandlers` which Document and Element both implement.

mrkmg: Would it then also make sense to make Keyboard have the same type?

mrkmg: I have implemented GlobalEventHandlers on both Pointers and Keyboard

PR #819, opened 2017-05-24
Add Pointer tests

mrkmg:
## Changes

- Add initial tests for pointers

### DIFF
--- a/src/spec/PointerInputSpec.ts
+++ b/src/spec/PointerInputSpec.ts
@@ -1,0 +1,101 @@
+/// <reference path="jasmine.d.ts" />
+
+/// <reference path="Mocks.ts" />
+
+describe('A pointer', () => {
+   
+   var mockWindow = null;
+   var engine: ex.Engine = null;
+   var pointers: ex.Input.Pointers = null;
+   var mocker = new Mocks.Mocker();
+   var noop = () => { return null; };
+
+   beforeEach(() => {
+      mockWindow = <any>mocker.window();
+      engine = mocker.engine(600, 400);
+      pointers = new ex.Input.Pointers(engine);
+      pointers.init(mockWindow);
+   });
+   
+   it('should exist', () => {
+      expect(ex.Input.Pointers).toBeDefined();
+      expect(pointers).toBeTruthy();
+   });
+   
+   it('should fire mousedown events', () => {
+      var eventLeftFired = false;
+      var eventRightFired = false;
+      var eventMiddleFired = false;
+      
+      pointers.primary.on('down', function(ev: ex.Input.PointerEvent) {
+         if (ev.button === ex.Input.PointerButton.Left) {
+            eventLeftFired = true;
+         }
+         if (ev.button === ex.Input.PointerButton.Right) {
+            eventRightFired = true;
+         }
+         if (ev.button === ex.Input.PointerButton.Middle) {
+            eventMiddleFired = true;
+         }
+      });
+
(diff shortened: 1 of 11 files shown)


src/spec/PointerInputSpec.ts
```diff
@@ -0,0 +1,101 @@
...
+   
+   var mockWindow = null;
+   var engine: ex.Engine = null;
+   var pointers: ex.Input.Pointers = null;
+   var mocker = new Mocks.Mocker();
+   var noop = () => { return null; };
+
+   beforeEach(() => {
+      mockWindow = <any>mocker.window();
+      engine = mocker.engine(600, 400);
```
kamranayub: I think we want to try switching over to `TestUtils.engine` instead for new tests, which is a real engine instance that can be tested. You will need:

```
afterEach(() => {
  engine.stop();
});
```

As well, to make sure it gets stopped after each test.

src/spec/PointerInputSpec.ts
```diff
@@ -0,0 +1,101 @@
+/// <reference path="jasmine.d.ts" />
+
+/// <reference path="Mocks.ts" />
+
+describe('A pointer', () => {
+   
+   var mockWindow = null;
```
kamranayub: Since we're running in Phantom, what happens if you simply remove these and reference `window`?

mrkmg: Changed

src/spec/PointerInputSpec.ts
mrkmg: Changed

src/spec/PointerInputSpec.ts
```diff
@@ -0,0 +1,135 @@
...
+         }
+         if (ev.button === ex.Input.PointerButton.Right) {
+            eventRightFired = true;
+         }
+         if (ev.button === ex.Input.PointerButton.Middle) {
+            eventMiddleFired = true;
+         }
+      });
+
+      var mousedownEventLeft = document.createEvent('MouseEvent');
```
kamranayub: Can we create a little helper function at the top of the file to reduce the duplication here?

mrkmg: I was thinking of doing that, but I was not sure where the best to put that helper function? Should I put it in the PointerInputSpec file, or the TestUtils file?

kamranayub: Just at the top of the Input spec file I think, since it's a local helper function

mrkmg: Moved to a helper function. Much cleaner!

src/spec/PointerInputSpec.ts
```diff
@@ -0,0 +1,111 @@
+/// <reference path="jasmine.d.ts" />
+
+/// <reference path="TestUtils.ts" />
+
+function executeMouseEvent(type: string, button: ex.Input.PointerButton, x: number, y: number, target: HTMLElement) {
```
kamranayub: Sorry, one last change. If you move this function inside the `describe` function, it can access `engine.canvas`, eliminating the extra argument. Additionally, TS supports default params so you can set `x = 0` and `y = 0` to eliminate other duplication below too.

mrkmg: I am leaving the "target" as an optional property so I can add the Wheel Event tests once PR #809 is merged. In order to test I will need to trigger wheel events on both the canvas and document.

mrkmg:
@kamranayub I moved the helper function and created the default parameter values. Should be all set. Once #809 is merged, I will create the tests for the Wheel Events

PR #823, opened 2017-05-26
[#517] Add static method Vector.distance

### LINKED ISSUE #517: utility method to calculate the distance between two points
`ex.Util.distance(point1: Point, point2: Point)`

Crizzooo:
Closes #517

## Changes:

- Added a static method `distance` to the Vector class
- Added a test spec in AlgebraSpec.ts to unit test this method

### DIFF
--- a/src/engine/Algebra.ts
+++ b/src/engine/Algebra.ts
@@ -68,14 +68,14 @@ export class Vector {
       return true;
    }
 
-  /**
-  * Calculates distance between two Vectors
-  * @param vec1
-  * @param vec2
-  */
-  public static distanceBetweenVectors(vec1: Vector, vec2: Vector) {
-     return Math.sqrt(Math.pow(vec1.x - vec2.x, 2) + Math.pow(vec1.y - vec2.y, 2));
-  }
+    /**
+    * Calculates distance between two Vectors
+    * @param vec1
+    * @param vec2
+    */
+    public static distance(vec1: Vector, vec2: Vector) {
+       return Math.sqrt(Math.pow(vec1.x - vec2.x, 2) + Math.pow(vec1.y - vec2.y, 2));
+    }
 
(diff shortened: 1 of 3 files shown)


Crizzooo:
@kamranayub Successful new branch without old commits?

src/engine/Algebra.ts
```diff
@@ -68,6 +68,15 @@ export class Vector {
       return true;
    }
 
+  /**
+  * Calculates distance between two Vectors
+  * @param vec1
+  * @param vec2
+  */
+  public static distanceBetweenVectors(vec1: Vector, vec2: Vector) {
```
kamranayub: I think the indentation might be off? Also, let's just name this `distance`, it's obviously between Vector types 👍

Crizzooo: Ah yes the indentation was off, thanks. 

Will rename as well, originally was unsure how to name as the instance method is also called distance and I wanted to avoid confusion. But I guess that is okay or even preferable?

src/engine/Algebra.ts
```diff
@@ -68,14 +68,14 @@ export class Vector {
...
 
-  /**
-  * Calculates distance between two Vectors
-  * @param vec1
-  * @param vec2
-  */
-  public static distanceBetweenVectors(vec1: Vector, vec2: Vector) {
-     return Math.sqrt(Math.pow(vec1.x - vec2.x, 2) + Math.pow(vec1.y - vec2.y, 2));
-  }
+    /**
```
kamranayub: Have to remove space so asterisks vertically align

Crizzooo:
@kamranayub Ah, I will have to look into using TSLint next time to save you the trouble. Not sure if eslint is possibly inserting spaces at the ends of lines I do not touch myself

PR #824, opened 2017-05-27
[#795] Fix bug in raycasting

### LINKED ISSUE #795: RayCasts fail when the start of a ray shares a coordinate with a BoundingBox
### Steps to Reproduce
This ray should not intersect with this bounding box 
```typescript
it('ray cast can miss a box far away', () => {
     var bb = new ex.BoundingBox(1176, 48, 1200, 72);
     var ray = new ex.Ray(new ex.Vector(48, 72), ex.Vector.Down);
     expect(bb.rayCast(ray)).toBe(false);
});
```
Basically this rayCast is incorrectly reporting intersection 

image

### Expected Result
The above test should pass, and the ray should not intersect

### Environment
N/A

### Proposal
There is a fix for this in the ludum-dare-38 branch that can be pulled into it's own PR https://github.com/excaliburjs/Excalibur/commit/113c272b197b09700ef06fcf6bac17bb167a3290

eonarheim:
Closes #795 

## Changes:

- Fix raycast bug
- Add test for raycast demonstrating bug
- Rewrite raycasts to be slightly more efficient by avoiding branching
- Add a new method to calculate time to impact in raycasts

### DIFF
--- a/src/engine/Collision/BoundingBox.ts
+++ b/src/engine/Collision/BoundingBox.ts
@@ -120,21 +120,47 @@ export class BoundingBox implements ICollidable {
       var tmin = -Infinity;
       var tmax = +Infinity;
 
-      if (ray.dir.x !== 0) {
-         var tx1 = (this.left - ray.pos.x) / ray.dir.x;
-         var tx2 = (this.right - ray.pos.x) / ray.dir.x;
-         tmin = Math.max(tmin, Math.min(tx1, tx2));
-         tmax = Math.min(tmax, Math.max(tx1, tx2));
-      }
+      var xinv = ray.dir.x === 0 ? 999999999999 : (1 / ray.dir.x);
+      var yinv = ray.dir.y === 0 ? 999999999999 : (1 / ray.dir.y);
+      
+      var tx1 = (this.left - ray.pos.x) * xinv;
+      var tx2 = (this.right - ray.pos.x) * xinv;
+      tmin = Math.min(tx1, tx2);
+      tmax = Math.max(tx1, tx2);
+            
+      var ty1 = (this.top - ray.pos.y) * yinv; 
+      var ty2 = (this.bottom - ray.pos.y) * yinv;
+      tmin = Math.max(tmin, Math.min(ty1, ty2));
+      tmax = Math.min(tmax, Math.max(ty1, ty2));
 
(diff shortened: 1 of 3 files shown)


src/engine/Collision/BoundingBox.ts
```diff
@@ -120,23 +120,46 @@ export class BoundingBox implements ICollidable {
...
-         tmax = Math.min(tmax, Math.max(tx1, tx2));
-      }
-
-      if (ray.dir.y !== 0) {
-         var ty1 = (this.top - ray.pos.y) / ray.dir.y;
-         var ty2 = (this.bottom - ray.pos.y) / ray.dir.y;
-         tmin = Math.max(tmin, Math.min(ty1, ty2));
-         tmax = Math.min(tmax, Math.max(ty1, ty2));
-      }
+      var xinv = ray.dir.x === 0 ? 999999999999 : (1 / ray.dir.x);
```
kamranayub: `Number.MAX_SAFE_INTEGER`? Or do all those 9's mean something?

eonarheim: That might be the best thing to do, the 9's were a quick and dirty way to accomplilsh the same thing. I'll fix that.

eonarheim: Looks like `MAX_SAFE_INTEGER` doesn't exist in ts 2.2.2 :(

```
Using tsc v2.2.2
src/engine/Collision/BoundingBox.ts(123,43): error TS2339: Property 'MAX_SAFE_INTEGER' does not exist on type 'NumberConstructor'.
src/engine/Collision/BoundingBox.ts(124,43): error TS2339: Property 'MAX_SAFE_INTEGER' does not exist on type 'NumberConstructor'.
src/engine/Collision/BoundingBox.ts(144,43): error TS2339: Property 'MAX_SAFE_INTEGER' does not exist on type 'NumberConstructor'.
src/engine/Collision/BoundingBox.ts(145,43): error TS2339: Property 'MAX_SAFE_INTEGER' does not exist on type 'NumberConstructor'.
```

eonarheim: I think `MAX_VALUE` should do the trick as well

eonarheim:
Tests pass with the `Number.MAX_VALUE` constant

PR #826, opened 2017-05-27
[#797] Clone when setting Vector props

### LINKED ISSUE #797: When setting Vector type setters/getters, we should implicitly clone to avoid bugs
### Context
This recently bit me during Ludum Dare 38. I assigned an `ex.Actor.pos` to a point that was part of another type which consisted of a grid of `ex.Vector`'s 

The problem can be easily demonstrated thus
```typescript
var actor1 = new ex.Actor();

var actor2 = new ex.Actor();

var point = new ex.Vector(10, 10);

actor1.pos = point;
actor2.pos = point;

actor1.moveTo(20, 100);
// BOTH ACTOR1 and ACTOR2 move!!!
// they share the pos reference
```
The reason this happens, is because Vector (even though I want to treat it like a value type) is actually a reference type, and the reference in this case is shared causing problems and tricky bugs to find since you'd expect mathematically for points to be just a value.

### Proposal

I think we should implicitly call the `.clone()` on vector when setters and getters take or emit `ex.Vector` anywhere in excalibur, I think this will eliminate a class of hard to track down bugs when working with actors. 

You do sacrifice the ability to retrieve the reference of the Vector and treat it as one and do tricky things with it and squeeze perfomance out of it. However, I'd argue that people should be doing things like that with things that are conceptually value types, it would be like retrieving a reference to the number 5 and changing it's value for people using it to 8 or something.

```typescript
/**
 * Gets the position vector of the actor in pixels
 */
public get pos(): Vector {
   return this.body.pos.clone();
}

/**
 * Sets the position vector of the actor in pixels
 */
public set pos(thePos: Vector) {
   this.body.pos = thePos.clone();
}
```

kamranayub:
Closes #797

## Changes:

- Clone when setting Vector props on Actor

### DIFF
--- a/src/engine/Actor.ts
+++ b/src/engine/Actor.ts
@@ -106,7 +106,7 @@ export class Actor extends Class implements IActionable, IEvented {
     * Sets the position vector of the actor in pixels
     */
    public set pos(thePos: Vector) {
-      this.body.pos = thePos.clone();
+      this.body.pos.setTo(thePos.x, thePos.y);
    }
 
    /**
@@ -120,7 +120,7 @@ export class Actor extends Class implements IActionable, IEvented {
     * Sets the position vector of the actor in the last frame
     */
    public set oldPos(thePos: Vector) {
-      this.body.oldPos = thePos.clone();
+      this.body.oldPos.setTo(thePos.x, thePos.y);
    }
 
    /**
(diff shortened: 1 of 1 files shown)


kamranayub:
@eonarheim I didn't clone on getters because the only time using the same reference causes a problem is when you set a property to that instance. So, I think setters should clone but I think it's safe to leave getters alone.

eonarheim:
@kamranayub I agree, that'll cut down on object thrashing in garbage collection

PR #850, opened 2017-06-27
[#846] Add backgroundColor to IEngineOptions

### LINKED ISSUE #846: Add `backgroundColor` to IEngineOptions
### Context

You can set the `backgroundColor` property on an instance of engine but cannot pass it in via the constructor options.

```
var game = new ex.Engine({
    width: 1280,
    height: 720    
});
game.backgroundColor = ex.Color.White;
```

### Proposal

Add `backgroundColor` to `IEngineOptions` and allow setting it that way. It should be as easy as adding a new property and then setting it in the Engine constructor.

```
var game = new ex.Engine({
    width: 1280,
    height: 720,
    backgroundColor: ex.Color.White
});
```

htalat:
Add backgroundColor property to the Engine constructor options and use ex.Color.Blue as default if none is provided.

Closes #846 

## Changes:

- Add `backgroundColor` to `IEngineOptions` as an optional field.
- Use `ex.Color.Blue` as the default `backgroundColor`.
- Set the `backgroundColor` in the constructor `from IEngineOptions`.
- Add a test in the Engine Specs.


kamranayub:
Thanks @htalat this is great. This is also a good opportunity to maybe switch to using "excalibur blue" brand color. Do you want to change it to default to `ex.Color.fromHex("#2185d0")`?

htalat:
sure!

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Breaking Changes
 ### Added
+- Added `backgroundColor` `ex.Engine` constructor, with `ex.Color.Blue` as the default, and updated test specs ((#846) [https://github.com/excaliburjs/Excalibur/issues/846])
```
kamranayub: 1. I would move this to under Changed because it's not a "new" feature per se
2. I would rephrase as "Added `backgroundColor` to engine options as part of `Engine` constructor"

No need to mention default color (didn't really change) or test specs (end users don't care)

src/engine/Engine.ts
```diff
@@ -401,6 +408,9 @@ O|===|* >________________>\n\
          this.displayMode = DisplayMode.FullScreen;
       }
 
+      if (options.backgroundColor) {
+         this.backgroundColor = options.backgroundColor;
```
kamranayub: Let's do a `options.backgroundColor.clone()` to ensure we pass a clone of the instance to avoid issues later.

src/spec/EngineSpec.ts
```diff
@@ -121,4 +121,23 @@ describe('The engine', () => {
...
    });
    
+   it('should accept backgroundColor', () => {
+      var game = new ex.Engine({
+        height: 600,
+        width: 800,
+        backgroundColor: ex.Color.White
+      });
+
+      expect(game.backgroundColor).toEqual(ex.Color.White);
```
kamranayub: Because we pass a clone, this will fail I expect. If it doesn't, ignore this. You may need to verify the string representation `toString()` or the individual color channels.

`expect(game.backgroundColor.toString()).toEqual(ex.Color.White.toString());

kamranayub:
Ah all your code looks good but this affected our image tests. If you run the tests you will see some base64 text that represents the actual images. Do you mind just copy/pasting those base64 strings into Chrome and saving the new images over the old ones (assuming all that's changed is the bg)?

htalat:
ahh okay,

htalat:
this should be it?

kamranayub:
Thanks @htalat, once it's ready and someone else gives it a once over we'll merge it. Appreciate it! 💯

htalat:
Great!

PR #855, opened 2017-06-30
[#852] Add `Color.toHex()`, `toRGBA()`, `toHSLA()`

### LINKED ISSUE #852: Add `Color.toHex()`, `toRGBA()`, and `toHSLA()` methods
### Context

Color supports converting *from* Hex/HSL representation but not back to them (only RGBA).

### Proposal

Modify Color class:

1. Add a new instance method `toHex()`.
2. Add a new instance method `toRGBA()`
3. Add a new instance method `toHSLA()`
3. Add an optional param `type` to `toString()` that accepts a string, either `rgb`, `hsl` or `hex` parameter (default: "rgb" to maintain backwards compat). Depending on value, it should call either of the three new instance methods.

htalat:
Closes #852 

## Changes:

- included a optional param `format` in the `toString()` method.
- added `toHex()`, `toHSLA()` (which basically calls the new `toString()` method in class `HSLColor`) , `toRGBA()`
- included the appropriate tests in ColorSpec.ts


src/engine/Drawing/Color.ts
```diff
@@ -188,8 +188,46 @@ export class Color {
...
+         return this.toHex();
+      }
+
+      throw new Error('Invalid Color format');
+   }
+
+   /**
+    * Returns Hex Value of a color component
+    * @param c color component
+    * https://stackoverflow.com/questions/5623838/rgb-to-hex-and-hex-to-rgb
```
kamranayub: Prefix with `@see`

src/engine/Drawing/Color.ts
```diff
@@ -198,6 +236,14 @@ export class Color {
    }
 
    /**
+    * Return HSLA representation of a color.
+    */
+   public toHSLA() {
+      var temp = HSLColor.fromRGBA(this.r, this.g, this.b, this.a);
+      return temp.toString();
```
kamranayub: Can combine into single return I think

src/engine/Drawing/Color.ts
```diff
@@ -366,4 +412,10 @@ class HSLColor {
 
       return new Color(r * 255, g * 255, b * 255, this.a);
    }
+
+   public toString(): string {
+      var result = String(this.h.toFixed(0)) + ', ' + String(this.s.toFixed(0)) + ', ' + String(this.l.toFixed(0))
```
kamranayub: I know this is similar to the rgba one, but I think string templating will help make this clearer:

```ts
let h = this.h.toFixed(0),
    s = this.s.toFixed(0),
    l = this.l.toFixed(0),
    a = this.a.toFixed(0);

return `hsla(${h}, ${s}, ${l}, ${a})`;
```

src/spec/ColorSpec.ts
```diff
@@ -17,6 +17,18 @@ describe('A color', () => {
...
+   it('should display hex values #000000', () => {
+      expect(color.toString('hex')).toBe('#000000');
+   });
+
+   it('should display hsla(0,0,0,1)', () => {
+      expect(color.toString('hsl')).toBe('hsla(0, 0, 0, 1)');
+   });
+
+   it('should display an error message', () => {
+      expect(color.toString('invalid')).toThrow(new Error('Invalid Color format'));
```
kamranayub: Once you add type unions this will be a compiler error, you can force ignore it by doing `'invalid' as any` I think.

src/engine/Drawing/Color.ts
```diff
@@ -188,8 +188,46 @@ export class Color {
 
    /**
     * Returns a CSS string representation of a color. 
+    *
+    * @param format Color representation
+    */
+   public toString(format = 'rgb') {
```
htalat: oh, didn't know we can do this.
makes things much cleaner!

src/spec/ColorSpec.ts
htalat: compiles with any.
Error is thrown correctly but the test fails.

   - should display an error message......X
   Error: Invalid Color format in file:///<path>/Excalibur/.grunt/grunt-contrib-jasmine/build/dist/excalibur.js (line 9) (1)

Also tried using `.toThrowError('Invalid Color format');`

htalat:
so the Spec compiles with any.
However, Error is thrown correctly but the test fails.

   - should display an error message......X
   Error: Invalid Color format in file:///<path>/Excalibur/.grunt/grunt-contrib-jasmine/build/dist/excalibur.js (line 9) (1)

Also tried using `.toThrowError('Invalid Color format');`

also how can i supress warnings when im runnnig a grunt task e.g. `npm run all`

kamranayub:
@htalat ah yeah, so you can't expect one instance of Error to equal another. What you can do is use `toThrowError` like this:

```ts
var fn = () => color.toString('foo');

expect(fn).toThrowError('whatever')
```

The expectation has to be made on a call to the function. Jasmine will then invoke it and expect the error.

For the warnings, you can't. We have an open issue to quiet those down 👍

htalat:
ahhh!
thanks for clarifying that!

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Changed `ex.Scene.children` to be `ex.Scene.actors`, and updated unit test specs to reflect this change ((#796)[https://github.com/excaliburjs/Excalibur/issues/796])
 - Added backgroundColor to engine options as part of Engine constructor ((#846)[https://github.com/excaliburjs/Excalibur/issues/846])
+- Added options for hsl,hex representation in Color.toString() method with rgb as default if none provided to maintain backward compatability((#852)[https://github.com/excaliburjs/Excalibur/issues/852])
```
kamranayub: Since this adds new output formats, it might be a better fit for Added:

- Added new `hsl` and `hex` format options in `Color.toString(format)` using `rgb` as the default to maintain backwards compatibility ([#852](https://github.com/excaliburjs/Excalibur/issues/852))

I think the other issue links above are using wrong Markdown syntax too, it's `[text](link)` instead of the other way.

htalat: oh my bad.
I just followed the one on line 13, should i fix line 13,14 as well then?

src/engine/Drawing/Color.ts
```diff
@@ -188,8 +188,43 @@ export class Color {
 
    /**
     * Returns a CSS string representation of a color. 
+    *
+    * @param format Color representation, accepts: rgb, hsl, or hex
+    */
+   public toString(format: 'rgb'|'hsl'|'hex' = 'rgb') {
```
eonarheim: Super cool! Type unions are great!

PR #878, opened 2017-09-04
Update to Node v8.4.0

kamranayub:
## Changes:

- Update to Node v8

## Notes

- [x] Ready to merge
- [ ] Wait

Performance of npm and Node itself is a lot better in v8.


GruntFile.js
```diff
@@ -291,6 +291,17 @@ module.exports = function (grunt) {
...
+      },
+
+      connect: {
+         sandbox: {
+            options: {
+               port: '3001',
+               useAvailablePort: true,
+               keepalive: true,
+               base: './sandbox'
+            }
```
eonarheim: Nice! Will this keep us from having to spin up a local server every type we want to test sandbox???? 🥇

kamranayub:
Going to remove shrinkpack for now because it cannot be used yet with npm 5. Since package-lock (shrinkwrap.json) and modules are cached on CI, I don't see a big issue living without shrinkpack for awhile.

GruntFile.js
kamranayub: YEE 💯

kamranayub: And can use to do functional tests if needed.

PR #882, opened 2017-09-12
[#840] Implicitly clone a color when setting it

### LINKED ISSUE #840: Implicitly clone a Color when setting it
### Context
When setting an ex.Color on an Actor, Excalibur does not implicitly clone the Color. This currently has the potential to cause confusion if multiple Actors are set to the same color, and then graphics transformations are applied that affect that Color (for example, changing the alpha channel value).

(see also #797)

### Proposal
Clone ex.Color when it is set, to avoid potentially unintended effects across multiple Actors.

Nysosis:
Closes #840 

## Changes:

- Made `color` on `Actor` have a getter/setter, which implicitly clones the color value when set

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Deprecated
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


eonarheim:
@Nysosis Thanks for the help!

Nysosis:
No problem!

PR #889, opened 2017-09-29
[#612] Add position vector to PointerEvent

### LINKED ISSUE #612: Add position vector to ex.Input.PointerEvent

jedeen:
Closes #612 

## Changes:

- mark `x` and `y` parameters as obsolete
- add `pos` Vector that will replace `x` and `y`

### DIFF
--- a/src/engine/Input/Pointer.ts
+++ b/src/engine/Input/Pointer.ts
@@ -66,9 +66,11 @@ const ScrollWheelNormalizationFactor = -1 / 40;
  */
 export class PointerEvent extends GameEvent<any> {
 
+   public pos: Vector;
+
    /**
-    * @param x            The `x` coordinate of the event (in world coordinates)
-    * @param y            The `y` coordinate of the event (in world coordinates)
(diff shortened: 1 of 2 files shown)


src/engine/Input/Pointer.ts
```diff
@@ -89,6 +92,7 @@ export class PointerEvent extends GameEvent<any> {
                public button: PointerButton,
                public ev: any) {
       super();
+      this.pos = new Vector(x, y);
```
eonarheim: There should be 1 source of truth for x,y.

x,y can be set independently from pos.

Either let's make x/y getters that ref pos, or make pos a getter that references x/y

PR #892, opened 2017-09-30
[#745] Implement `once` event listener

### LINKED ISSUE #745: EventDispatcher should allow `one` single-use handler
### Context

In the scenario of a unit test, for example, it would be nice to allow subscribing a handler to an event **once** and then auto-unsubscribing it.

```
engine.on('postdraw', () => {
  engine.off('postdraw');
  imagediff.expectCanvasImageMatches('UIActorSpec/emptyctor.png', engine.canvas, done);
});
```

Especially in the above context, as the canvas matcher is asynchronous and the handler will be called next frame which will again trigger a matcher call. In this instance, we need to unsubscribe immediately.

### Proposal

Add a new method `one(eventName: string, handler: Function)` to EventDispatcher.

See: http://api.jquery.com/one/

eonarheim:
Closes #745

## Changes:

- Implements a new way to listen to events `.once(...)` and then unsubscribe
- Additional tests

### DIFF
--- a/src/engine/EventDispatcher.ts
+++ b/src/engine/EventDispatcher.ts
@@ -107,7 +107,7 @@ export class EventDispatcher implements IEvented {
    }
 
    /**
-    * Once listens to an event once then auto unsubscribes from that event
+    * Once listens to an event one time, then unsubscribes from that event
     *
     * @param eventName The name of the event to subscribe to once
     * @param handler   The handler of the event that will be auto unsubscribed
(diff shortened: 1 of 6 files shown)


src/engine/EventDispatcher.ts
```diff
@@ -107,6 +107,25 @@ export class EventDispatcher implements IEvented {
    }
 
    /**
+    * Once listens to an event once then auto unsubscribes from that event
```
jedeen: Perhaps "Once listens to an event one time, then unsubscribes from that event"

PR #902, opened 2017-10-17
[#863] Update to Triggers

### LINKED ISSUE #863: Triggers aren't the best
### Context
Currently it is tough to do anything with triggers that is useful, I suggest revamping the current implemenation so that it supports an option bag constructor, conditional filtering on triggering, and a friendlier interface.

### Proposal
Something like this would be a great start
```typescript
interface ITriggerOptions {
   pos?: ex.Vector;
   width?: number;
   height?: number;
   // action to take when triggered
   trigger?: () => void;
   // Returns true if the triggers should fire on the collided actor
   filter?: (actor: ex.Actor) => boolean;
   // -1 if it should repeat forever
   repeat?: number;
}

class FilteredTrigger extends ex.Trigger {
   public filter: (actor: ex.Actor) => boolean = () => true;
   constructor(opts: ITriggerOptions ) {
      super(opts.pos.x, opts.pos.y, opts.width, opts.height, opts.trigger, opts.repeat);
   }

   public collides(other: ex.Actor) {
      if (this.filter(other)) {
         return super.collides(other);
      }
      return null;
   }
}
```

eonarheim:
### Work in progress

Needs:
- [x] More discussion about engine lifecycle
- [x] Collision pair refactoring
- [x] Expose collision `wasTouching` and `justTouching` as something less confusing
- [x] Tests

Closes #863

## Changes:

- New Enter/Exit trigger event
- Triggers are now `Passive` collision type so we can make use of the collision system
- Scene graph ordering to push vanilla actors to the front


eonarheim:
@kamranayub @jedeen Ready for review

sandbox/src/game.ts
```diff
@@ -433,20 +433,39 @@ game.add(emitter);
...
-var trigger = new ex.Trigger(400, 200, 100, 100, () => {
-   if (!exploding) {
-      exploding = true;
-      emitter.isEmitting = true;
-      camera.shake(10, 10, 2000);
-      game.addTimer(new ex.Timer(() => {
-         emitter.isEmitting = false;
-         exploding = false;
-      }, 2000));
+// var trigger = new ex.Trigger(400, 200, 100, 100, () => {
```
kamranayub: Did you intend to keep this for reference?

eonarheim: Yah, I wanted to make sure it behaved the same. I can remove this if you want

kamranayub:
If we don't plan to reference it then maybe we can delete it

eonarheim:
@kamranayub updated

src/spec/TriggerSpec.ts
```diff
@@ -0,0 +1,297 @@
...
+      // Act
+      for (let i = 0; i < 2; i++) {
+         loop.advance(1000);
+      }
+
+      // Assert
+      expect(trigger.action).not.toHaveBeenCalled();
+   });
+
+   it('will not only trigger if the filter is true', () => {
```
jedeen: Is this meant to be it('will only trigger if the filter is true' ?

eonarheim: Yup that's correct, copy paste typo

eonarheim:
@jedeen For sure, I definitely think smaller PRs to break things up would be good in the future.

src/engine/Trigger.ts
```diff
@@ -3,82 +3,126 @@ import { Engine } from './Engine';
...
+};
+
 
 /**
  * Triggers are a method of firing arbitrary code on collision. These are useful
  * as 'buttons', 'switches', or to trigger effects in a game. By default triggers
- * are invisible, and can only be seen when [[Engine.isDebug]] is set to `true`.
+ * are invisible, and can only be seen when [[Trigger.visible]] is set to `true`.
  * 
  * [[include:Triggers.md]]
```
kamranayub: Can you ensure `Triggers.md` is up-to-date too?

kamranayub: Feel free to dismiss this and merge if it's already been updated or N/A

PR #909, opened 2017-11-18
[#908] Fix BaseCamera documentation

### LINKED ISSUE #908: Documentation on ex.BaseCamera is wrong for `setActorToFollow`
### Steps to Reproduce
See this google groups thread: https://groups.google.com/forum/#!topic/excaliburjs/KqqqgCw-B14

BaseCamera.setActorToFollow does not actually get the camera to follow an actor must replace the current camera implementation with `ex.LockedCamera` for it to work.

```typescript
game.currentScene.camera = new ex.LockedCamera();
game.currentScene.camera.setActorToFollow(actor);
```
See pen with working example on latest ex:
https://codepen.io/eonarheim/pen/gXogGp?editors=0010

### Expected Result
BaseCamera.setActorToFollow documenation should outline the process to get this to work, or we should deprecate it. Or update the BaseCamera implementation to support this.

### Actual Result
Following the documentation did not work

eonarheim:
Closes #908

## Changes:

- Updates documentation to clarify current process
- Kill some dead code in `BaseCamera`
- Removes `setActorToFollow` method on `BaseCamera` that doesn't work when set and is confusing

### DIFF
--- a/src/engine/Camera.ts
+++ b/src/engine/Camera.ts
@@ -51,7 +51,6 @@ export class BaseCamera {
 
    protected _isZooming: boolean = false;
    private _maxZoomScale: number = 1;
-   private _zoomDuration: number = 0;
    private _zoomPromise: Promise<boolean>;
    private _zoomIncrement: number = 0.01;
    private _easing: EasingFunction = EasingFunctions.EaseInOutCubic;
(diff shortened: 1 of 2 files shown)


src/engine/Camera.ts
```diff
@@ -88,14 +87,6 @@ export class BaseCamera {
       }
    }
 
-   /**
```
jedeen: This is a breaking change, and should be documented in the changelog.

eonarheim: Actually no, it was dead code, did nothing

PR #921, opened 2017-12-20
[#920] Use setTargetAtTime to set volume for Sound

### LINKED ISSUE #920: Sound.setVolume uses a deprecated (soon to be eliminated) API
### Context
When you call `Sound.setVolume`, you'll see a deprecation warning printed to the console stating that the backing WebAudio method will be removed sometime around January 2018.
deprecation

### Proposal
Update / correct the backing method per that warning message's suggestion, but keep the original call as a fallback to ensure browser compatibility.

alanag13:
Closes #920

## Changes:
- Use `GainNode.gain.setTargetAtTime` as recommended instead of setting `GainNode.gain.value` directly where possible

### DIFF
--- a/src/engine/Resources/Sound.ts
+++ b/src/engine/Resources/Sound.ts
@@ -567,7 +567,11 @@ class WebAudioInstance implements IAudio {
 
    public setVolume(value: number) {
       this._volume = value;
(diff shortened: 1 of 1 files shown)


eonarheim:
Tested locally works

Issue #305, opened 2014-05-06
Add Missing test files


jedeen:
🔨 🔧 ❗️  
This is a collaborative effort to increase the valuable testing coverage for Excalibur. Any time you are working within the files and methods below, please add useful tests.
- [x] Algebra
- [ ] Scene onInitialize, onActivate, onDeactivate
- [ ] Cullbox.prototype.isSpriteOffScreen
- [ ] Cullbox.prototype.update whe offscreen
- [ ] Capture pointer update when is killed
- [ ] TileMapCollisionDetection update
- [ ] Vector normalize on zero length vectors
- [ ] Vector normals
- [ ] Ray non intersect
- [ ] base64encode (maybe remove?)
- [ ] drawLine
- [ ] randomIntInRange
- [ ] getPosition
- [ ] removeItemToArray
- [x] Sprites
- [x] SpriteSheet
- [ ] SpriteFont
- [ ] TileMap, Cell
- [x] SATBoundingBox (remove)
- [ ] Class
- [x] Timer.getTimeRunning
- [ ] NaiveCollisionResovler
- [ ] DynamicAABBTree
- [ ] EaseTo action
- [ ] RotateBy Type CounterClockwise
- [x] Fade action
- [ ] CallMethod
- [ ] Repeat
- [ ] ActionQueue remove
- [ ] ActionContext all actions
- [ ] SortedList _cleanUp
- [ ] Easing functions
- [x] predraw, postdraw, preupdate postupdate
- [ ] EventDispatcher unsubscribe
- [ ] Color effects (lighten, darken, etc)
- [ ] HSLColor fromRGBA
- [ ] UIActor contains
- [x] Trigger
- [x] Particles
- [x] Animation
- [x] FallbackAudio
- [x] Sound
- [ ] Resource load
- [ ] Loader
- [x] Texture load
- [ ] Detector
- [x] Template/Binding (remove maybe?)
- [x] Label
- [ ] Label#getTextWidth (#679)
- [x] Pointers
- [ ] Engine

eonarheim:
We should look at https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md to ignore things that can't be tested

eonarheim:
This is fairly out of date, and coverage is pretty good

PR #934, opened 2018-02-26
[#918] Add support for packed SpriteSheets

### LINKED ISSUE #918: Add feature to pull Animations out of a packed SpriteSheet
### Context

Currently Excalibur sprite sheets only optimize for neatly arranged in rows and columns when extracting animations. Many assets now are arranged into tightly packed sprite sheets to save on space and download size. 

### Proposal
Excalibur `ex.SpriteSheet` should support a way to specify a list of bespoke sprite coordinates to generate an `ex.Animation`. 

Phil on the forums posted this motivation which makes a lot of sense

> Most of the image resources I have are not aligned in rows and columns but are tightly packed, e.g. lots of resources from kenney.nl. This means I can't use them to create an ex.Spritesheet. Is there any way to create an animation from the single sprites from a texture? E.g. pass an array of {x,y,w,h} objects instead of indices on getAnimationByIndices?

A possible api might look like this

```typescript

export interface ISpriteCoordinates {
   x: number;
   y: number;
   width: number;
   height: number;
}

// create sprite sheet with 5 columns, 1 row, 80x80 frames
let playerIdleSheet = new ex.SpriteSheet(txAnimPlayerIdle);

  // create animation (125ms frame speed)
let playerIdleAnimation = playerIdleSheet.getAnimationByCoords({x: 10, y: 10, width: 100, height: 100},{x: 120, y: 10, width: 100, height: 100},{x: 120, y: 10, width: 100, height: 100},);

```

eonarheim:
Closes #918

## Changes:

- Allows bespoke sprite coordinates to pull animations out of packed sprite sheets for example http://kenney.nl/assets/generic-items
- Deprecates confusing sx, sy, swidth, and sheight properties on sprites
- Adds tests for spritesheet coordinates
- Adds missing tests for sprites #305
- Update changelog


src/engine/Drawing/SpriteSheet.ts
```diff
@@ -133,6 +133,35 @@ export class SpriteSheetImpl {
...
+
+   /**
+    * Get an animation with bespoke sprite coordinates. This is useful if the SpriteSheet is
+    * packed and not a uniform width or height. The resulting [[Animation]] will have the height and width of the
+    * largest dimension (width, height) from among the sprite coordinates
+    * @param engine 
+    * @param spriteCoordinates 
+    * @param speed 
+    */
+   public getSpriteByCoords(engine: Engine, spriteCoordinates: ISpriteArgs[], speed: number): Animation {
```
alanag13: I would expect a method with this name to return a `Sprite`, not an `Animation`.  

I would probably change this method to just return a `Sprite` array, then add a static method (or perhaps constructor) to `Animation` that creates an `Animation` from an arbitrary `Sprite` array.

alanag13: on second look, it looks like that constructor already exists -- so just changes this to return a `Sprite` array.

alanag13: Scratch all that, I just saw this `SpriteSheet`, not `Sprite`. just rename this to `getAnimationByCoords`.

eonarheim: @alanag13 Updated per feedback

PR #935, opened 2018-02-26
[#582] Make overriding lifecycle and event hooks safe and typesafe

### LINKED ISSUE #582: Overriding onInitialize, onActivate, onEventType is not safe
### Context

Currently we have "eventish" like method overrides like `onInitialize` and `onActivate` that don't have a corresponding `.on('initialize')` or `.on('activate')` event handlers and vice versa, this is not consistent. Additionally it is often not safe to override `onInitialize` implementations without the base implementations.

For example, when overriding OnInitialize, currently you have to manually call super.OnInitialize to get the default UIActor OnInitialize to trigger.

### Proposal

Follow .NET's pattern and create `initialize` implementation that does the base work, and calls the overridable `onInitialize` when complete.

Also for every `onEventType` method, have a corresponding `on('eventtype')` and vice versa.

eonarheim:
This is a work in progress, please feel free to comment

Closes #582 

## Changes:

- Added lifecycle interfaces to describe flows
- Refactor types so `onEventType` is always safe to override
- Added tests


kamranayub:
So, question for @excaliburjs/core-contributors 

I really like how we started with `draw` and `update`. I don't like how we forced people to call `super.draw` or `super.update` because even *we* forget.

Can I propose to keep `predraw` and `preupdate` but then just call the other ones `draw` and `update` again? I think with fresh eyes, it's fairly clear that hooking into those events will be after the core engine does it's thing. If we are worried about what to call the original methods, they could just be renamed to `coredraw` and `coreupdate`.

We know that in 80% of cases, folks will use the "post" methods when implementing their own logic. The case where you need to totally change the core logic is less common and so I would prioritize making the 80 case method names easy to remember.

Thoughts/concerns?

alanag13:
I agree with @kamranayub, also for consistency's sake I think we should add "pre" and "post" events for activate and initialize too.

So for all events it would be like

Overrideable preEvent
Internal coreEvent
Overrideable Event

kamranayub:
Also, should the overridable pre- and post-event methods be marked as `protected` or would we expect them to be called outside the class inheritance chain?

eonarheim:
@excaliburjs/core-contributors In an effort to keep this PR small I'm focusing on Actor at the moment. Let me know what you think so far.

eonarheim:
Also if there are any life cycle or event things I'm missing on actor that we want to add let me know

src/engine/Actor.ts
```diff
@@ -1060,17 +1154,61 @@ export class ActorImpl extends Class implements IActionable, IEvented {
       this.integrate(delta);
 
       // Update actor pipeline (movement, collision detection, event propagation, offscreen culling)
-      for (var trait of this.traits) {
+      for (let trait of this.traits) {
```
kamranayub: Should these be `const`?

src/engine/Scene.ts
```diff
@@ -161,13 +207,45 @@ export class Scene extends Class {
       }
    }
 
+   /**
+    * Activates the scene with the base behavior, then calls the overridable `onActivate` implementation.
+    */
+   public _activate(): void {
```
kamranayub: Mark all these as `@internal`?

src/spec/ActorSpec.ts
```diff
@@ -1250,4 +1255,131 @@ describe('A game actor', () => {
       expect(numPointerUps).toBe(1, 'Pointer up should be triggered once');
       expect(propSpy).toHaveBeenCalledTimes(1);
    });
+
+   describe('lifecycle overrides', () => {
+
+      let actor: ex.Actor = null;
+      // let engine: ex.Engine = null;
```
kamranayub: Can remove commented code

src/engine/Actor.ts
eonarheim: They can be, we should actually add a ts-lint rule eventually to force const where appropriate

eonarheim:
@excaliburjs/core-contributors This feels done, let me know what you find. I've run things locally as well to give it an extra check, feels good :shipit:

src/engine/Engine.ts
```diff
@@ -1058,6 +1103,22 @@ O|===|* >________________>\n\
    }
 
 
+   /**
+    * Gets wether the actor is Initialized
```
jedeen: 'whether'

PR #957, opened 2018-04-08
[bugfix] Fix missing variable assignments in TileMapImpl

jamesmcgeorge:
Add missing variable assignments to constructor


eonarheim:
@excaliburjs/core-contributors Ran this against the visual tests, everything is good

@jamesmcgeorge Thanks for the assist! An alpha package should be built automatically soon if you want to use that until a release can be made

eonarheim:
npm install 0.16.0-alpha.2218

PR #958, opened 2018-04-13
[chore][docs] Pass a ILoadable[] instead of ILoadable to constructor

rougepied:


rougepied:
The Loader constructor is expecting a ILoadable[]. Another solution would be to use addResource.

```ts
var game = new ex.Engine();
var txPlayer = new ex.Texture("/assets/tx/player.png");
// load assets
var loader = new ex.Loader();
loader.addResource(txPlayer);

// start game
game.start(loader).then(function () {

  // create a sprite (quick)
  var playerSprite = txPlayer.asSprite();
  // create a sprite (custom)
  var playerSprite = new ex.Sprite(txPlayer, 0, 0, 80, 80);
});
```

eonarheim:
@rougepied we can stick with the more terse constructor example.

PR #963, opened 2018-04-19
[bugfix] Prevent negative scaling in hidpi/non-hidpi hybrid scenarios

eonarheim:
Currently Chrome and possibly other browsers can incorrectly report the scaling as less than a pixel, which never makes sense. This seems to happen if you have monitors that are HiDPI and non-HiDPI connected at the same time. This change forces scaling to be at least 1 to prevent odd visual artifacts.

## Changes:

- Force the minimum dpi scale to be at least 1


eonarheim:
@jedeen Updated

Issue #884, opened 2017-09-19
Add Prettier


kamranayub:
### Context

[Prettier](https://github.com/prettier/prettier) auto-formats code before commit (or on save) to code standards set for the project. This should make code style a non-issue for new contributors and for the team.

### Proposal

Implement Prettier with our current TSLint rules.

1. Add the git hook using husky, as outlined in their README
2. Recommend using Prettier VSCode plug-in (optional) in our README
3. Remove lint step from grunt

htalat:
can i have a go at this one?

eonarheim:
@htalat For sure! I've assigned myself to prevent anyone else from grabbing this

htalat:
so i have set up prettier but tslint wont run.
is it just me getting this error on running tslint?

```
Warning:
            Could not find implementations for the following rules specified in the configuration:
            array-bracket-spacing
block-spacing
brace-style
conditional-expression-parens
import-destructuring-spacing
import-spacing
jsx-alignment
jsx-attribute-spacing
jsx-curly-spacing
jsx-equals-spacing
jsx-expression-spacing
jsx-no-closing-bracket-newline
jsx-no-multiline-js
jsx-wrap-multiline
literal-spacing
no-empty-line-after-opening-brace
no-irregular-whitespace
no-multi-spaces
no-semicolon-interface
no-unnecessary-parens-for-arrow-function-arguments
no-unnecessary-semicolons
object-curly-spacing
react-tsx-curly-spacing
space-before-function-paren
space-in-parens
space-within-parens
ter-arrow-parens
ter-arrow-spacing
ter-func-call-spacing
ter-indent
ter-max-len
ter-no-irregular-whitespace
            Try upgrading TSLint and/or ensuring that you have all necessary custom rules installed.
            If TSLint was recently upgraded, you may have old rules configured which need to be cleaned up.
         Use --force to continue.
```

kamranayub:
Tslint runs normally, otherwise we'd notice builds failing.

eonarheim:
Hi @htalat, have you run an npm install after cloning/forking? 

I double checked a fresh clone of master that tslint is running

```
git clone https://github.com/excaliburjs/Excalibur.git test-clone
cd test-clone
npm install
npm run all
```

The tslint portion runs for me and outputs 
```
Local build, using version "local"
Running "tslint:src" (tslint) task
>> 130 files lint free.
```

htalat:
[...]

PR #971, opened 2018-04-20
[#912] Add stopping PointerEvent from propagating

### LINKED ISSUE #912: Stop Pointer-Event from propagating
### Context
If there are 2 actors overlapping and both listen to pointer events and the overlapping area is clicked, both receive the click event. This is not always desirable.

Demo: https://codepen.io/piko_/pen/ooPNaB

### Proposal
It would be good to be able to stop the propagation of input events.

LoserAntbear:
Closes #912 

## Changes:

- Adds PointerEvent.stopPropagation() functionality that behaves as described at https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation.
- `jasmine.d.ts` type definitions are now for Jasmine 2.8.0. It equals to the version used in current development.
- Added `jasmine-core` package from `v2.8.0` accordingly to version it should use.
- Event propagation made less greedy.
- Class `BubblingEvent` added which extends `GameEvent`.
- Property scope `Pointer.actorsUnderPointer` changed to private.
- New interface `IActorUnderPointer` added, which provides `length` property to old untyped `Pointer.actorsUnderPointer` object
- `CapturePointer.update()` method now doesn't propagate event to actor, just verifies pointer events for actor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,15 +11,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Added
  - New `PointerEvent.stopPropagation()` method added. Works the same way as    (https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation)
     ([#912](https://github.com/excaliburjs/Excalibur/issues/912))
- - New `Engine.input.pointers.verifyPointerEvents(actor)` method, which checks if event can be fired on provided Actor
  - New abstract Class `BubblingEvent` which extends `GameEvent` and provides event with bubbling functionality
  - New `Actor.getAncestors()` method, which retreives full array of current Actors` ancestors
- - Added better functionality for working with actors under pointer:
-   - New method `addActorUnderPointer(actor)`
-   - New method `removeActorUnderPointer(actor)`
-   - New method `hasActorUnderPointer(actor)`
-   - New property `hasActorsUnderPointer`
-   * New interface `IActorUnderPointer` which adds `length` property to old untyped object
 
 ## Changed
  - `CapturePointer.update()` method now doesn't propagate event to actor, just verifies pointer events for actor.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   -New `Actor.getAncestors()` method, which retreives full array of current Actor ancestors
 
 ## Changed
+- Refactored Easing functions to be reversable ([#944](https://github.com/excaliburjs/Excalibur/pull/944))
 
   -`CapturePointer.update()` method now doesn't propagate event to actor, just verifies pointer events for actor.
 
@@ -29,6 +30,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Correct setting audio volume level from `value` to `setValueAtTime` to comply with deprecation warning in Chrome 59 ([#953](https://github.com/excaliburjs/Excalibur/pull/953))
 - Force HiDPI scaling to always be at least 1 to prevent visual artifacts in some browsers
 - Recalculate physics geometry when width/height change on Actor ([#948](https://github.com/excaliburjs/Excalibur/pull/948))
+- Fix camera move chaining ([#944](https://github.com/excaliburjs/Excalibur/pull/944))
 
 <!--------------------------------- DO NOT EDIT BELOW THIS LINE --------------------------------->
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
  - New typesafe and override safe event lifecycle overriding, all `onEventName` handlers will no longer be dangerous to override ([#582](https://github.com/excaliburjs/Excalibur/issues/582))
    * New lifecycle event `onPreKill` and `onPostKill`
  - SpriteSheets can now produce animations from bespoke sprite coordinates `SpriteSheet.getAnimationByCoords(engine, coords[], speed)` ([#918](https://github.com/excaliburjs/Excalibur/issues/918))
- - Added drag and drop suppor for Actors ([#134](https://github.com/excaliburjs/Excalibur/issues/134))
+ - Added drag and drop support for Actors ([#134](https://github.com/excaliburjs/Excalibur/issues/134))
    - New Event `enter`
    - New Event `leave`
    - New Event `pointerenter`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,19 +11,13 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Added
 - Allow timers to limit repeats to a finite number of times ([#957](https://github.com/excaliburjs/Excalibur/pull/974))
-
-  -New `PointerEvent.stopPropagation()` method added. Works the same way as    (`https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation`)
+- New `PointerEvent.stopPropagation()` method added. Works the same way as    (`https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation`)
     ([#912](https://github.com/excaliburjs/Excalibur/issues/912))
-  -New `Actor.getAncestors()` method, which retreives full array of current Actor ancestors
+- New `Actor.getAncestors()` method, which retreives full array of current Actor ancestors
 
 ## Changed
 - Refactored Easing functions to be reversable ([#944](https://github.com/excaliburjs/Excalibur/pull/944))
-
-  -`CapturePointer.update()` method now doesn't propagate event to actor, just verifies pointer events for actor.
-
- <!----------------------------------------------------------------------------------------------->
-
-## Unreleased
+- `CapturePointer.update()` method now doesn't propagate event to actor, just verifies pointer events for actor.
 
 ## Fixed
 - Added missing variable assignments to TileMapImpl constructor ([#957](https://github.com/excaliburjs/Excalibur/pull/957))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   -Property scope `Pointer.actorsUnderPointer` changed to private;
 
 ## Added
+- Allow timers to limit repeats to a finite number of times ([#957](https://github.com/excaliburjs/Excalibur/pull/974))
 
   -New `PointerEvent.stopPropagation()` method added. Works the same way as    (`https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation`)
     ([#912](https://github.com/excaliburjs/Excalibur/issues/912))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,33 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ## Breaking Changes
+ - Property scope `Pointer.actorsUnderPointer` changed to private;
 
 ## Added
+ - New `PointerEvent.stopPropagation()` method added. Works the same way as    (https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation)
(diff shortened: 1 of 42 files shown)


src/engine/Events.ts
```diff
@@ -93,8 +93,29 @@ export class GameEvent<T> {
    public target: T;
 }
 
+export abstract class BubblingEvent extends GameEvent<Actor> {
```
eonarheim: I think it would be okay, to move this implementation to GameEvent... that way all events can bubble without extra effort.

eonarheim:
@LoserAntbear Thanks for the PR!

src/engine/Input/Pointer.ts
```diff
@@ -321,106 +327,254 @@ export class Pointers extends Class {
...
+      this._verifyActorUnderPointers(actor, this._pointerDown);
+      this._verifyActorUnderPointers(actor, this._pointerUp);
+      this._verifyActorUnderPointers(actor, this._pointerMove);
+      this._verifyActorUnderPointers(actor, this._pointerCancel);
+      this._validateWheelEventPath(this._wheel, actor);
+   }
 
-      i = 0;
-      len = this._pointerDown.length;
+   private _verifyActorUnderPointers(actor: Actor, pointers: PointerEvent[]) {
```
eonarheim: `verify` implies that the method is not side-effecting.

Let's refactor this method and break it into 2 methods:
* `_generateActorsUnderPointers()` That builds a list of pointer/actor groupings
* `_propagate(listOfPointerActor)` That propagates events through a given list

src/engine/Input/Pointer.ts
```diff
@@ -321,106 +327,254 @@ export class Pointers extends Class {
...
+      this._propagatePointerEvent('pointermove', this._pointerMove);
+      this._propagatePointerEvent('pointercancel', this._pointerCancel);
+      this._propagateWheelEvent('pointerwheel', this._wheel);
+   }
+
+   /**
+    * Verifies Actor pointer events accordingly to actor provided
+    * @param actor  Actor to be verified
+    */
+   public verifyPointerEvents(actor: Actor) {
```
eonarheim: `verify` implies that the method is not side-effecting. `propagate` may be a better verb here

Is there a difference between `propagate` and `verifyPointerEvents`? See refactor proposal below

src/engine/Input/Pointer.ts
```diff
@@ -579,6 +733,7 @@ export class Pointer extends Class {
 
    private _isDown: boolean = false;
    private _wasDown: boolean = false;
+   private _actorsUnderPointer: IActorsUnderPointer = { length: 0 };
```
eonarheim: Why did this need to become private?

CHANGELOG.md
```diff
@@ -6,12 +6,33 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+    ([#912](https://github.com/excaliburjs/Excalibur/issues/912))
+ - New `Engine.input.pointers.verifyPointerEvents(actor)` method, which checks if event can be fired on provided Actor
+ - New abstract Class `BubblingEvent` which extends `GameEvent` and provides event with bubbling functionality
+ - New `Actor.getPath()` method, which retreives full array of current Actors` ancestors
+ - Added better functionality for working with actors under pointer:
+   - New method `addActorUnderPointer(actor)`
+   - New method `removeActorUnderPointer(actor)`
+   - New method `hasActorUnderPointer(actor)`
+   - New property `hasActorsUnderPointer`
+   * New intreface `IActorUnderPointer` which adds `length` property to old untyped object
```
eonarheim: Typo `interface`

CHANGELOG.md
```diff
@@ -6,12 +6,33 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ## Breaking Changes
+ - Property scope `Pointer.actorsUnderPointer` changed to private;
 
 ## Added
+ - New `PointerEvent.stopPropagation()` method added. Works the same way as    (https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation)
```
jedeen: Thanks for adding a changelog entry! The changelog file is a place for consumer-facing changes, so these entries can probably be reduced to API changes, and the relevant new/changed features that will be immediately helpful to a user trying to make a game.

In this case, a number of these new methods, etc. can be documented appropriately in the code comments so that consumers can utilize them correctly.

Typically, a completed GitHub issue usually results in a line or two in the changelog, like what you have here at the beginning of the "Added" section.

CHANGELOG.md
```diff
@@ -6,12 +6,33 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+   * New intreface `IActorUnderPointer` which adds `length` property to old untyped object
 
 ## Changed
+ - `CapturePointer.update()` method now doesn't propagate event to actor, just verifies pointer events for actor.
+ - `Scene.update()` method. Now it calls `engine.input.pointers.propagate()` method for starting propagation of events accordingly to paths provided. Event propagation became less greedy.
+ - Class `BubblingEvent` now extends `GameEvent`;
 
-## Deprecated
+ ## Fixed
+ - `jasmine.d.ts` type definitions are now for Jasmine 2.8.0. It equals to the version used in current development.
```
jedeen: Thanks for updating this! However, we don't usually include developer tooling updates in the changelog, though.

src/engine/Scene.ts
```diff
@@ -307,43 +307,32 @@ export class Scene extends Class implements ICanInitialize, ICanActivate, ICanDe
...
       // Cycle through triggers updating
-      for (i = 0, len = this.triggers.length; i < len; i++) {
-         this.triggers[i].update(engine, delta);
-      }
+      this.triggers.forEach((trigger) => trigger.update(engine, delta));
 
       this._collectActorStats(engine);
 
+      // propagates all events through their paths assigned
+      engine.input.pointers.propagate();
```
eonarheim: Is this going to cause a performance hit by running propagation every frame? Pointer events won't necessarily fire every frame.

Can we move propagation to be contained in the `CapturePointer.ts` trait where it only runs if an actor has opted in to pointer events?

src/engine/Actor.ts
```diff
@@ -5,8 +5,8 @@ import { Texture } from './Resources/Texture';
 import {
```
eonarheim: There are a lot of empty whitespace removal changes, can you undo these in files/areas that aren't related to the change?

A separate PR focused on correcting trailing whitespace would be great.

src/spec/SoundSpec.ts
```diff
@@ -7,7 +7,7 @@ const mockData = 'testdata';
 describe('Sound resource', () => {
```
eonarheim: Please remove whitespace changes.

src/spec/EngineSpec.ts
```diff
@@ -3,20 +3,20 @@
```
eonarheim: Please remove whitespace changes

LoserAntbear:
Wow! Didn't expect to get review so fast. 
Thanks a lot for all notices and recommendations. I'll rework this commit accordingly to your comments as fast as i can.
Really sorry to see i did a lot of stuff wrong, will try my best to give better PR's next time.

src/engine/Scene.ts
LoserAntbear: The thing is, that events are propagated by paths assigned to them. If Event's path is empty, nothing'll happen. I moved assignment of event's paths to CapturePointer Trait. So event'll be updated not every frame, but only when the happen.

src/engine/Events.ts
LoserAntbear: Now i am preparing updated PR, and i have extended functionality of that class. So, i am not moving Merging BubblingEvent to GameEvent to let you check all changes i made. And after revision, if you'll feel like merging this two guys together, just comment and i'll do it. :3

src/engine/Scene.ts
```diff
@@ -307,43 +307,32 @@ export class Scene extends Class implements ICanInitialize, ICanActivate, ICanDe
...
       this._preupdate(engine, delta);
-      var i: number, len: number;
-
-      
 
       // Remove timers in the cancel queue before updating them
-      for (i = 0, len = this._cancelQueue.length; i < len; i++) {
-         this.removeTimer(this._cancelQueue[i]);
-      }
+      this._cancelQueue.forEach((timer) => this.removeTimer(timer));
```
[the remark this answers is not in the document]
LoserAntbear: Shure thing, reverted this back. I am sorry for messing this up. :C

src/engine/Events.ts
```diff
@@ -16,7 +16,7 @@ export type prekill = 'prekill';
 export type postkill = 'postkill';
```
LoserAntbear: Also, as You can see, i moved all propagation functionality to event itself, after did some research about this in internet and books and discussing it with my friends. I find it more suitable to address all this responsibilities to event itself. And it  now has more adequate structure.

src/engine/Events.ts
eonarheim: I think this looks good, I like that the propagation logic is stored with the event. I'd still like to have this as a feature of all GameEvents, let's move this implementation merge to GameEvent.

src/spec/EngineSpec.ts
eonarheim: @LoserAntbear We only want changes relevant to feature at hand. Having whitespace changes in unrelated files makes it difficult for us to use git history to see what has changed when for a feature. It makes reviewing more difficult as well.

Git allows you to checkout files back to master `git checkout origin/master -- path/to/file` this could be used on files that don't have changes pertinent to event propagation to revert those whitespace changes, ie: `EngineSpec.ts`, `SoundSpec.ts`, etc.

Normally I'd totally agree with this whitespace cleanup, it just belongs in a separate PR, probably with some tooling (#884) to prevent extraneous whitespace from being committed again.

src/engine/Scene.ts
```diff
@@ -307,9 +307,8 @@ export class Scene extends Class implements ICanInitialize, ICanActivate, ICanDe
     */
    public update(engine: Engine, delta: number) {
       this._preupdate(engine, delta);
-      var i: number, len: number;
-
-      
+      let i: number;
```
eonarheim: While I agree with these changes in principle, lets keep this PR focused on the issue at hand and not modify things that aren't part of event propagation.

src/spec/EngineSpec.ts
LoserAntbear: I made new PR, which closes #884. Here it is: https://github.com/excaliburjs/Excalibur/pull/988
I also moved there `Jasmine 2.8.0` type definitions changes and all this whitespace removals and fixes in `EngineSpec.ts` and `SoundSpec.ts`.
But i have to wait for this one to be merged, 'cause i  need Jasmine 2.8.0 features for using in testes for this PR.

LoserAntbear:
So, next pack of changes has come.
Merged `BubblingEvent` with `GameEvent`, covered all with tests and reverted back to `origin/master` all irrelevant changes.

src/spec/EventSpec.ts
```diff
@@ -63,7 +63,7 @@ describe('An Event Dispatcher', () => {
       pubsub.on('someevent', () => {
```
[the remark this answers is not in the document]
LoserAntbear: Done!

src/engine/Scene.ts
```diff
@@ -309,16 +309,14 @@ export class Scene extends Class implements ICanInitialize, ICanActivate, ICanDe
...
-
       // Remove timers in the cancel queue before updating them
       for (i = 0, len = this._cancelQueue.length; i < len; i++) {
          this.removeTimer(this._cancelQueue[i]);
       }
       this._cancelQueue.length = 0;
 
       // Cycle through timers updating timers
-      for (var timer of this._timers) {
+      for (const timer of this._timers) {
```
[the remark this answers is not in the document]
LoserAntbear: I am sorry, i haven't noticed this one. :c fixed this in new commit. Also i've created an issue #989 for for tracking the upgrades to modern JS;

CHANGELOG.md
LoserAntbear: I have removed excessive entries from this Changelog, please, check  if it now looks fine.  :3

PR #983, opened 2018-04-27
[#982] Implement isCurrentScene convenience method

### LINKED ISSUE #982: Add convenience method to determine if a given Scene is the current scene
### Context
`Engine.currentScene` returns the current scene, but there is not a way from within `Scene` itself to determine if that scene is the current scene. You can achieve this by checking `engine.currentScene === scene`, but this requires that you take a dependency on `engine` and we should make this more accessible.

### Proposal
Add an `isCurrentScene` method to `Scene`. We can track this state by hooking into the `onActivate` and `onDeactivate` events. This would be useful in circumstances where an event is bound to the engine but we only want its callback to execute in a specific scene, like we did [here](https://github.com/excaliburjs/ludum-41/blob/1549e06a953ead1e0a56fb1d931ba5c64eaa4303/game/top/top.ts#L137).

FerociousQuasar:
Closes #982 

## Changes:

- added isCurrentScene() to scene objects
..* This is property is updated on activate(), deactivate() which is managed by engine.gotoScene.

## Note:

### DIFF
--- a/src/engine/Scene.ts
+++ b/src/engine/Scene.ts
@@ -156,7 +156,7 @@ export class Scene extends Class implements ICanInitialize, ICanActivate, ICanDe
 
    /**
     * Safe to override onPreUpdate lifecycle event handler. Synonymous with `.on('preupdate', (evt) =>{...})`
-    * 
+    *
     * `onPreUpdate` is called directly before a scene is updated.
     */
    public onPreUpdate(_engine: Engine, _delta: number): void {
@@ -165,7 +165,7 @@ export class Scene extends Class implements ICanInitialize, ICanActivate, ICanDe
 
    /**
     * Safe to override onPostUpdate lifecycle event handler. Synonymous with `.on('preupdate', (evt) =>{...})`
-    * 
+    *
     * `onPostUpdate` is called directly after a scene is updated.
     */
    public onPostUpdate(_engine: Engine, _delta: number): void {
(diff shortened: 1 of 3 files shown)


src/engine/Scene.ts
```diff
@@ -223,7 +224,7 @@ export class Scene extends Class implements ICanInitialize, ICanActivate, ICanDe
          this._initializeChildren();
          
          this._logger.debug('Scene.onInitialize', this, engine);
-
+
```
eonarheim: Let's remove this added whitespace

src/engine/Scene.ts
```diff
@@ -732,6 +734,12 @@ export class Scene extends Class implements ICanInitialize, ICanActivate, ICanDe
       this._sortedDrawingTree.add(actor);
    }
 
+   public isCurrentScene():boolean {
```
eonarheim: Fix linting error, space between `:` and `boolean`

alanag13:
That's a very valid point. Right now engine is only passed to scene's ctor optionally, so  i would be more in favor of the setter you mentioned rather than asking `engine.currentScene == this`, otherwise it would mean a hard dependency on engine.

eonarheim:
@alanag13 I don't think it's an issue, scenes that have yet to be associated with the engine cannot be the active scene.

eonarheim:
Also scenes cannot run without engine, so they already have a hard dependency.

src/engine/Scene.ts
```diff
@@ -734,12 +731,13 @@ export class Scene extends Class implements ICanInitialize, ICanActivate, ICanDe
       this._sortedDrawingTree.add(actor);
    }
 
-   public isCurrentScene():boolean {
-      return this._isCurrentScene;
+   public isCurrentScene(): boolean {
```
kamranayub: Thanks, this feels much better

PR #984, opened 2018-04-27
[#977] Fix pick set with duplicates

### LINKED ISSUE #977: ex.Random.pickSet with duplicates doesn't return expected results
### Steps to Reproduce

```typescript
let array = [1, 2, 3, 4, 5];
let random = new ex.Random(42);

let newArray = random.pickSet(array, 6, true);
console.log(newArray)
//[undefined, undefined, undefined, undefined, undefined, undefined, 2, 4, 5, 1, 4, 4]
```

### Expected Result

Pick set with duplicates should return a new array with the length that matches the number of picks requested.

### Actual Result

Pick set with duplicates produces an array prefixed with `undefined` 

### Current Workaround
Use random to generate an integer index and manually produce a new array.

FerociousQuasar:
Closes #977 

## Changes:

- Changed array from pushing new randomly chosen elements to setting the indices to these elements as in the other function

## Note:

### DIFF
--- a/src/engine/Math/Random.ts
+++ b/src/engine/Math/Random.ts
@@ -194,7 +194,7 @@ export class Random {
       }
       var result = new Array<T>(numPicks);
       for (var i = 0; i < numPicks; i++) {
-         result.push(this.pickOne(array));
+         result[i] = this.pickOne(array);
       }
       return result;
    }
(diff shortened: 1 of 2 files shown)


src/engine/Math/Random.ts
```diff
@@ -194,7 +194,7 @@ export class Random {
       }
       var result = new Array<T>(numPicks);
       for (var i = 0; i < numPicks; i++) {
-         result.push(this.pickOne(array));
+         result[i] = this.pickOne(array);
```
eonarheim: 👍

jedeen:
Moved to https://github.com/excaliburjs/Excalibur/pull/990

PR #985, opened 2018-04-27
[#981] Fix scene remove should kill actor

### LINKED ISSUE #981: ex.Scene.remove and ex.Actor.kill should behave the same
### Steps to Reproduce
ex.Scene.remove does not kill the actor, this causes many subtle issues.

### Expected Result
ex.Scene.remove should kill the actor.

### Actual Result
ex.Scene.remove does not kill the actor.

### Current Workaround
Use ex.Actor.kill instead of ex.Scene.remove

FerociousQuasar:
Closes #981 

## Changes:

- Actor sets _isKilled to true before calling scene remove
- Scene.Remove calls entity.kill() if it isn't already dead (i.e. it was triggered through a process other than actor.kill()
- Scene.Remove this._broadphase.untrack(entity.body) is redudant to this._removeChild(entity) which also calls this method

## Note:

### DIFF
--- a/src/engine/Actor.ts
+++ b/src/engine/Actor.ts
@@ -702,8 +702,8 @@ export class ActorImpl extends Class implements IActionable, IEvented, IPointerE
       if (this.scene) {
          this._prekill(this.scene);
          this.emit('kill', new KillEvent(this));
-         this.scene.remove(this);
          this._isKilled = true;
+         this.scene.remove(this);
          this._postkill(this.scene);
       } else {
          this.logger.warn('Cannot kill actor, it was never added to the Scene');
--- a/src/engine/Scene.ts
+++ b/src/engine/Scene.ts
@@ -566,9 +566,16 @@ export class Scene extends Class implements ICanInitialize, ICanActivate, ICanDe
          this.removeUIActor(entity);
          return;
       }
+      
       if (entity instanceof Actor) {
-         this._broadphase.untrack(entity.body);
-         this._removeChild(entity);
+
+         if(this._killQueue.indexOf(entity) == -1) {
+            this._removeChild(entity);
+         }
+
+         if (!entity.isKilled()) {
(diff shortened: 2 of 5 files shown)


src/engine/Scene.ts
```diff
@@ -566,9 +566,16 @@ export class Scene extends Class implements ICanInitialize, ICanActivate, ICanDe
          this.removeUIActor(entity);
          return;
       }
+      
       if (entity instanceof Actor) {
-         this._broadphase.untrack(entity.body);
-         this._removeChild(entity);
+
+         if(this._killQueue.indexOf(entity) == -1) {
```
kamranayub: This might be a bigger task but it'd be great to make this a O(1) call. We have actorId, we could set `_killQueue[actorId] = true`

src/engine/Actor.ts
```diff
@@ -702,8 +702,8 @@ export class ActorImpl extends Class implements IActionable, IEvented, IPointerE
       if (this.scene) {
          this._prekill(this.scene);
          this.emit('kill', new KillEvent(this));
-         this.scene.remove(this);
          this._isKilled = true;
+         this.scene.remove(this);
```
kamranayub: What's the impact of this?

src/engine/Scene.ts
eonarheim: I think this is too big for this issue, I'm going to recommend we add this type of change to our performance milestone as a separate issue. But O(1) kill queue lookups would be ideal.

src/engine/Actor.ts
FerociousQuasar: this needed to be moved in order for there not to be an infinite loop of
scene-> actor.kill()
actor->scene.remove(actor)
scene->actor.kill()

By moving this I can easily check if the actor has been killed before trying to kill it again in the scene.

PR #987, opened 2018-05-03
[chore] [docs] Fix breakout ball collision step

clofresh:
If you're following along with the quickstart guide, after the step
where you add the ball, it says the ball will bounce off of the paddle
but it doesn't until you've added the precollision handler later on
in the tutorial.

This change adds the precollision handler when adding the ball, then
updates it later for breaking the bricks.

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -128,6 +128,22 @@ collisions, which is sufficient for breakout.
    // "ex.CollisionType.Active - this means participate and let excalibur resolve the positions/velocities of actors after collision"
    // "ex.CollisionType.Fixed - this means participate, but this object is unmovable"
 
+    // On collision bounce the ball
+    ball.on('precollision', function (ev) {
+        // reverse course after any collision
+        // intersections are the direction body A has to move to not be clipping body B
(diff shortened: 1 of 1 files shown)


eonarheim:
@clofresh Good catch! Thanks!

Vyeche:
Hi, it seems its broken in version "0.31.0" ... The ball goes right through.

eonarheim:
@Vyeche Oh no! Thanks for the comment on this, investigating now

eonarheim:
@Vyeche I've pushed a fix

`other` is now a Collider type https://github.com/excaliburjs/sample-breakout/commit/71e311c2dd43648b6121a321d11d26315d9935be#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3d

image

eonarheim:
@Vyeche Looks like the tutorial docs were updated to be correct  https://github.com/excaliburjs/Excalibur/blob/main/site/docs/00-tutorial/00-breakout.mdx

Let me know if you spot anything else!

image

Vyeche:
Yep the code changed fixed the issue of it skipping the bounce off the bar.

PR #970, opened 2018-04-20
[#794] [#936] Update TypeScript & add webpack support

kamranayub:
Closes #936 
Closes #794 

## Changes:

- Update to TypeScript 2.8.1 & typedoc 0.11.1
- Add webpack v4 support for engine
- Outputs raw ES2015 modules
- Outputs production and minified bundles
- Simplify Grunt tasks
- Rename `npm run tests` to `npm test`
- Rename `npm run compile` to `npm run core`

PR #992, opened 2018-05-10
[#705] Allow setting a global anchor default for Actors

### LINKED ISSUE #705: Allow setting a global anchor default for Actors
### Context

Awhile back we decided it might be a good default to center Actors. I think this is still 50/50 that it is a good default--sometimes I overwrite it to be top left.

I think it might be beneficial to allow us to just decide globally how we want the default anchor to act--i.e. I can choose for my game to anchor all actors (0, 0) or (0.5, 0.5) by default.

### Proposal

A workaround of course is to create a new base actor for your game and set it that way. I propose also allowing a Engine option.

```ts
ex.Engine.defaults.actors.anchor.setTo(0, 0);
```

This could be extended for *x* number of properties and might be useful to set globals.

LoserAntbear:
Closes #705 

## Changes:

- Add `EngineDefaultOptions` Singleton to provide access for Engine's default options
- Add basic support of Global and Local scopes for `Actor.anchor`.

### DIFF
--- a/src/engine/Actor.ts
+++ b/src/engine/Actor.ts
@@ -57,6 +57,10 @@ export interface IActorArgs extends Partial<ActorImpl> {
    collisionType?: CollisionType;
 }
 
+export interface IActorDefaults {
+   anchor: Vector;
+}
+
 /**
  * @hidden
  */
@@ -64,6 +68,12 @@ export interface IActorArgs extends Partial<ActorImpl> {
 export class ActorImpl extends Class implements IActionable, IEvented, IPointerEvents, ICanInitialize, ICanUpdate, ICanDraw, ICanBeKilled {
    // #region Properties
 
+   /**
+    * Indicates the next id to be set
+    */
+   public static defaults: IActorDefaults = {
+      anchor: Vector.Half.clone()
+   };
    /**
     * Indicates the next id to be set
     */
@@ -480,9 +490,11 @@ export class ActorImpl extends Class implements IActionable, IEvented, IPointerE
       this.actionQueue = new ActionQueue(this);
       this.actions = new ActionContext(this);
 
+      // initialize default options
+      this._initDefaults();
+
       // Initialize default collision area to be box
       this.body.useBoxCollision();
-      this._initDefaults();
    }
 
    /**
@@ -519,11 +531,9 @@ export class ActorImpl extends Class implements IActionable, IEvented, IPointerE
          child._initialize(engine);
       }
    }
-   // Initialize default options, that depends on Environment defaults
+
    private _initDefaults() {
-      console.log(Engine);
-      // this.anchor = Engine.defaults.actorAnchor;
-      this.anchor = Vector.Half;
+      this.anchor = Actor.defaults.anchor.clone();
    }
 
    // #region Events
--- a/src/engine/Actor.ts
+++ b/src/engine/Actor.ts
@@ -185,8 +185,8 @@ export class ActorImpl extends Class implements IActionable, IEvented, IPointerE
    public oldAcc: Vector = Vector.Zero.clone();
 
    /**
-    * Gets the acceleration vector of the actor in pixels/second/second. An acceleration pointing down such as (0, 100) may be 
-    * useful to simulate a gravitational effect.  
+    * Gets the acceleration vector of the actor in pixels/second/second. An acceleration pointing down such as (0, 100) may be
+    * useful to simulate a gravitational effect.
     */
    public get acc(): Vector {
       return this.body.acc;
@@ -325,7 +325,7 @@ export class ActorImpl extends Class implements IActionable, IEvented, IPointerE
     */
    public oldScale: Vector = Vector.One.clone();
 
-   /** 
+   /**
     * The x scalar velocity of the actor in scale/second
     */
    public sx: number = 0; //scale/sec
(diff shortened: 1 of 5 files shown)


LoserAntbear:
I have began some work on this issue, so i decided to show you some ideas i have about it. Have i got you right about this issue at all. :D

LoserAntbear:
Hello again. Sadly, but i am experiencing some kind of annoying problem right now.
Karma build fails constantly with the same error "`Uncaught TypeError: Object prototype may only be an Object or null: undefined`" when i am tying to call `Engine` from any class and in any form. I've checked to the master branch and tried to call it from current master. It also crashes. Maybe you are aware of this thing? I'll appreciate any help!

eonarheim:
Odd, I'll have some time tonight. I'll take a look

eonarheim:
@LoserAntbear I dug into this, the problem appears to be a circular dependency between Actor <-> Engine. Removing the `console.log(Engine)` line allows the build to pass. The problem is that when almond (our umd bundler) tries to find the Actor prototye and it fails in UIActor. I believe this may be do to the dependency graph changing when we reference `Engine` from `Actor` like this.

Theoretically our #970 webpack update should address problems with circular dependencies, but I haven't been able to get this branch working with it yet. It's throwing a similar error on this branch

image

image

src/engine/Actor.ts
```diff
@@ -521,6 +519,12 @@ export class ActorImpl extends Class implements IActionable, IEvented, IPointerE
          child._initialize(engine);
       }
    }
+   // Initialize default options, that depends on Environment defaults
+   private _initDefaults() {
+      console.log(Engine);
+      // this.anchor = Engine.defaults.actorAnchor;
+      this.anchor = Vector.Half;
```
eonarheim: When referencing the `Vector.Constants` we should also `.clone()`

`this.anchor = Vector.Half.clone()`

eonarheim: Since these aren't value types in javascript we can run into trouble with subtle bugs sharing vector references, it's just good practice to `.clone()` vectors when you can.

src/engine/Engine.ts
```diff
@@ -153,16 +153,44 @@ export interface IEngineOptions {
    backgroundColor?: Color;
 }
 
+export class EnvDefaultOptions {
```
eonarheim: * Since these are `Actor` defaults, let's move these to `Actor.ts` keep the code near where what it configures

* Additionally let's not build a singleton for this, how about we hang a static bag off `Actor` instead and clone values in the `_initDefaults` method?

* Not referencing `Engine`'s prototype should also fix our current circular dependency problem

Something like
```typescript

export interface IActorDefaults {
   anchor: ex.Vector;
}

export class ActorImpl extends ...{
   public static defaults: IActorDefaults = {
      anchor: Vector.Half.clone()
   };
   ...

   _initDefaults() {
       this.anchor = Actor.defaults.anchor.clone();
   } 
}
```

src/engine/Actor.ts
```diff
@@ -480,11 +480,9 @@ export class ActorImpl extends Class implements IActionable, IEvented, IPointerE
       this.actionQueue = new ActionQueue(this);
       this.actions = new ActionContext(this);
 
-      // default anchor is in the middle
-      this.anchor = new Vector(.5, .5);
-
       // Initialize default collision area to be box
       this.body.useBoxCollision();
+      this._initDefaults();
```
eonarheim: You will also need to move the `_initDefaults()` at least for `anchor` above `this.body.useBoxCollision()`;

`useBoxCollision()` needs the anchor to perform calculations for physics geometry

```typescript
this._initDefaults();
// Initialize default collision area to be box
this.body.useBoxCollision();
```

src/engine/Actor.ts
LoserAntbear: So, there is the point, where all the fuss has stared!
Idea, which i wanted to introduce was a little rethink of @kamranayub original one, presented in #705 . And this part is just some code to represent the issue. 
Mine idea is commented one: `// this.anchor = Engine.defaults.actorAnchor;`

The point is to have an `Engine.defaults` property, as @kamranayub . And restrict this property for changing for everyone, except `Engine`. 
By default every created `Actor` has their `Actor.anchor` set to `Engine.defaults.actorAnchor` and scope for this property is `Global`. It means, that we can only change `Engine.default.actorAnchor` to value we need, and every `Actor` with `Global` scope will change this value. And if we do not need some `Actor` to change his value, we could just simply change his scope to `Local`. Like, 'Actor.setLocalAnchor(Vector)` or smth like that. Isn't it meaningless?

eonarheim: @LoserAntbear I see where you're coming from, but I don't think it's necessary to block changing the default except when on engine. 

If a game developer wants all anchors to be changed `Actor.defaults.anchor` could be set at start of the game, then all actors created after that would get the new default. I think it would be unexpected to set the `Actor.defaults.anchor` then have every actor in the game simultaneously update, I feel like that would violate the "principle of least surprise".

If an Actor needs a different anchor, the developers can have their individual anchors updated already with the `.anchor` property. Currently it is possible to do this:

```typescript
// game start
let engine = new ex.Engine({width: 600, height: 400});
Actor.defaults.anchor = new ex.Vector(.75, .75);
game.start(() => {

...

// defaults anchor to new ex.Vector(.75, .75) at constructor time
let actor = new ex.Actor({x: 10, y: 20, width: 50, height: 50, color: ex.Color.Red.clone()})

// set to something else if needed
actor.anchor = new ex.Vector(0, 0);

game.add(actor)

});

```

LoserAntbear:
@eonarheim Actually, fixed issue with `undefined` `Actor` passed to `UIActor` constructor, by fixing export order in main `Index.ts`. I am assuming it is happens, 'cause class statements are not hoisting. 

So, by now this realization is working file.
I just have to rewrite it accordingly to your requests. :)

eonarheim:
@LoserAntbear Thanks for the help!

PR #1016, opened 2018-07-14
[#979] prevent removal of actors that are killed and readded during the same frame

### LINKED ISSUE #979: Removed and re-added actors in the same frame are erroneously killed
### Steps to Reproduce

```
let game = new ex.Engine({
  width: 600,
  height: 400
});

let actor = new ex.Actor(100, 100, 50, 50, ex.Color.Red);
actor.on('postupdate', () => {
  actor.kill();
  game.add(actor);
})
game.add(actor);
game.start();
```

### Expected Result
Actor should be unkilled and re-added to the scene, in the above example the actor is no longer drawn

### Actual Result
Actor remains killed

### Environment
excalibur@0.16.0

### Current Workaround
None

alanag13:

Closes #979

## Changes:

- Add check to ensure Actor is still killed before removing it from drawing tree.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fix issue where Actors would not be properly added to a scene if they were removed from that scene during the same frame ([#979](https://github.com/excaliburjs/Excalibur/issues/979))
+
 <!--------------------------------- DO NOT EDIT BELOW THIS LINE --------------------------------->
 
 ## [0.17.0] - 2018-06-04
--- a/src/engine/Scene.ts
+++ b/src/engine/Scene.ts
@@ -390,10 +390,13 @@ export class Scene extends Class implements ICanInitialize, ICanActivate, ICanDe
     // Remove actors from scene graph after being killed
     var actorIndex: number;
     for (let killed of killQueue) {
-      actorIndex = collection.indexOf(killed);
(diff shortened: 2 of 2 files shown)


alanag13:
I was not sure of a good way to add a test for this since everything affected by it is private, but let me know if you think of something.

kamranayub:
If it's possible the private method(s) don't need to be tested directly,
the test should behave like a consumer would and the behavior should be
tested. We wouldn't want to make private methods not private just for code
coverage. If the methods are used, then they should result in a change of
state that is testable. Line coverage doesn't matter that much.

alanag13:
@kamranayub what I meant was that the public methods that would actually be used to test this case (actor.kill/scene.remove and scene.add) only affect private member variables (the draw tree) for the purposes of this test (the actors are correctly added to/removed from the scenes even in the current "broken" case).

I suppose we could do something like implement an onupdate event and run a test to make sure that it still fires in this case, but that seemed to me like it would not be testing the *direct* result we were looking for, which is just that it hasn't been removed from the draw tree.

kamranayub:
Maybe that means we should introduce a new public method, Scene.hasActor?

eonarheim:
@alanag13 I agree with @kamranayub perhaps a new method is in order here. When something is difficult to test, it can be a sign of an api deficiency or a bad design (probably my fault 💀).

alanag13:
added method and test to check if actor is in a scene's draw tree.

PR #1033, opened 2018-09-01
[#975] Fix collision pair generation to ensure ordering doesnt matter

### LINKED ISSUE #975: PreventCollision and Active Collision Types are generating collision pairs sometimes
### Steps to Reproduce

Add the `PreventCollision` type actor to the scene first then a `Active` collision type actor.

Possible fix https://github.com/excaliburjs/Excalibur/commit/2e0f2e7d5ed9ae8df3e9a08dc50c321f632bcdde

This appears to be the offending line of code in `Pair.ts`:

```typescript
if (actorB.collisionType === CollisionType.PreventCollision || actorB.isKilled()) { return false; }
```

### Expected Result
`PreventCollision` and `Active` collision types should never collide. 

### Actual Result
Collision pairs are being generated and postcollision events are firing erroneously.

### Environment
Excalibur @ v0.16.0

### Current Workaround
Currently this does not cause too much trouble, but it does generate pairs unecessarily no physics resoultion happens. The workaround is to inspect postcollision events before handling them.

eonarheim:

==================

Closes #975

## Changes:

- Checks both sides of the collision
- Handles edge case with kill
- Adds tests

### DIFF
--- a/src/spec/CollisionSpec.ts
+++ b/src/spec/CollisionSpec.ts
@@ -50,6 +50,91 @@ describe('A Collision', () => {
     expect(actor2Collision).toBe(1);
   });
 
+  it('should order of actors collision should not matter when an Active and Active Collision', () => {
+    let collisionTree = new ex.DynamicTreeCollisionBroadphase();
+
+    actor1.collisionType = ex.CollisionType.Active;
+    actor2.collisionType = ex.CollisionType.Active;
+    collisionTree.track(actor1.body);
+    collisionTree.track(actor2.body);
+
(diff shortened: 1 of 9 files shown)


src/spec/CollisionSpec.ts
```diff
@@ -50,6 +50,91 @@ describe('A Collision', () => {
     expect(actor2Collision).toBe(1);
   });
 
+  it('should order of actors collision should not matter when an Active and Active Collision', () => {
```
kamranayub: I'd probably remove `should` from beginning of these tests as it reads weird

PR #1042, opened 2018-10-01
[docs] Update the trigger documentation example

### LINKED ISSUE #1041: Trigger Docs outdated
The current docs at https://excaliburjs.com/docs/api/edge/classes/_trigger_.trigger.html is a bit outdated and this example won't work.

The actor doesn't have any collision so it will never trigger.

`actor.collisionType = ex.CollisionType.Active;`

The moveto is not on actor but on actions

`actor.actions.moveTo(100,200,100);`

I'm also quite new in this project but it seems these API docs aren't on Github or am I just that blind? If so, could you point me in the correct direction and I'll change the docs myself.

Kind regards,

T00mm:

==================

Closes #1041

### DIFF
--- a/src/engine/Docs/Triggers.md
+++ b/src/engine/Docs/Triggers.md
@@ -1,25 +1,38 @@
 ## Creating a trigger
 
 ```js
-var game = new ex.Game();
+// Start the engine
+var game = new ex.Engine({ width: 800, height: 600, displayMode: ex.DisplayMode.FullScreen });
+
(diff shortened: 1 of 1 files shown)


kamranayub:
@eonarheim We should think about pulling these examples into our official jsfiddle/codesandbox accounts to link to/embed. These are the types of examples we'll want to ensure we compile and test every build 👍

eonarheim:
@T00mm @kamranayub Thanks, I'll have time tonight to take a look!

PR #1060, opened 2018-10-22
[#1060] Implement SpriteSheet spacing

### LINKED ISSUE #1058: Spritesheet spacing
Current tilesheet with spacing

After adding spacing to the spritesheet class

### Context
Some tilesheets contain spacing. ExcaliburJS can't work with it yet.

### Proposal
To ensure that Tiled (and excalibur-tiled) can work properly with excalibur we need the extra spacing feature, I've put an issue on excalibur-tiled for this problem as wel (#25 but it seems the github problems today have let it vanished again, maybe it will pop up in a couple of hours).

I'm not very well adopted yet in this whole workflow but the things I've changed are in the Spritesheet class.

1. Added in the constructor an optional spacing param with default zero
2. Changed the line in the constructor

`this.sprites[j + i * this.columns] = new Sprite(this.image, j * this.spWidth, i * this.spHeight, this.spWidth, this.spHeight);`

to

`this.sprites[j + i * this.columns] = new Sprite(this.image, j * this.spWidth + this.spacing * j, i * this.spHeight  + this.spacing * i, this.spWidth, this.spHeight);`

I'm terribly sorry I didn't do it the correct way but I'm a bit out of time for today :-/ I'll try to do it the next couple of days.

T00mm:

==================

Closes #1058 

## Changes:

- Spacing has been added to the spritesheet constructor
- For some reason I was forced to add || 0 for the default value else it would give undefined


eonarheim:
@T00mm I think the reason is that the `spacing?: number` is effectively a type union between `number | undefined` so some sort of type guard was required to satisfy the `spacing: number` only type

PR #1065, opened 2018-11-11
[#1066] Add EventTypes enum to Events

### LINKED ISSUE #1066: Create strongly typed EventTypes
### Context
Having a strongly typed source for our event types would be an asset to game development. As folks build games in TypeScript they do not want to rely on magic strings.

### Proposal
Create an enum called EventTypes following pascal casing for the individual types. Using the same string literal value in the enum still allows the types to flow for built in events while allowing others to still define there own "string" based events

image

That way the type signature in the final implementation would look like this when exported out of the Events namespace in Excalibur:

`ex.Events.EventTypes.PostUpdate` 

image

atanasbozhkov:

==================

Closes #1066

## Changes:

- Added EvenTypes enum to Events.ts

First time PR Questions:

- Not sure how to test this - if there is a place where I can drop in the enum values to make sure it's encluded do let me know.

- Additionally - I noticed that there two events mapping to the same string value - `EventType.Enter` as well as `EventType.EnterTrigger` - what's the reason for this?
- Do I need to open an issue for this?
- Do I update the changelog?

Tested locally - with this simple scenario and seems good. 
image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Added strongly typed EvenTypes enum to Events.ts to avoid magic strings [#1066](https://github.com/excaliburjs/Excalibur/issues/1058)
+
 ### Changed
 
 - Added parameter on SpriteSheet constructor so you can define how many pixel spacing is between sprites ([#1058] https://github.com/excaliburjs/Excalibur/issues/1058)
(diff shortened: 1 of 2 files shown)


eonarheim:
@atanasbozhkov to answer your questions:

* Please do create a changelog entry under the `Added` section

* I've already created an issue to document our conversation on the forum and update the PR accordingly #1066 
* I don't think we need to add any tests for this particular enhancement, since we are doing string literal typing in the `on/off/once` methods I'm not too concerned.
* I think this is a bug, they should not be ambiguous! `EventType.Enter`'s literal value appears to be used in `Pointer.ts#createPointerEventByName` and is meant to handle a specific pointer event, and `EventType.EnterTrigger` is meant to fire on actor entry into a trigger area! It's possible `EventType.Enter` needs to be renamed or moved to match it's usage in `Pointer.ts`

eonarheim:
@atanasbozhkov Also thanks for contributing!

PR #1081, opened 2018-12-29
[#1078] Fix issue early offscreen culling related to zoom

### LINKED ISSUE #1078: Offscreen culling kicking in too early
Actors are being culled while still visible when the camera zooms in, and when zooming out they pop back in after their position is within the visible area.

See the pen here: https://codepen.io/mjkayson/pen/YdzEpZ (it gradually zooms in and out)

Ideally (for me at least) culling would happen a fair while after the actor is outside the field of view - also would be nice to have a flag disable culling on a given actor.

eonarheim:

==================

Closes #1078

## Changes:

- Add new `viewport` property to camera to return a world space bounding box of the current visible area
- Add visual tests

## TODO:

* [ ] Add unit test w/ image diff
* [ ] Refactor cullbox on sprites to use bounding boxes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fixed issue of early offscreen culling related to zooming in and out [#1078](https://github.com/excaliburjs/Excalibur/issues/1078)
+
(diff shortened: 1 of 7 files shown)


eonarheim:
@excaliburjs/core-contributors This is ready to go, added a visual test as well

image

PR #1109, opened 2019-02-27
[#1107] Test event handler execution order

### LINKED ISSUE #1107: [chore] Write test to confirm event handlers execute in order of subscription
### Context
This was an existing TODO see #1105 

Currently we dont have a test that shows handlers execute in the same order as subscription.

### Proposal
Write a test in `EventSpec.ts` that registers `n` number of handlers in a certain order and tests that those handlers execute in the same order.

agrgic16:

==================

Closes #1107 

## Changes:

- wrote new test in EventSpec.ts asserting that event handlers are called in the order subscribed

### DIFF
--- a/src/spec/EventSpec.ts
+++ b/src/spec/EventSpec.ts
@@ -44,9 +44,27 @@ describe('An Event Dispatcher', () => {
     expect(eventFired).toBeTruthy();
   });
 
-  //it('has event handlers called in the right order', () => {
-  //   //TODO
-  //});
+  it('has event handlers called in the right order', () => {
+    const subscriptions: number[] = [];
+    let eventHistory: number[];
+
+    for (let i = 0; i < 5; i++) {
+      subscriptions.push(i);
+    }
+
+    eventHistory = [];
+    subscriptions.forEach(i => pubsub.on('event', () => eventHistory.push(i)));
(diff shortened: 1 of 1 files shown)


src/spec/EventSpec.ts
```diff
@@ -44,9 +44,22 @@ describe('An Event Dispatcher', () => {
...
+  it('has event handlers called in the right order', () => {
+    const subscriptions = [];
+    const eventHistory = [];
+
+    for (let i = 0; i < 5; i++) {
+      subscriptions.push(i);
+    }
+
+    subscriptions
+      .sort(() => Math.random() - 0.5)
```
alanag13: Tests should always do exactly the same thing, so I think we should steer away from Math.random here.
To show that they're always called in the correct order, we could have the test change the order in a predictable way a couple of times and have `expect` assertions show that the order is correct after each one of those changes.

agrgic16: I can definitely do that. I think testing with controlled random numbers is a viable option conceptually, if done correctly and in the right circumstance, but I'd like to conform to how you guys are currently doing things so I will gladly make the change.

agrgic16: @alanag13 changes pushed to remove the random ordering.

Issue #994, opened 2018-05-22
Replace Excalibur promise implementation with browser implementation


eonarheim:
### Context
Currently we have a basic functioning custom promise implementation, but this prevents new browser features like `async`/`await` from working natively. Additionally we will get a bunch of promise features for free by switching to native promises. It may also come with a bonus of closing these #533 #365 #341

### Proposal

- [ ] Add the tsconfig native promise polyfill [option](https://www.typescriptlang.org/docs/handbook/compiler-options.html) `lib: ES2018.Promise`
- [ ] For backwards compatibility for other games, change `ex.Promise` implementation to wrap the native one.
- [ ] Replace existing methods in excalibur that use `ex.Promise` and use the native Promise returns (this will be a breaking change, unless use the `ex.Promise` wrapper, there can be no easy transition unless we rename methods)

LoserAntbear:
Looks like we also need to add `lib: [es2015, dom]`. Because `es2018.Promise` adds `final` method to Promise interface, defined  in `es2015.promise` `PromiseConstructor`, if i understand this right. I already made this changes and also replaced all uses of `ex.Promise` with native `es6-promises` and ready to push here. But i am a bit confused with providing backwards compatibility to `ex.Promise`. Now i just added as prop `ex.Promise._nativePromise` which resolves on `ex.Promise.resolve` and rejects on `ex.Promise.reject`, but i feel it is not the way it is meant to be. :)

LoserAntbear:
Hello again! I am sorry, bear with me. :D Could you look into this, please?

eonarheim:
@LoserAntbear Adding `lib: [es2015, dom]` makes sense to me.

Please send a PR, as long as the new implementation of `ex.Promise` backed by the `_nativePromise` behaves in a similar way to the old one (the tests should be a good indicator) that's what we want.

kamranayub:
You could be even more specific I think and do es2015.promise or something
like that.

PR #1110, opened 2019-03-01
[#1108] Check that spritesheet is valid given source texture dimensions

### LINKED ISSUE #1108: [chore] Check that spritesheet is valid given source texture dimensions
### Context
Was a TODO #1105 

It is currently possible to create a spritesheet that is invalid given it's source texture dimensions
See https://github.com/excaliburjs/Excalibur/pull/1106/files#diff-c3d2f62c7d54f809d5dc0e29d6e45f7dL62

### Proposal
* [ ] Validate spritesheet against the texture's loaded dimensions.
* [ ] Write a test that checks this validation works

edualb:

==================

Closes #1108 

## Changes:

- It was updated the package.json and package-lock.json with current versions
- It was implemented two exceptions in SpriteSheet in which check if the image Texture size is correct.
- It was implemented two tests for check if the exceptions is correct.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
   "devDependencies": {
     "@types/jasmine": "2.8.7",
     "@types/node": "8.10.18",
+    "acorn": "6.1.0",
     "copy-webpack-plugin": "4.5.1",
     "cross-env": "5.1.6",
     "css-loader": "1.0.0",
@@ -56,7 +57,7 @@
     "grunt-karma": "3.0.0",
     "grunt-shell": "2.1.0",
     "grunt-string-replace": "1.3.1",
-    "grunt-ts": "6.0.0-beta.19",
+    "grunt-ts": "6.0.0-beta.22",
     "grunt-tslint": "3.3.0",
     "grunt-webpack": "3.1.2",
     "husky": "0.14.3",
@@ -64,12 +65,11 @@
     "istanbul-instrumenter-loader": "^3.0.1",
     "jasmine": "2.99.0",
     "jasmine-core": "2.99.1",
-    "karma": "3.1.0",
+    "karma": "4.0.0",
     "karma-chrome-launcher": "2.2.0",
     "karma-coverage": "1.1.2",
-    "karma-coverage-istanbul-reporter": "^2.0.4",
+    "karma-coverage-istanbul-reporter": "2.0.5",
     "karma-jasmine": "1.1.0",
-    "karma-typescript": "3.0.13",
     "karma-webpack": "3.0.5",
     "prettier": "1.13.4",
     "pretty-quick": "1.6.0",
@@ -81,10 +81,10 @@
     "travis-ci": "2.2.0",
     "ts-loader": "4.3.0",
     "tslint": "3.15.1",
-    "typedoc": "0.11.1",
-    "typescript": "2.8.3",
-    "url-loader": "1.1.1",
-    "webpack": "4.27.1",
-    "webpack-cli": "3.1.2"
+    "typedoc": "0.14.2",
+    "typescript": "3.3.3333",
+    "url-loader": "1.1.2",
+    "webpack": "4.29.4",
+    "webpack-cli": "3.2.3"
   }
 }
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,14 +1,17 @@
-<!-- 
-Hi, and thanks for contributing to Excalibur! 
+<!--
+Hi, and thanks for contributing to Excalibur!
 Before you go any further, please read our contributing guide: https://github.com/excaliburjs/Excalibur/blob/master/.github/CONTRIBUTING.md
-especially the "Submitting Changes" section: 
+especially the "Submitting Changes" section:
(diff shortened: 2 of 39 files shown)


package.json
```diff
@@ -56,20 +56,20 @@
...
     "jasmine": "2.99.0",
     "jasmine-core": "2.99.1",
-    "karma": "3.1.0",
+    "karma": "^4.0.1",
     "karma-chrome-launcher": "2.2.0",
     "karma-coverage": "1.1.2",
     "karma-coverage-istanbul-reporter": "^2.0.4",
     "karma-jasmine": "1.1.0",
-    "karma-typescript": "3.0.13",
+    "karma-typescript": "^4.0.0",
```
eonarheim: Please revert `package.json` & `package-lock.json`

edualb:
I will check this errors...

edualb:
@eonarheim So, there is something very wrong with the tests. When I run `npm run all`, all tests are successfully. When I run `npm run test` one more time, I'm received differents results,some instants with succesfully and other instants with errors. Do you have any idea to solved it? I'll commit my changes to you check.

edualb:
Travis-CI got one of errors.
This is that showing in my computer:

image

If you want I can make a video to show you.

eonarheim:
@edualb Odd, locally I get a different number of errors. I wonder if we are running into a platform specific bug. TravisCI is running linux, my machine is windows, what's yours?

image

edualb:
@eonarheim windows 10. I receive this error too. If you run many times in a some moment all tests are successfully. Really, is odd.

eonarheim:
@edualb Poking around I think the issue is that the backing texture may or may not be loaded at constructor time. Adding some defense and checking if the naturalWidth exists and is non-zero seems to fix the flaky tests. This can only happen if the texture is not yet loaded.

image

```typescript
if (this.image instanceof Texture && 
        (this.image && this.image.image)) {
      if (this.image.image.naturalWidth && 
          this.spWidth * this.columns > this.image.image.naturalWidth) {
        throw new Error(`SpriteSheet specified is wider, 
        ${this.columns} cols x ${this.spWidth} pixels > ${this.image.image.naturalWidth} 
        pixels than image width`);
      }
      if (this.image.image.naturalHeight && 
          this.spHeight * this.rows > this.image.image.naturalHeight) {
        throw new Error(`SpriteSheet specified is taller, 
        ${this.rows} rows x ${this.spHeight} pixels > ${this.image.image.naturalHeight} 
        pixels than image height`);
      }
    }
```
Since we do not force people to only use loaded textures, instead of checking at constructor time, we should probably wait until the texture is loaded something like 

```typescript
if(this.image) {
  this.image.loaded.then(() => {
    // check image height/width and throw
 });
}
```

edualb:
@eonarheim 
So, I think put this code inside of constructor will work:
```
    if (this.image instanceof Texture) {
        this.image.load().then((imageAux) => {
            if (this.spWidth * this.columns > imageAux.naturalWidth) {
                throw new Error(`SpriteSheet specified is wider, 
                    ${this.columns} cols x ${this.spWidth} pixels > ${this.image.image.naturalWidth} 
                    pixels than image width`);
            }
            if (this.spHeight * this.rows > imageAux.naturalHeight) {
                throw new Error(`SpriteSheet specified is taller,
                    ${this.rows} rows x ${this.spHeight} pixels > ${this.image.image.naturalHeight} 
                    pixels than image height`);
            }
        });
    }
```
With this code when the image is loaded we will can check the throws errors.

But when I compile i receive this error:

`Uncaught Error: Cannot resolve a promise that is not in a pending state! thrown`

Do you have any idea what is wrong with my code?

eonarheim:
@edualb Switch `.load()` to `.loaded`, I think the issue is because `.load()` is used by the excalibur load machinery

edualb:
@eonarheim I changed `.load` to `.loaded`. Yesterday I was trying run tests with this modifications but i think exists some "delay" between resolve promise and finish constructor. Because when I put a `console.log` before conditions to throw error, in the tests shows this `console.log` but dont shows the throw error. Can you help me solved this test? I will commit my modifications and you can check it.

edualb:
@eonarheim Do you know any way to  put a catch block in .loaded Promise? Because when I try put catch block after then block I receive this error: Property 'catch' does not exist on type 'Promise<any>'.

Something like this:
```
        this.image.loaded.then((image: HTMLImageElement) => {
            try {
                if (this.spWidth * this.columns > image.naturalWidth) {
                    throw new RangeError(`SpriteSheet specified is wider, 
                        ${this.columns} cols x ${this.spWidth} pixels > ${image.naturalWidth} 
                        pixels than image width`);
                }
                if (this.spHeight * this.rows > image.naturalHeight) {
                    throw new RangeError(`SpriteSheet specified is taller,
                        ${this.rows} rows x ${this.spHeight} pixels > ${image.naturalHeight} 
                        pixels than image height`);
                }
            } catch (e) {
                return e;
            }
        }).catch(e => {
            return e;
        });
```

eonarheim:
@edualb Unfortunately Excalibur's promise implementation is incomplete (it predates the ES promise specification) it is on our roadmap to remove and replace with native browser promises #994 (@alanag13 has expressed interest in fixing this)

Instead of re-throwing, for now lets use the ex.Logger to log and error

edualb:
@eonarheim  this way is good to merge? 

More later i'll check the #994 =)

Issue #1102, opened 2019-02-27
Update Excalibur to follow styleguide


eonarheim:
### Context
Some of the codestyle in Excalibur is not consistent. The team has drafted a new styleguide to help guide development

### Proposal
In several PRs update Excalibur with deprecation warnings where possible

* [x] PR to add STYLEGUIDE.md + update PR template for Excalibur
* [x] PR to update interface naming 
* [x] PR to update method naming 1 commit per file
* [x] PR to update to modern ES features `let` and `const` 
* [x] PR to update TSLint/ESLint rules to match styleguide

alanag13:
we should consider adding ESLint rules for these items where possible.

eonarheim:
@alanag13 good call, I've updated the issue to reflect that

DavidLi119:
Hi, are there still any files to update to modern ES features `let` and `const`? I was thinking of doing ActionSpec.ts. Is that fine?

eonarheim:
@DavidLi119 That would be awesome, open a PR!

DavidLi119:
@eonarheim Quick question before I get started, what module do I install, when I open VSCode, every single method has red underline and it says it cannot find module `'../../build/dist/excalibur'`

djcsdy:
Hi. Heads up: I am working on replacing all remaining instances of `var` with `let`/`const`.

eonarheim:
@djcsdy Thanks! Please split with one commit per updated file as best you can

djcsdy:
Oops. Sorry. I didn’t mean to spam this issue with every commit. I was trying to follow the style I observed in previous commit messages, and I forgot that GitHub would helpfully cross-reference the commits.

eonarheim:
Complete!

PR #1118, opened 2019-04-13
[#1102] Update Excalibur to follow style guidelines

DavidLi119:
Changed `var` to `const` and `let` where applicable. Left `var` in `for` loops alone. 
Works on Issue #1102
Files changed: src/spec/ActionSpec.ts
Notes: VSCode has problems finding module `../../build/dist/excalibur` requesting solution please?
           Unsure whether to change `var` on lines 597 and 618 to `let` or `const`, currently changed to `let` if it is fine as it is currently, please let me know thanks.

==================


eonarheim:
@DavidLi119 Thanks for the contribution! 

Responses:

* The for in this form `for (var i = 1; i < 10; i++) {` can be safely changed to use `let`.

* The tests in VSCode rely on a successful build of the main library, that's why vscode is upset about `../../build/dist/excalibur`.
  - run `npm install` then `npm run all` to produce a build and run the full test suite. VSCode should find the built files after that.

DavidLi119:
@eonarheim Changes made. should be good to go no?

eonarheim:
Thanks for the contribution @DavidLi119!

PR #1131, opened 2019-04-20
[#1091][#862] Collision Group Re-implementation

### LINKED ISSUE #862: CollisionGroups do not work at all
### Steps to Reproduce
The `.onCollidesWith` handler does not fire when active collision types collide with groups added with `.addCollisionGroup` that match the filter in the handler.

Here is a codepen reproducing the issue https://codepen.io/eonarheim/pen/LjjeEm 
Left click to move the read block right, right click to move the blue block left. 

We also have a second "Group" type that is confusing on which should be used.

### Expected Result
What should happen is that a message should be written if the blocks collide since there is a collision group handler on actor1.

### Actual Result
As far as I can tell this does not happen, possibly a regression during the collision system over-haul.

### LINKED ISSUE #1091: Collision Groups Re-implementation
### Context
Depends on #1119 

The previous implementation of collision groups was meant to describe groups that an Actor could collide with at once. This implementation was not efficient and was fairly brittle, and doesn't really match with what others are doing with collisions in industry.

In other engines, groups indicate like members that do not collide with each other. For example:
* Players have collision group A
* Enemies have collision group B
* Blocks have collision group C

Players don't collide with each other, but enemies and blocks. Likewise, enemies don't collide with each other but collide with players and blocks.

A common way to implement this quickly is with bit masking. For example

PlayerGroup  = `0b001`
PlayerGroupMask =`0b110`

EnemyGroup = `0b010`
EnemyGroupMask =`0b101`

BlockGroup =   `0b100`
BlockGroupMask =`0b011`

Should Players collide? No because the bitwise mask evaluates to 0
`(player1.group & player2.mask) === 0`
`(0b001 & 0b110) === 0`

Should Players and Enemies collide? Yes because the bitwise mask is non-zero
`(player1.group & enemy1.mask) === 1`
`(0b001 & 0b101) === 1`

Should Players and Blocks collide? Yes because the bitwise mask is non-zero
`(player1.group & blocks1.mask) === 1`
`(0b001 & 0b011) === 1`

### Proposal

* [ ] Deprecate existing collision groups
* [ ] Use possible collision group management implementation 
* [ ] Update documentation around usage

[stackblitz example](https://stackblitz.com/edit/collision-groups?file=index.ts)

```typescript

class CollisionGroup {
  private _name: string;
  private _category: number;
  private _mask: number;
  constructor(name: string, category: number, mask: number) {
    this._name = name;
    this._category = category;
    this._mask = mask;
  }
  public get name() {
    return this._name;
  }
  public get category() {
    return this._category;
  }
  public get mask() {
    return this._mask;
  }

  public shouldCollide(other: CollisionGroup): boolean {
    return (this.category & other.mask) != 0 && (other.category & this.mask) != 0;
  }
}

class CollisionGroupManager {
  private static MAX_GROUPS = 32; // using bitmasking the maximum number of groups is 32, because that is the heighest 32bit integer that JS can represent.
  private static _currentGroup = 1;
  private static _currentBit = 0b1 | 0;
  private static _groups: {[name: string]: CollisionGroup} = {};
  public static create(name: string, mask?: number) {
    if (this._currentGroup > this.MAX_GROUPS) throw new Error('Cannot have more than 32 collision groups')
    const group = new CollisionGroup(name, this._currentBit, mask != undefined ? mask : ~this._currentBit);
    this._currentBit = (this._currentBit << 1) | 0;
    this._currentGroup++;
    this._groups[name] = group;
    return group;
  }

  public static get groups(): CollisionGroup[] {
    return Object.keys(this._groups).map(g => this._groups[g]);
  }

  public static groupByName(name: string) {
    return this._groups[name];
  }
}

let playerGroup = CollisionGroupManager.create("player");
let enemyGroup = CollisionGroupManager.create("enemy");
let blockGroup = CollisionGroupManager.create("block");
console.log('Player and player should collide', playerGroup.shouldCollide(playerGroup)); // false
console.log('Player and enemey should collide', playerGroup.shouldCollide(enemyGroup)); // true
console.log('Player and block should collide', playerGroup.shouldCollide(blockGroup)); // true
```

eonarheim:

==================

Closes #1091
Closes #862

## Changes:

- Implement collision group proposal in #1091
- Wire into collider infrastructure
- Add unit tests 
- Update visual tests

blocks


src/engine/Collision/CollisionGroup.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Players have collision group A
+ * Enemies have collision group B
+ * Blocks have collision group C
+ *
+ * Players don't collide with each other, but enemies and blocks. Likewise, enemies don't collide with each other but collide
+ * with players and blocks.
+ */
+export class CollisionGroup {
+  public static All = new CollisionGroup('Collide with all groups', -1, -1);
```
eonarheim: Add a comment to explain the all group

src/engine/Collision/CollisionGroup.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * CollisionGroups indicate like members that do not collide with each other. Use [[CollisionGroupManager]] to create [[CollisionGroups]]
+ *
```
eonarheim: Add a picture explaining this visually

src/engine/Collision/CollisionGroupManager.ts
```diff
@@ -0,0 +1,42 @@
+import { CollisionGroup } from './CollisionGroup';
+
+export class CollisionGroupManager {
```
eonarheim: Needs a comment, also this should be the recommended way to create collision groups

eonarheim:
I think we are ready @excaliburjs/core-contributors

sandbox/tests/group/group.ts
```diff
@@ -40,25 +42,12 @@ for (var i = 0; i < numActors; i++) {
...
   });
-  actor.on('precollision', function() {
-    //console.log('inner collision');
-  });
 
   actor.vel.x = ex.Util.randomInRange(minVel, maxVel);
   actor.vel.y = ex.Util.randomInRange(minVel, maxVel);
-
-  blockGroup.add(actor);
+  engine.add(actor);
```
kamranayub: Should we not keep the visual indication of collision by flipping colors?

src/engine/Collision/CollisionGroupManager.ts
```diff
@@ -0,0 +1,52 @@
+import { CollisionGroup } from './CollisionGroup';
+
+/**
+ * Static class for managing collision groups in excalibur, there is a maximum of 32 collision groups possible in excalibur
+ */
+export class CollisionGroupManager {
+  // using bitmasking the maximum number of groups is 32, because that is the heighest 32bit integer that JS can present.
```
kamranayub: JS supports `2^53 - 1` [max safe integer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER), while not quite 64-bit, it's higher than 32-bit, does this limit still apply to this code?

sandbox/tests/group/group.ts
eonarheim: Would you like the blocks to change color, maybe to a teal-ish color to match the block?

src/engine/Collision/CollisionGroupManager.ts
eonarheim: That's a good question, let me do some digging and find the correct answer here. We could potentially support additional groups.

eonarheim: @kamranayub After experimenting in the tests this does not work unfortunately.

So the answer is tricky here, while JS does indeed support `2^53-1` integers, it does not support bitwise operators to that length before they are mangled by IEEE representation. JS bitwise operators only work on 32 bit signed integers unfortunately.

image

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Signed_32-bit_integers
https://stackoverflow.com/questions/2983206/bitwise-and-in-javascript-with-a-64-bit-integer

eonarheim: In fact you can see the signed numbers 32 integer overflow

image

sandbox/tests/group/group.ts
eonarheim: collisiongroups

src/engine/Collision/CollisionGroupManager.ts
kamranayub: Well now we know! Thanks 👍

src/engine/Collision/CollisionGroupManager.ts
```diff
@@ -0,0 +1,52 @@
...
+  private static _groups: { [name: string]: CollisionGroup } = {};
+
+  /**
+   * Create a new named collision group up to a max of 32.
+   * @param name Name for the collision group
+   * @param mask Optionally provide your own 32-bit mask, if none is provide the manager will generate one
+   */
+  public static create(name: string, mask?: number) {
+    if (this._currentGroup > this._MAX_GROUPS) {
+      throw new Error('Cannot have more than 32 collision groups');
```
kamranayub:
```suggestion
      throw new Error(`Cannot have more than ${this._MAX_GROUPS} collision groups`);
```

src/engine/Collision/CollisionGroupManager.ts
```diff
@@ -0,0 +1,52 @@
...
+  public static groupByName(name: string) {
+    return this._groups[name];
+  }
+
+  /**
+   * Resets the managers internal group management state
+   */
+  public static reset() {
+    this._groups = {};
+    this._currentBit = 0b1;
```
kamranayub: Above it's `0b1 | 0` does it need to match here? Also should this be a constant?

src/engine/Collision/CollisionGroupManager.ts
```diff
@@ -0,0 +1,52 @@
...
+    this._currentGroup++;
+    this._groups[name] = group;
+    return group;
+  }
+
+  /**
+   * Get all collision groups currently tracked by excalibur
+   */
+  public static get groups(): CollisionGroup[] {
+    return Object.keys(this._groups).map((g) => this._groups[g]);
```
kamranayub:
```suggestion
    return Object.values(this._groups);
```

src/engine/Collision/CollisionGroupManager.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Static class for managing collision groups in excalibur, there is a maximum of 32 collision groups possible in excalibur
+ */
+export class CollisionGroupManager {
+  // using bitmasking the maximum number of groups is 32, because that is the heighest 32bit integer that JS can present.
+  private static _MAX_GROUPS = 32;
+  private static _currentGroup = 1;
+  private static _currentBit = 0b1 | 0;
+  private static _groups: { [name: string]: CollisionGroup } = {};
```
kamranayub: I propose using native `Map` as it's waaaaaaaay faster: https://jsperf.com/map-values-vs-object-values/1

src/engine/Collision/CollisionGroupManager.ts
kamranayub: See above, using `Map.values()` will be 150X faster.

src/engine/Collision/CollisionGroupManager.ts
kamranayub: In fact when/if you change it, you should measure what the impact is 👍

src/engine/Collision/CollisionGroupManager.ts
eonarheim: We can make a constant for the starting bit 👍 I'll make the change later today

src/engine/Collision/CollisionGroupManager.ts
eonarheim: I'll make the change later today

sandbox/tests/group/group.ts
```diff
@@ -40,25 +42,22 @@ for (var i = 0; i < numActors; i++) {
...
     }
   });
-  actor.on('precollision', function() {
-    //console.log('inner collision');
+
+  actor.on('postcollision', function(e: ex.PostCollisionEvent) {
+    if (e.actor.currentDrawing instanceof ex.Sprite && e.other === player) {
+      e.actor.currentDrawing.colorize(ex.Color.Cyan);
+    } else {
+      debugger;
```
eonarheim: Remove `debugger`

PR #1133, opened 2019-04-20
[#1132] Guarantee polyfill application

### LINKED ISSUE #1132: undefined is not a constructor (evaluating 'new window.AudioContext') on Safari
the error is `TypeError: undefined is not a constructor (evaluating 'new window.AudioContext')`

### Steps to Reproduce

I reproduced using a fresh clone of the example-ts-webpack, and tested with both 0.21.0 and 0.22.0 versions of excalibur. Start the running the project and press the play button and error will appear and the game will not start.

More specifically the error comes from calling `ex.AudioContextFactory.create();` the `create` function does not properly set the audio context if the case is webkit.

 

### Expected Result

Expect for the audio context to be defined or gracefully fail

### Actual Result

Error is thrown and execution halts. 

### Environment

- browsers and versions: Safari 12.1 (14607.1.40.1.4) 
- operating system: OSX 
- Excalibur versions: 0.21.0, 0.22.0 

### Current Workaround

Its possible to redefine the static `create` method to return the webkitAudioContext

`ex.AudioContextFactory.create = () => new (<any>window).webkitAudioContext();`

which would allow at least for the game to load.

### Possible Solution
possible fix might be something like:

```
public static create(): AudioContext {
    if (!this._INSTANCE) {
      if ((<any>window).AudioContext) {
        this._INSTANCE = new (<any>window).AudioContext();
      } else if (<any>window).webkitAudioContext) {
         this._INSTANCE = new (<any>window).webkitAudioContext();
      }
    }
    return this._INSTANCE;
  }
```

eonarheim:

==================

Closes #1132 

The polyfill import was a side-effect based import, meaning webpack could not reasonably know that it needed to preserve a certain order. 

Fix confirmed in browserstack latest safari

image

@excaliburjs/core-contributors Should we consider a hotfix for this?

## Changes:

- Fix p1 bug where minified excalibur polyfills were not being applied correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
-Fixed wrong `Camera` and `Loader` scaling on HiDPI screens when option `suppressHiDPIScaling` is set.
+- Fixed wrong `Camera` and `Loader` scaling on HiDPI screens when option `suppressHiDPIScaling` is set. ([#1120](https://github.com/excaliburjs/Excalibur/issues/1120))
+- Fixed polyfill application by exporting a `polyfill()` function that can be called. ([#1132](https://github.com/excaliburjs/Excalibur/issues/1132))
 
 <!--- Bug fixes here --->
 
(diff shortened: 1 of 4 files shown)


eonarheim:
@kamranayub Let me try this tonight https://webpack.js.org/guides/shimming/#loading-polyfills

eonarheim:
@kamranayub The polyfill entrypoint doesn't seem to work as expected with umd output. I'm going to go with our current fix for now.

PR #1141, opened 2019-05-01
[#1102] Prefer const/let to var

djcsdy:
I am working towards replacing all uses of `var` with `const` or `let`, as discussed in #1102.

This is mostly just a case of running `tslint --fix` with the `no-var-keyword` and `prefer-const` rules enabled. I am reviewing the changes manually before I commit each file.

~~This PR is a work in progress, I will push my changes to the corresponding branch as I go.~~

==================


eonarheim:
We'll start reviewing the changes as soon as we can 👍 thanks for all this work!

@djcsdy Let me know when we can pull the work-in-progress label off.

djcsdy:
@eonarheim I think you can take the wip label off now :-)

I have updated every source file except for `sandbox`.

`sandbox` requires changes to the build scripts before it can be updated to use const/let. The problem is that all the source files in sandbox are compiled as scripts (not modules), so any variables declared at top level become global variables. Since all the source files are compiled in a single pass, TypeScript considers multiple variables with the same name declared at the top level with `let` or `const` to be duplicate declarations, and this causes the build to fail. So I have left these using `var` for now.

I have also updated the documentation so that code samples now use `const` or `let`, but only where those code samples are marked as TypeScript. I have left the ones marked as JavaScript as they are, because I am not sure of the intention there. It seemed to me that perhaps pure JavaScript examples should be limited to ES5 syntax to avoid confusion. Let me know what you think.

In the last commit I’ve turned on the `no-var-keyword` and `prefer-const` rules in tslint in order to enforce this style in the future.

eonarheim:
@djcsdy Thanks so much for working hard on this! We'll be review as soon as we can. I expect we'll make it through this week.

Understood on the sandbox, I've run into the same issue in the past before.

src/engine/Docs/Actions.md
```diff
@@ -42,7 +42,7 @@ itself and repeating that forever.
 ```ts
 public Ship extends ex.Actor {
   public onInitialize() {
-    var path = [
+    const path = [
```
eonarheim: Excellent thanks for updating the docs too 👍

eonarheim:
During team standup this was approved

eonarheim:
@djcsdy Thanks again for all the hard work! This was a ton work, really appreciate it.

PR #1148, opened 2019-05-06
[#1147] Make event handlers strict null/function check compatible

### LINKED ISSUE #1147: Excalibur event handlers are not contravariantly or strict null check safe
Related to #1143 

### Steps to Reproduce
1. Enabled strict null checks and stict function checks in tsconfig.json
2. Define a resonable handler

### Expected Result
There should be no error when defining a handler like so 
image

### Actual Result
Compilation error

### Environment
- operating system: N/A
- Excalibur versions: master

### Current Workaround
1. Disable strict null checks and strict function checks
2. Add a more specific handler signature and a type guard
image

eonarheim:

==================

Closes #1147 

## Changes:

- The handler signature implies the event object may be null, this is untrue and a source of bugs and unessary checks

image


eonarheim:
@excaliburjs/core-contributors I don't believe this is a breaking change, narrowing null out of the type will still allow any existing type guards to work.

PR #1149, opened 2019-05-18
[#1084] Fix color lighten

### LINKED ISSUE #1084: ex.Color.Black.lighten() does not work
### Steps to Reproduce
See codepen
https://codepen.io/eonarheim/pen/BvORaG?editors=0010#0

```
let newColor = ex.Color.Black.lighten();
expect(newColor.r).not.toEqual(ex.Color.Black.r);
expect(newColor.g).not.toEqual(ex.Color.Black.g);
expect(newColor.b).not.toEqual(ex.Color.Black.b);
```
Looks like this issue is right here, it is using multiplication. When the black it is `0` so multiplication will always yeild a `0`

### Expected Result
The new color should be lighter than black

### Actual Result
New color is still black

### Environment

- browsers and versions: N/A
- operating system: N/A
- Excalibur versions: master

### Current Workaround
None

ZuzuZain:
[#1084] Fix Color.lighten()

Closes #1084

## Changes:

- Update lighten with a different formula
- Added test for for Color.lighten()
- Update ..\src\spec\images\SpriteSpec\icon-lightened.png


eonarheim:
@ZuzuZain Thanks for the contribution!

PR #1157, opened 2019-05-28
[#1071] Actor.within checks from geometry surface

### LINKED ISSUE #1071: Within only checks from center of object not from bounding box
### Steps to Reproduce
Create a long block in a rectangle shape and then create 2 other blocks less then 100 px from it. The first block should be less then 100 PX from the center of the rectangle and the second block should be 100 px from the edge of the rectangle. When within is called on the other 2 blocks the first will be within 100px but the other will not since its great then 100px from the center.

Code example: https://codepen.io/anon/pen/mQvgXx?editors=0010

Code shows 4 objects. First floor object and then 3 square blocks all nearby with checks to see if they are within 200 px. If they are they turn to red otherwise green.

### Expected Result
All blocks are within 200 px from the bounding box or an edge of the floor so all should be red/within 200px and within should return true

### Actual Result
Within returns false for the block that is more then 200px from the center of the floor

eonarheim:

==================

Closes #1071

## Changes:

- Change `within` to use object geometry instead of the center
- Added a `getClosestLineBetween` method to `CollisionShape`'s for return the closest line
- Added a visual test
- Added unit tests

closest


eonarheim:
@excaliburjs/core-contributors Ready for review

src/engine/Algebra.ts
```diff
@@ -413,6 +436,22 @@ export class Line {
     return distance;
   }
 
+  /**
+   * Find the perpendicular line from the line to a point
+   * @param point
+   */
+  public findVectorToPoint(point: Vector): Vector {
+    // https://en.wikipedia.org/wiki/Distance_from_a_point_to_a_line
```
jedeen: Would it make sense to have these comments as part of the method comments, like the comment for ClosestLine() ?

src/spec/CollisionShapeSpec.ts
```diff
@@ -116,6 +116,15 @@ describe('Collision Shape', () => {
       expect(pointTooFar).toBe(null, 'The circle should be too far away for a hit');
     });
 
+    it('can be raycast against only positive toi', () => {
```
jedeen: It might be more clear to avoid acronyms, or at least have their expansion next to them.

Issue #1172, opened 2019-06-16
Animation.flipHorizontal can only be set to true


gargrave:
I believe this is the cause of the issue I am describing below: https://github.com/excaliburjs/Excalibur/blob/master/src/engine/Drawing/Animation.ts#L276

Animation simply passes `flipHorizontal` onto its sprite, but it's doing a truthy-only check, so a `false` value will never be passed on, and thus never get "unset". I believe we would more or less want to pass on any value here, as it seems like it is basically acting as a pass-thru for the exact same property (i.e. I don't think the conditional should be necessary).

FWIW, I cloned and tested on my end, and confirmed that simply removing the conditional makes this behave as expected. I can submit a PR if you would like! 👍 I just didn't know if there were some specific reason you were doing a conditional there.

(Side question, but is there any reason that you guys haven't defined a "main" in `package.json`? I had to add `"main": "build/dist/excalibur.js"` to get Yarn linking working correctly locally, and it would probably be useful to add in there if there isn't a strong reason against it.)

### Steps to Reproduce
1. Set up an Animation (I followed these steps: https://excaliburjs.com/docs/api/v0.23.0/classes/_drawing_animation_.animation.html) 
2. Call `[anim].flipHorizontal = true`
3. Sprite flips to have an x-scale of `-1` (good so far)
4. Call `[anim].flipHorizontal = false`
5. BUG: Sprite does not "unflip"

### Expected Result
Calling `[anim].flipHorizontal = false` should "unflip" the Animation/Sprite, but it does not.

### Actual Result
Animation/Sprite stays "flipped"

### Environment
- browsers and versions: All current browsers
- operating system: MacOS 10.14.4
- Excalibur versions: 0.23

### Current Workaround
If you access the Animation's current Sprite directly, you can toggle this state, like `this.currentDrawing.sprites[0].flipHorizontal = !this.currentDrawing.sprites[0].flipHorizontal`

But of course that is a pretty ugly workaround, and wouldn't really be sustainable for Animations with many frames.

eonarheim:
@gargrave Good find, please do open a PR. This is definitely a bug in flipping.

@gargrave We have a separate [dist repo](https://github.com/excaliburjs/excalibur-dist/blob/master/package.json) from our our main source, it's probable the main source package.json has languished (npm link/yarn link isn't a usual part of our workflow for local integration testing, but it probably should be). We definitely should fix that to enable this scenario.

gargrave:
[...]

PR #1174, opened 2019-06-16
#1172 Fix animation flipping behavior

gargrave:

==================

Closes https://github.com/excaliburjs/Excalibur/issues/1172

## Changes:

- Updates Animation to remove the conditional around passing `flipHorizontal/Vertical` to its current Sprite instance, as these should _always_ be passed to ensure their "flipped" state is accurate.
- Added tests to ensure that the flag can be flipped in _both_ directions now, as previously it could only be set to `true`

### DIFF
--- a/src/spec/AnimationSpec.ts
+++ b/src/spec/AnimationSpec.ts
@@ -2,8 +2,9 @@ import * as ex from '../../build/dist/excalibur';
 import { TestUtils } from './util/TestUtils';
 
 describe('An animation', () => {
-  let animation;
+  let animation: ex.Animation;
   let engine: ex.Engine;
+
   beforeEach(() => {
     animation = new ex.Animation(null, null, 0);
     engine = TestUtils.engine({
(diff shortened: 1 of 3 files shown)


src/spec/AnimationSpec.ts
```diff
@@ -48,4 +50,28 @@ describe('An animation', () => {
     expect(animation.speed).toBe(20);
     expect(animation.sprites.length).toBe(0);
   });
+
+  it('should always pass "flipped" state to the current Sprite', () => {
+    const mockSprite = {
+      anchor: new ex.Vector(1, 1),
+      draw: () => void 0,
```
gargrave: Even though these are effectively throwaway values, `anchor` and `draw` were required to be defined in some way in order to prevent the tests from crashing when "drawing" a Sprite.

eonarheim:
@gargrave Thanks for contributing!

PR #1187, opened 2019-06-25
[chore] Clean up changelog

jedeen:


jedeen:
The changelog just seemed to be leaning more consistently that way already. I don't think we currently have any guidance on how to format changelog entries (other than previous changelog entries).

PR #1248, opened 2019-09-05
[chore] Tests: Migrate visual tests to Storybook

kamranayub:

==================

Closes #

## Changes:

- Add Storybook support and configure for TypeScript
- Add support for new `EngineOptions.canvasElement` option to directly pass `HTMLCanvasElement`
- Migrate a visual test as an example

## TODO:

- [x] Use new Storybook 5.2 beta
- [x] Finalize conventions to write stories

## Summary

This PR is the beginning of a wholesale migration effort for existing visual tests to Storybook-based stories. The idea is that we use the current visual tests to make sure things work as expected but they are not well-documented nor truly designed for end-user consumption.

I'd like to move to Storybook to make our visual tests more like live documentation for us and end-users. For example, using Storybook we can:

- Organize visual examples
- Show code examples
- Add "knobs" for tweaking live in-browser
- Add DocPages support
- Double as test aids during releases
- Get hot module reload (HMR) during development 🔥 

This PR will introduce Storybook but each set of visual tests will be a follow-up PR until none remain. At the same time, I'll probably work on theming and having the docs repo deploy Storybook to the site (latest stable + edge-only?).

## Examples

fade example

Basic story with Knobs configured and Notes

### Docs panel

image


src/engine/Actions/Actions.stories.ts
```diff
@@ -0,0 +1,31 @@
+import { storiesOf } from '@storybook/html';
+import { withKnobs, number } from '@storybook/addon-knobs';
+import { Actor, Texture, Loader } from '../';
+import { withEngine } from '../story-utils';
+
+import heartTexture from '../__assets__/heart.png';
+
+storiesOf('Actions', module)
```
kamranayub: This story syntax will get simpler when I install the Storybook beta package. It will change to

```ts
export default {
  title: "Actions",
  decorators: [withKnobs]
}

export const fade = withEngine(game => { ... });
fade.story = {
  parameters: {
    notes: '...'
  }
}
```

src/engine/Actions/Actions.stories.ts
```diff
@@ -0,0 +1,31 @@
+import { storiesOf } from '@storybook/html';
+import { withKnobs, number } from '@storybook/addon-knobs';
+import { Actor, Texture, Loader } from '../';
+import { withEngine } from '../story-utils';
+
+import heartTexture from '../__assets__/heart.png';
```
kamranayub: ❤️

src/engine/story-utils.ts
```diff
@@ -0,0 +1,16 @@
+import { Engine, DisplayMode } from '.';
+
+/**
+ * Helper to generate Storybook game engine instance
+ * @param storyFn The storybook fn to pass the engine to
+ */
+export const withEngine = (storyFn: (game: Engine) => void) => {
+  return () => {
```
kamranayub: This will get invoked every time story re-renders or HMR reloads modules. It did expose some issues with Ex where we e.g. don't clean up the Play button in the DOM.

eonarheim: Uff, yah. We probably should open an issue to clean up the Play button after the game has started.

kamranayub:
And ideally reuse DOM element if already added.

kamranayub:
Something is up with my `package-lock.json` so I'll need to resolve that.

kamranayub:
One issue right now. When running in Storybook `iframe`, key events aren't captured even when clicking on the canvas.

image

Workaround is to open in a new tab using the button:

image

eonarheim:
@kamranayub That's odd, could it be the way we are listening to events in excalibur? Embedding the engine in an iframe seems like a use case we should support. Should we open a new issue?

PR #1249, opened 2019-09-07
[docs] Fix Loader call example in Texture documentation

butlersrepos:
I believe on `latest` this Loader call wants an array.

- [ ] :see_no_evil: code conforms to the [style guide](https://github.com/excaliburjs/Excalibur/blob/master/STYLEGUIDE.md)

==================

Closes #

## Changes:

- change 1
- change 2

### DIFF
--- a/src/engine/Docs/Textures.md
+++ b/src/engine/Docs/Textures.md
@@ -9,7 +9,7 @@ is loaded, you can generate a [[Sprite]] with it.
 
 ```js
 var txPlayer = new ex.Texture('/assets/tx/player.png');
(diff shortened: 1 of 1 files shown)


butlersrepos:
Is there any interest in further updating these docs with style points like 
- changing `var`s to `const`s 
?

eonarheim:
@butlersrepos Good catch 👍 you are correct that is the new expectation for Loader. 

We would be 100% interested in switching `var`'s to `const`'s or `let`'s where appropriate. Please break the changes into a PR per page if possible.

Issue #396, opened 2015-01-23
Add random timer range as a feature of timers 


eonarheim:
### Context
There is a use case for creating a timer that has a variable duration. For example, if you wanted to randomly space events during a game.

### Proposal
Add the ability to specify a range (min and max) of time and allow the timer to randomly choose a duration within that range.

yammik:
Hi! I'd like to work on this if it's still up for grabs 🙏

eonarheim:
All yours! 👍

DProBoy:
Is this still up for grabs ?

eonarheim:
@DProBoy All yours, I've added the in-progress label

DProBoy:
@eonarheim Can i (or Should I?) add new sandbox tests for the Random Timer ? I've committed the changes to my fork of Excalibur and i'm about to submit a pull request for #875 and #396 . (hehe ... still haven't gotten used to Git, as I was a Perforce user before ...)

kamranayub:
It may be good to add some visual tests to the sandbox/tests/timers
directory (which doesn't exist yet but feel free to follow an existing
example).

c3ho:
Hi, is this issue still open?

eonarheim:
@c3ho I haven't seen any work on it, all yours!

Excalibur has a built in seeded `ex.Random` implementation that we should use to do the random calculations. 

It may be worth while to rethink the constructor and the Timer interface as well.

Perhaps something along these lines? If possible we should allow the previous constructor to be used as well avoid a breaking change

```typescript
export interface TimerOptions {
    // milliseconds before the timer fires
   interval: number;
   // if number of repeats null, undefined means it fires once
   // if number of repeats <= 0 or Infinity means repeat forever
   repeats?: number;
   // maximum number of milliseconds to randomly add
   randomRange?: number; 
   // timer callback
   fcn?: () => void;
}

type Callback = () => void;

export class Timer {

    // Constructor overload
    constructor(options: TimerOptions);
    constructor(fcn: Callback, interval: number, repeats?: boolean, numberOfRepeats?: number);
    constructor(optionsOrFcn: TimerOptions | Callback, interval?: number, repeats?: boolean, numberOfRepeats?: number) {
       //... todo implement
    }
}
```

// Example new usage
```typescript
// repeat once in 200 ms
const timerOnce = new ex.Timer({
   fcn: () => { console.log('Called Once'); },
   interval: 200
});

// repeat 3 times every 200 ms
[...]

PR #1273, opened 2019-10-05
[#839] Change Timer implementation to support multiple handlers

### LINKED ISSUE #839: Add a ex.Timer.on/off method to easily add multiple callbacks
### Context
Add `ex.Timer.on` and `ex.Timer.off` methods to timer, right now it is hard to add callbacks to timer since it is a constructor argument. Also I want to make the constructor argument optional if we add these methods.

### Proposal
Something like this
```typescript
// forever 200 ms timer
var timer = new ex.Timer(200, true);
timer.on(() => {
   spawnNewEnemy();
})
```

```typescript
var timer = new ex.Timer(200, true);
var spawnHandler = spawnNewEnemy;
timer.on(spawnHandler);
timer.off(); // remove all handlers
timer.off(spawnHandler); // remove specific handler
```

chrispanag:

==================

Closes #839

## Changes:

- Change the constructor parameters of Timer to allow easier optional options. (added an options object and TimerOptions interface)
- Add method .on and .off to add a callback to the timer
- Change `new Timer()` everywhere to support new format

(Pending: Add tests and changelog entry)


src/engine/Timer.ts
```diff
@@ -1,5 +1,11 @@
 import { Scene } from './Scene';
 
+export interface TimerOptions {
```
eonarheim: I like where you are going with this, we have come up with similar ideas for refactoring the constructor 👍 

https://github.com/excaliburjs/Excalibur/issues/396#issuecomment-538583725

src/engine/Timer.ts
```diff
@@ -17,16 +23,20 @@ export class Timer {
...
   /**
-   * @param fcn        The callback to be fired after the interval is complete.
    * @param interval   Interval length
+   * @param options    Options - repeats, numberOfRepeats, fcn
    * @param repeats    Indicates whether this call back should be fired only once, or repeat after every interval as completed.
    * @param numberOfRepeats Specifies a maximum number of times that this timer will execute.
+   * @param fcn        The callback to be fired after the interval is complete.
    */
-  constructor(fcn: () => void, interval: number, repeats?: boolean, numberOfRepeats?: number) {
+  constructor(interval: number, options: TimerOptions = {}) {
```
eonarheim: Let's go ahead and make `interval` part of `TimerOptions` as well. We can add a constructor overload to preserve backwards compatibility

https://github.com/excaliburjs/Excalibur/issues/396#issuecomment-538583725

chrispanag: Totally agree with that. Proceeding with the change :)

src/engine/Timer.ts
```diff
@@ -37,6 +47,32 @@ export class Timer {
...
+      this.on(fcn);
+    }
+  }
+
+  /**
+   * Adds a new callback to be fired after the interval is complete
+   * @param fcn The callback to be added to the callback list, to be fired after the interval is complete.
+   */
+  public on(fcn: () => void) {
+    const key = fcn.toString();
```
eonarheim: Please use an array of functions, and not a dictionary. `this._callbacks.push(fcn)`. Functions have a unique reference and can be compared that when when `off(...)` is called. 

```typescript
private _callbacks: (() => void)[];
...
public off(fcn: () => void) {
  const index = this._callbacks.indexOf(fcn);
  if (index > -1) {
    this._callbacks.splice(index, 1);
  }
}	  
```

I'm slightly concerned about using the function code as the key, it would mean that functions that are syntactically the same but not, contextually the same could overwrite each other. This is probably unlikely but it would be a tricky bug. Example of the bug:

https://codepen.io/eonarheim/pen/MWWWvGb?editors=0011

eonarheim:
@chrispanag Thanks again this is looking good so far 👍

src/engine/Timer.ts
chrispanag: I was thinking of that also, nice tip about the reference! Thank you!

chrispanag:
@eonarheim If you want check the tests I just added. Do you think there is a need for any more elaborate testing of this addition?

eonarheim:
@chrispanag This looks really good, @excaliburjs/core-contributors any other feedback?

eonarheim:
This just needs a changelog entry 👍

eonarheim:
@chrispanag I have the approval from the team offline, we'll merge with a changelog entry

eonarheim:
@chrispanag Thanks for your contribution!

PR #1285, opened 2019-10-10
[#1214] Switch to core-js polyfills

### LINKED ISSUE #1214: Switch Excalibur polyfils to core-js
### Context
In the past we've used hand written polyfills, or polyfills sourced from MDN.

The [core-js polyfills](https://github.com/zloirock/core-js) are better maintained, patched, and up to date. This will also give us a mechanism to easily add polyfills as the browser landscape changes and evolves.

### Proposal
* Remove existing polyfill file
* Compile required core-js polyfills into the Excalibur dist

eonarheim:

==================

Closes #1214

## Changes:

- Include core-js for array & function
- Increases the size of our bundle


eonarheim:
One issue right now is that bundling core-js polyfills into the main artifact is counting against us in code coverage, not sure at the moment how to fix this. It may involve changes how we calculate code coverage?

eonarheim:
@kamranayub Any ideas? The problem is the core-js source is included in the final UMD bundle which is what code coverage runs against. We could change how code coverage works, or find a way to decorate parts of the UMD output :(

kamranayub:
Hmm ideally code coverage should only be done on source files 😫

eonarheim:
@kamranayub If this looks good to your eyes it's probably ready 🚀

PR #1310, opened 2019-10-28
Don't require GH_TOKEN to be set on Travis

### LINKED ISSUE #1309: Travis builds require GH_TOKEN to be set for no good reason
### Steps to Reproduce
 1. On GitHub, fork Excalibur to your own account or to an organization other than excaliburjs.
 2. Commit some minor change to trigger a Travis CI build.

### Expected Result
Travis CI should successfully build Excalibur.

### Actual Result
Build scripts complain that `GH_TOKEN` is not set and exit, causing the build to fail. This is annoying because it makes it difficult for third-party contributors to ensure their changes are correct before submitting them to the project.

image

The problem is that [version.js explicitly checks that `GH_TOKEN` is set](https://github.com/excaliburjs/Excalibur/blob/c09f07e79b634f307beab8884c36744f409df7d1/version.js#L46-L48) and throws an error if it is not. ~~However, this check is artificial because a GitHub token is not actually needed. The only place it is used is to [fetch the latest tag version of Excalibur from GitHub](https://github.com/excaliburjs/Excalibur/blob/c09f07e79b634f307beab8884c36744f409df7d1/version.js#L50-L56), and this API call does not require authorization.~~ Besides which, the version number is probably only relevant when building on the master branch of the main Excalibur repository at excaliburjs/Excalibur.

### Environment
Travis CI

### Current Workaround
Developers could create a personal access token and set `GH_TOKEN` in their Travis CI environment to the value of that token. For security the token should not have any security scopes assigned.

djcsdy:

Closes #1309 

This is the minimum possible change to fix #1309:

 - Removes the requirement for `GH_TOKEN` to be set in order to build on CI.
 - Removes the `Authorization` header from the request to `api.github.com/repos/excaliburjs/Excalibur/releases/latest`, since it is not needed.

In addition to this change it might also be worth considering checking `TRAVIS_REPO_SLUG`. If this is not equal to `excaliburjs/Excalibur`, then a fork is being built and version numbering could be handled similarly to the way it is currently handled when building a PR branch.

### DIFF
--- a/version.js
+++ b/version.js
@@ -1,77 +1,68 @@
-const { execSync }  = require('child_process');
-const request       = require('sync-request');
+const { execSync } = require('child_process');
+const request = require('sync-request');
 const appveyorBuild = process.env.APPVEYOR_BUILD_NUMBER || '';
-const travisBuild   = process.env.TRAVIS_BUILD_NUMBER   || '';
-const commit        = process.env.TRAVIS_COMMIT         || '';
-const travisPr      = process.env.TRAVIS_PULL_REQUEST   || 'false';
-const travisTag     = process.env.TRAVIS_TAG            || '';
+const travisBuild = process.env.TRAVIS_BUILD_NUMBER || '';
+const commit = process.env.TRAVIS_COMMIT || '';
+const travisPr = process.env.TRAVIS_PULL_REQUEST || 'false';
+const travisTag = process.env.TRAVIS_TAG || '';
 
 // ignore local builds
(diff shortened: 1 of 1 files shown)


djcsdy:
...and of course, the build immediately fails with 'Fatal error fetching GH release version' thus proving me wrong, despite the fact that it worked every time I tested it up to the point of actually creating a PR 🤦‍♂.

djcsdy:
Closing this because clearly I was fundamentally mistaken.

Issue #1092, opened 2019-01-26
Implement Draggable Helper on Actors


eonarheim:
### Context
In order to properly implement drag and drop on an actor, it takes a lot of ceremonies to make it work. There should be a simpler way to enable actor drag and drop for actors.

Currently this:

```typescript
var block2 = new ex.Actor({
  pos: ex.Vector.Zero.clone(),
  anchor: ex.Vector.Zero.clone(),
  width: 50,
  height: 50,
  color: ex.Color.Red.clone(),
  collisionType: ex.CollisionType.PreventCollision
});

// Set drag flag
var boxPointerDragging = false;
block2.on('pointerdragstart', (pe: ex.Input.PointerEvent) => {
  boxPointerDragging = true;
});

// Set drag flag
block2.on('pointerdragend', (pe: ex.Input.PointerEvent) => {
  boxPointerDragging = false;
});

// Drag box around
block2.on('pointerdragmove', (pe: ex.Input.PointerEvent) => {
  if (boxPointerDragging) {
    block2.pos = pe.pointer.lastWorldPos;
  }
});

// Drag box around
block2.on('pointerdragleave', (pe: ex.Input.PointerEvent) => {
  if (boxPointerDragging) {
    block2.pos = pe.pointer.lastWorldPos;
  }
});
```

### Proposal

Add a `ex.Actor.draggable = true` utility gettter/setter/method to help remove this boilerplate

MaanasArora:
Can I work on this issue?

eonarheim:
@MaanasArora All yours! I've put the community in-progress label on the issue. Please open a draft PR as soon as you can.

MaanasArora:
@eonarheim I have created a draft PR—sorry for the time taken! I was making sure I did everything correctly (I'm new to open source)

PR #1323, opened 2019-11-02
[#1092] Add draggable getter/setter

MaanasArora:
## Changes:

- drafted ex.Actor.draggable utility getter/setter

### DIFF
--- a/src/engine/Actor.ts
+++ b/src/engine/Actor.ts
@@ -539,11 +539,7 @@ export class ActorImpl extends Class implements Actionable, Eventable, PointerEv
 
   public set draggable(isDraggable: boolean) {
     if (isDraggable) {
-      this._draggable = true;
-
-      if (isDraggable) {
-        this.collisionType = CollisionType.PreventCollision;
-
+      if (isDraggable && !this._draggable) {
         this.on('pointerdragstart', () => {
           this._dragging = true;
         });
@@ -563,9 +559,29 @@ export class ActorImpl extends Class implements Actionable, Eventable, PointerEv
             this.pos = pe.pointer.lastWorldPos;
           }
         });
-      } else {
-        this._draggable = false;
+      } else if (!isDraggable && this._draggable) {
+        this.off('pointerdragstart', () => {
+          this._dragging = true;
+        });
+
+        this.off('pointerdragstart', () => {
+          this._dragging = false;
+        });
+
+        this.off('pointerdragmove', (pe: PointerEvent) => {
+          if (this._dragging) {
+            this.pos = pe.pointer.lastWorldPos;
+          }
+        });
+
+        this.off('pointerdragleave', (pe: PointerEvent) => {
+          if (this._dragging) {
+            this.pos = pe.pointer.lastWorldPos;
+          }
+        });
       }
+
+      this._draggable = isDraggable;
     }
   }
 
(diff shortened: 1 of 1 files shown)


kamranayub:
Thank you @MaanasArora, nice start! I was wondering @eonarheim if you foresee any issues adding handlers during the set operation. One thing I think we'd want to make sure to do is remove the handlers when `draggable` is unset.

Is this something ECS will help with, would `Draggable` be a component that just gets added/removed? I don't know enough to say right now but that's what came to mind.

eonarheim:
Thanks @MaanasArora this is a good start! I agree, we should be sure that the handlers are removed when unset, and that multiple handlers cannot be accidentally set if `.draggable = true` is called multiple times.

@kamranayub I think you are on the right track. For now I don't think we should consider an ECS implementation, we'll likely want to look into a `Draggable` Component/System pair when we have a a more solid ECS foundation.

src/engine/Actor.ts
```diff
@@ -527,6 +527,48 @@ export class ActorImpl extends Class implements Actionable, Eventable, PointerEv
...
+  public get draggable(): boolean {
+    return this._draggable;
+  }
+
+  public set draggable(isDraggable: boolean) {
+    if (isDraggable) {
+      this._draggable = true;
+
+      if (isDraggable) {
+        this.collisionType = CollisionType.PreventCollision;
```
eonarheim: Setting the collision type to `PreventCollision` when `draggable = true` may be unexpected for users of Excalibur, I vote we do not set collision type. Thoughts @kamranayub?

MaanasArora:
Thanks, @kamranayub @eonarheim! 😃 I will update the PR with the edits you pointed out:

- remove handlers when `draggable` is unset
- ensure that multiple handlers are not accidentally added (by making sure `draggable` was not already set before adding handlers)
- remove the line that sets `collisionType` to `PreventCollision`

MaanasArora:
I have committed my changes :)

src/engine/Actor.ts
```diff
@@ -527,6 +527,64 @@ export class ActorImpl extends Class implements Actionable, Eventable, PointerEv
...
+      } else if (!isDraggable && this._draggable) {
+        this.off('pointerdragstart', () => {
+          this._dragging = true;
+        });
+
+        this.off('pointerdragstart', () => {
+          this._dragging = false;
+        });
+
+        this.off('pointerdragmove', (pe: PointerEvent) => {
```
eonarheim: @MaanasArora Last request for this PR, thanks for working so hard on this. In order for `this.off(...)` to work correctly it needs to be based the same function reference that was used in the `this.on(...)`. We'll need to store those somewhere, perhaps as private methods on actor?

Something like
```typescript
private _pointerDragStartHandler = () => {
   this._dragging = true;
}
...

 public set draggable(isDraggable: boolean) {
    if (isDraggable) {
      if (isDraggable && !this._draggable) {
        this.on('pointerdragstart', this._pointerDragStartHandler );
      ...
      } else if (!isDraggable && this._draggable) {
        this.off('pointerdragstart', this._pointerDragStartHandler );
      ...
}

MaanasArora:
@eonarheim No problem, sorry for the error, and thank you for bringing it to attention! I followed your suggestion to store the functions as private methods and committed the changes.

MaanasArora:
Alright, thanks! I have marked the PR as ready for review.

eonarheim:
@MaanasArora Thanks again!

PR #1491, opened 2020-04-14
Fix freeze when actor width or height is 0 and tilemap is shown

### LINKED ISSUE #1418: Excalibur crashes when using tilemaps and actors together
### Steps to Reproduce
```ts
import { Engine, Loader, Actor } from 'excalibur';
import { TiledResource } from '@excaliburjs/excalibur-tiled';

const game = new Engine();
const loader = new Loader();
const levelResource = new TiledResource('assets/map/single-layer.json');
loader.addResource(levelResource);

game.start(loader).then(() => {
  const tileMap = levelResource.getTileMap();
  game.addTileMap(tileMap);

  const actor = new Actor();
  game.add(actor); // Game crashes on this line
});
```

### Expected Result
The map should display, along with any defined actors.

### Actual Result
The entire browser tab freezes and needs to be force closed.

### Environment
- browsers and versions: Chromium 79.0.3945.130, Firefox 72.0.2
- operating system: Ubuntu 19.10
- Excalibur versions: 0.23.0

### Notes
- UIActors are fine; the problem specifically seems to happen when combining the Tiled TileMap and an Actor.
- `import * as ex from 'excalibur'` instead of the above import style makes no difference.

### Current Workaround
None.

DaVince:
Closes #1418 

TileMapImpl.collides() got stuck infinitely on the `trace points for overlap` loop when the actor's size is zero.

## Changes:

- Check size of actor before deciding whether to run the collision logic
- Return null when actor width or height is 0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fixed Excalibur crashing when displaying both a tilemap and a zero-size actor ([#1418](https://github.com/excaliburjs/Excalibur/issues/1418))
 - Fixed animation flipping behavior ([#1172](https://github.com/excaliburjs/Excalibur/issues/1172))
 - Fixed actors being drawn when their opacity is 0 ([#875](https://github.com/excaliburjs/Excalibur/issues/875))
 - Fixed iframe event handling, excalibur will respond to keyboard events from the top window ([#1294](https://github.com/excaliburjs/Excalibur/issues/1294))
--- a/src/engine/TileMap.ts
+++ b/src/engine/TileMap.ts
@@ -86,6 +86,9 @@ export class TileMapImpl extends Class {
     const height = actor.pos.y + actor.height;
     const actorBounds = actor.body.collider.bounds;
     const overlaps: Vector[] = [];
+    if (actor.width <= 0 || actor.height <= 0) {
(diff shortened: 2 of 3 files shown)


DaVince:
This is my second merge request on an open source project ever, please let me know if I did it right! 🙂

DaVince:
In the failing test, an object with a non-zero size is being tested. My implementation might have been naive. Will look into this further.

DaVince:
Test `'should collide with other polygons when touching'` fails, but I have no clue why, because on further inspection it does nothing with TileMaps. Is it even related?

On another note: can actor widths and heights ever be a negative value?

eonarheim:
I think this is a flakey test that was discovered, not related to TileMaps.

I think negative widths/heights are unexpected and should be an error. Zero
width and heights do have a use.

eonarheim:
@kamranayub ^ Any more info on this test

eonarheim:
@DaVince This looks great, is it possible to add a test for this fix?

eonarheim:
If it is too onerous to add a test for this let me know.

kamranayub:
I wouldn't worry about that test, it has failed in master before and will pass on a rebuild. Definitely flaky.

DaVince:
Thanks for all the great feedback!

> @DaVince This looks great, is it possible to add a test for this fix? (ignore the flakey test)

Sure, I'm on it.

PR #1507, opened 2020-04-26
fix: [#1417] [#1431] Loader positioning, allow customization, clean-up html

### LINKED ISSUE #1417: Play button position cannot be set with CSS
### Steps to Reproduce
Override the load button CSS by giving it the following properties:
```css
position: relative;
top: 25px;
```

### Expected Result
The button should be positioned at 25 pixels from the top of the game canvas.

### Actual Result
The  button is positioned 25 pixels down from the `_playButtonRootElement`, which is given a specific position and size, so the positioning happens relative to this element and the CSS overflow causes part of the button to be hidden.

image

(Note that I've switched to using a loader with a customized style to just get around this problem, without success, because the root element cannot be affected. Also, the off-center positioning of the button is caused by me using `DisplayMode.Fixed` and then changing the DOM canvas size to be 2x or 3x the normal size for perfect pixel scaling.)

### Environment
- Chromium79.0.3945.130
- Ubuntu 19.10
- Excalibur 0.23.0

### Current Workaround
Using a custom loader that doesn't have a button would get around this entirely.

### Suggestions for solutions
I can think of a few ways and levels to fix this on, but the most obvious one would be in the Loader class. The [Loader class](https://excaliburjs.com/docs/api/edge/classes/_loader_.loader.html) already provides customization options like changing the logo size and loader background. It would make sense to have options that allow you to set the positions of the various elements here.

```typescript
loader.autoPosition = false; // When true, automatically position using the current method
loader.playButtonPosition = new Vector(20, 20);
loader.logoPosition = new Vector(200, 20);
```

### LINKED ISSUE #1431: Play button is not cleaned up on stop
### Steps to Reproduce

1. Create a game
2. Start the game
3. Stop the game

### Expected Result

Play button should disappear/be cleaned up

### Actual Result

Play button sticks around

### Environment

- browsers and versions: Chrome 79
- operating system: Mac OS
- Excalibur versions: 0.23.0

### Current Workaround

Target HTML element manually and delete from DOM

eonarheim:

==================

Closes #1417
Closes #1431

## Changes:

- Adds an id to the play button root element `#excalibur-play-root`
- Allows modification of all loader positioned elements
   - Logo
   - Play button
   - Loading bar
- Additionally allows the loading bar color to be set


eonarheim:
@kamranayub @jedeen Planning to merge this today if no objections

kamranayub:
Think you'd be willing to fix #1431 too 😄

eonarheim:
Haha sure, I'll roll in #1431 @kamranayub

eonarheim:
@kamranayub all fixed up :+1:

PR #1511, opened 2020-04-28
fix(Keyboard): Fix accessing cross-origin frame exception

### LINKED ISSUE #1510: Embedding a page with Excalibur in an iframe from another domain results in an exception
### Steps to Reproduce

On Slides.com, try to embed a page with Excalibur running using the Iframe feature.

For example, https://excaliburjs.com/ludum-46/

### Expected Result

The game should start.

### Actual Result

You get an exception due to Keyboard (I think) attempting to access the parent frame (which is cross-origin) and fails.

image

### Environment

- browsers and versions: Chrome 81
- operating system: Windows 10
- Excalibur versions: 0.24.1 (0.23.x?)

It looks like Office Daydream embeds fine: http://excaliburjs.com/ludum-41/

But The Show Must Go On does not: https://excaliburjs.com/ludum-46/

I was testing using Storybook page: https://excaliburjs.com/examples/iframe.html?id=actions--fade

It looks like this may have been introduced in the time between Ludum 41 and Ludum 46?

### Current Workaround

Don't embed it in an iframe or use previous working Excalibur version.

kamranayub:

==================

Closes #1510

## Changes:

- Fix `Keyboard.init` throwing error when embedded in cross-origin iframe
- Add Keyboard story
- Fix Anchors stories so they show up again
- Fix Pointer stories and use `PointerScope.Canvas` in Storybook
- Disable Storybook shortcuts so they don't conflict with keyboard-enabled stories

Example of it working in iframe:

image

### DIFF
--- a/.storybook/manager.js
+++ b/.storybook/manager.js
@@ -1,0 +1,7 @@
+import { addons } from '@storybook/addons';
+
+addons.setConfig({
+
+  // Excalibur will be listening for all key events
+  enableShortcuts: false
+})
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fixed Excalibur crashing when embedded within a cross-origin IFrame ([#1151](https://github.com/excaliburjs/Excalibur/issues/1151))
+
 <!--------------------------------- DO NOT EDIT BELOW THIS LINE --------------------------------->
 <!--------------------------------- DO NOT EDIT BELOW THIS LINE --------------------------------->
 <!--------------------------------- DO NOT EDIT BELOW THIS LINE --------------------------------->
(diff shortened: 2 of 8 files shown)


kamranayub:
@eonarheim 

OK, so I discovered some things when testing this in jsfiddle (embed localhost `<iframe>`):

- Keyboard events will not fire AT ALL on `document` when doing cross-origin iframes
- Keyboard events WILL fire if a wrapping element around `<canvas />` is used
- Invoking Storybook's `action` helper is expensive and slows down FPS plus it _also_ triggers cross-origin exceptions

Implementing an equivalent `KeyboardScope` like `PointerScope` is more involved because we'd need to re-parent the canvas and the initialization routine happens before the game is actually started, so the `<canvas>` element may be in a detached state (which is the case for Storybook, actually).

Pointer events avoid this issue entirely because `canvas` receives events fine. I also tried to do a "hack" using `tabindex="1"` [as mentioned here](https://stackoverflow.com/questions/56771/how-do-you-set-focus-to-the-html5-canvas-element/57332#57332) but didn't get it to work.

Therefore, I will open a new issue to propose adding `KeyboardScope` so that people can fallback to listening to "canvas" key events if they want.

kamranayub:
Added https://github.com/excaliburjs/Excalibur/issues/1512

Issue #1513, opened 2020-04-28
Test fail on fresh clone


Xstoudi:
### Steps to Reproduce
```
git clone https://github.com/excaliburjs/Excalibur.git
npm install
npm run dev
```
### Expected Result
Run without failing

### Actual Result
16 FAILED

### Environment

- browsers and versions: test runner
- operating system: Windows 10
- Excalibur versions: master branch

Xstoudi:
Ok seems, like it is flacky tests : images diff often fail (between 96% and 99.50%) because it attempts 99.5% match.

eonarheim:
@Xstoudi which tests in particular are failing for you? Definitely not ideal if tests are failing on one machine and not another.

Xstoudi:
Ok, on Ubuntu everything seems to work correctly.

eonarheim:
@Xstoudi mine seem to pass on win10 running in chrome headless, what build do you have?

eonarheim:
Run `winver.exe` to output build number

Xstoudi:
Windows 10 - Version 1909. I figured out that `npm install` wasn't downloading Puppeteer's Chromium binary. More exactly, it downloads it but it doesn't unzip it.

PR #1514, opened 2020-04-28
fix: [#805] Replace mock.engine by real engine

### LINKED ISSUE #805: Replace mock.engine with real engine
### Context

I think new and core contributors alike are getting a bit confused with what approach to test the engine. We don't need to use `mock.engine` anymore, that is deprecated. This was used back before engine's dependencies were made testable.

The only key thing to remember is to stop the engine after each test so it gets cleaned up, otherwise there will be multiple engines running which causes all sorts of issues.

### Proposal

Delete `mock.engine` and replace with `TestUtil.engine`. See ScaleSpec for recommended approach. I *think* the loop mock can also be deprecated, but it could also be retooled to work with the real engine.

Xstoudi:
Changes:

- Remove usage of `mock.engine` from the tests. Use real engine instead.

==================

Closes #805 

## Changes:

- Remove usage of `mock.engine` from the tests. Use real engine instead.


Xstoudi:
Help wanted. Issue #1513 makes my life harder to debug my change.

eonarheim:
@Xstoudi thanks for the work on this! Sorry about the delay life has been crazy, hope you don't mind me fixing up the test

Xstoudi:
No problem @eonarheim! I couldn't figured out why the test were failing. Thank you for fixing it :)

PR #1526, opened 2020-05-01
[#1498] Implement Camera Bounds Strategy

### LINKED ISSUE #1498: Allow for camera constraints
### Context
Right now, a camera can move anywhere and everywhere. In the context of showing a map, this will mean that when bound to an Actor, the camera moves beyond the limits of the map and shows sudden emptiness.

### Proposal
I propose to add an extra constraints property to the CameraStrategy class, that, when set, will stop the camera once it goes beyond the defined constraints.

Example:
```ts
this.cameraStrategy.constraints = new Vector(mapWidth, mapHeight);
```
CameraStrategy.action() is then expected to follow the constraints and return the appropriate value.

DaVince:

Closes #1498

## Changes:

- Adds a new camera strategy named LimitCameraBoundsStrategy.
- Adds the strategy to the StrategyContainer.


DaVince:
I'd like some feedback on this before I add tests.

Also, is that failed PR Labeler check going to be an issue? It only exists because I forgot to sync my fork before working on this...

DaVince:
An example implementation is available at https://github.com/DaVince/excalibur-tiled-examples/blob/master/src/index.ts#L24

src/engine/Camera.ts
```diff
@@ -170,6 +178,35 @@ export class RadiusAroundActorStrategy implements CameraStrategy<Actor> {
   };
 }
 
+/**
+ *  Prevent a camera from going beyond the given camera dimensions.
+ */
+export class LimitCameraBoundsStrategy implements CameraStrategy<BoundingBox> {
```
eonarheim, line 184: @DaVince this is great, exactly what I was thinking for implementation

DaVince: Cool, I'll start writing some tests then!

DaVince:
Added tests, changelog, more documentation.

I'd like to note that this strategy does not transform any of the other strategies or camera actions. So say you try to `move()` the camera beyond the boundaries with an easing function, the strategy won't change the `move()` target to your bounding box limits. This can cause a jarring effect (sudden stop) but I figure that it's better dealt with within a game dev's own code. Maybe I can implement it later as an optional feature.

DaVince:
I need to add some more tests to this that checks if it handles non-zero x and y start positions properly.

DaVince:
Squashed a bug relating to my previous post.

I also observed crazy jittery behavior when the bounding box is smaller than the screen size. The only way I can fix that is by growing the bounding box on initialization. Is there any way to do that without passing an instance of `Engine` to the constructor? I don't want to run this logic on every `action()`.

DaVince:
If that last thing is a non-issue, this PR is good to go. :)

eonarheim:
> Squashed a bug relating to my previous post.
> 
> I also observed crazy jittery behavior when the bounding box is smaller than the screen size. The only way I can fix that is by growing the bounding box on initialization. Is there any way to do that without passing an instance of `Engine` to the constructor? I don't want to run this logic on every `action()`. (Or maybe I _should,_ considering a dev might change the size of the target bounding box?)

I think it's a non-issue, we could possibly log a warning about the bound box size and/or some documentation in the strategy about the expected bounds.

DaVince:
I've added documentation and a warning.

The warning is implemented a little clumsily inside of the `action` method since I don't have access to Engine properties in the constructor.

eonarheim:
@DaVince thanks again for your contributions!

PR #1531, opened 2020-05-03
[#1512] WIP Add basic KeyboardScope

### LINKED ISSUE #1512: Proposal: Add KeyboardScope option
### Context

There are some issues using Excalibur in cross-origin iframes, see #1294 and #1510 

What happens is that event listeners cannot listen on the top frame. In order to provide a fallback, we specific `window` as the target to listen to key events _but that doesn't work either_ (according to our code, it never did).

There are some approaches we can try that will enable keyboard events when embedding cross-origin, which is by targeting the canvas or an element surrounding the canvas.

### Proposal

**Option 1: Try to use tabindex and focus**

According to [this SO answer](https://stackoverflow.com/questions/56771/how-do-you-set-focus-to-the-html5-canvas-element/57332#57332) it should be possible to attach listeners if the canvas has `tabindex="1"` and is in focus. I couldn't get this to work, though, but I didn't have time to explore it much.

**Option 2: Use a wrapping element**

The other way is by targeting an element that wraps the canvas. A `<div>` or a `<span>` may work (I like the inline element since it should wrap to the bounds of the canvas better).

This **does work** and I got it to work by adding a new option to allow passing in the element ID or element to target keyboard events on. In Storybook, this is `document.getElementById('root')`.

However, I thought it may be better to implement the same idea we have for `PointerScope` and do a `KeyboardScope` and _automatically generate the wrapping capture element_

This presents some issues though, when  the canvas is detached from the DOM. It can't be re-parented to be under the capture element.

Ideally, this keyboard initialization logic would run when the game boots up in the DOM for the first time and I got close but not quite in my spike of doing this.

```
export enum KeyboardScope {
  Document,
  Canvas
}
```

eonarheim:

==================

Partial fix #1512

## Changes:

- Add KeyboardScope option to Engine similar to PointerScope

### DIFF
--- a/src/engine/Engine.ts
+++ b/src/engine/Engine.ts
@@ -134,6 +134,12 @@ export interface EngineOptions {
   pointerScope?: Input.PointerScope;
 
   /**
+   * Configures the pointer scope. Pointers scoped to the 'Canvas' can only fire events within the canvas viewport; whereas, 'Document'
+   * (default) scoped will fire anywhere on the page.
+   */
+  keyboardScope?: Input.KeyboardScope;
+
+  /**
    * Suppress boot up console message, which contains the "powered by Excalibur message"
    */
   suppressConsoleBootMessage?: boolean;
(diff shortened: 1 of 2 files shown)


eonarheim:
Close for now, does not work as intended

PR #1550, opened 2020-05-16
fix: [#1549] Remove unecessary sprite effect for opacity

### LINKED ISSUE #1549: Opacity sprite effect is unnecessary and causes perf degradation
### Steps to Reproduce
When adding a new drawing to excalibur it currently triggers sprite effect calculations on opacity (whether or not it needs it). Additionally there is no need to use sprite effects for opacity anymore.

image

eonarheim:

==================

Closes #1549

## Changes:

- Remove opacity sprite effect from actor
- Only process effects if necessary


eonarheim:
@kamranayub Definitely, just haven't had time yet 👍

src/engine/Drawing/Animation.ts
```diff
@@ -277,24 +280,35 @@ export class AnimationImpl implements Drawable {
...
+      this._drawWithOptions({ ctx: ctxOrOptions, x, y });
     } else {
       this._drawWithOptions(ctxOrOptions);
     }
   }
 
   private _drawWithOptions(options: DrawOptions) {
+    const animOptions = {
+      ...options,
+      rotation: nullish(options.rotation, this.rotation),
```
kamranayub, line 292: Does nullish account for 0 number or empty strings, since those aren't nullish? Sorry, I can't see the impl easily on my phone

kamranayub:
Okay I looked at the impl, question though: with TS 3.9 can we just use null coalescing? It would let us do the same right?

eonarheim:
> Okay I looked at the impl, question though: with TS 3.9 can we just use null coalescing? It would let us do the same right?

You are correct, I think `nullish()` was added to excalibur before `??` was at the final stage. Definitely will change it before merging

eonarheim:
@kamranayub Updated :+1:

PR #1556, opened 2020-05-21
fix: [#1555] Pointer events should only work on actors in scene

### LINKED ISSUE #1555: Actors not in Scene still receive pointer events
This bug was found and fixed in Ludum 46

### Steps to Reproduce
* Create an actor in a scene and attach a pointer event to it.
* Remove it from the scene with `kill()`
* Click in the same area where the actor was and the pointer event still fires

### Expected Result
Pointer events should only fire on actors in scene

### Actual Result
Pointer events fire on all actors 

### Environment
- browsers and versions: N/A
- operating system: N/A
- Excalibur versions: master

### Current Workaround
Double check in the handler that the event should be processed

eonarheim:

==================

Closes #1555

## Changes:

- Checks if actors are alive before dispatching events
- Renamed method to be clearer
- Adds a test


src/engine/Input/Pointer.ts
```diff
@@ -223,8 +223,8 @@ export class Pointer extends Class {
    * Checks if Pointer has a specific Actor in ActorsUnderPointer list.
    * @param actor An Actor for check;
    */
-  public isActorUnderPointer(actor: Actor): boolean {
-    return this._actorsUnderPointer.hasOwnProperty(actor.id.toString());
+  public isActorAliveUnderPointer(actor: Actor): boolean {
+    return !!(!actor.isKilled() && actor.scene && this._actorsUnderPointer.hasOwnProperty(actor.id.toString()));
```
kamranayub, line 227: It'll be good to go through the codebase and replace these dicts with Sets and Maps now they we have more efficient structures.

src/spec/PointerInputSpec.ts
```diff
@@ -129,4 +129,9 @@ describe('A pointer', () => {
 
     expect(engine.input.pointers.primary.checkActorUnderPointer(actor)).toBe(true);
   });
+  it('should only add actors under pointer that are in scene', () => {
```
kamranayub, line 132: :thumbsup:

src/engine/Input/Pointer.ts
eonarheim: Do Sets and Maps downlevel well in TypeScript?

PR #1598, opened 2020-07-02
feat: Screen Resolution Abstraction

### LINKED ISSUE #1617: Refactor Screen Logic to Support Independent Viewport and Resolution
### Context
The Screen does not have a convenient single place for logic to live.

Currently it is difficult build a game that has a different resolution than the viewport, this make it difficult to build lofi art style games. Additionally the current "Fullscreen" only means dynamic aspect ratio taking up the available window, there is now the browser Fullscreen API

### Proposal

* [ ] Refactor screen logic to one place
* [ ] Support independent viewport/resolution controls
* [ ] Support the browser fullscreen api

eonarheim:

==================

Closes #1617 

## Changes:

- Refactors and creates a screen abstraction for managing all things resolution & viewport related logic
- Adds and fixes tests for the screen abstraction

### DIFF
--- a/src/engine/Engine.ts
+++ b/src/engine/Engine.ts
@@ -73,7 +73,7 @@ export interface EngineOptions {
   height?: number;
 
   /**
-   * Optionally configure the width & height of the viewport in css pixels. 
+   * Optionally configure the width & height of the viewport in css pixels.
    * Use `viewport` instead of [[EngineOptions.width]] and [[EngineOptions.height]], or vice versa.
    */
   viewport?: ScreenDimension;
--- a/src/engine/Engine.ts
+++ b/src/engine/Engine.ts
@@ -293,7 +293,7 @@ export class Engine extends Class implements CanInitialize, CanUpdate, CanDraw {
    * Indicates whether the engine is set to fullscreen or not
    */
   public get isFullscreen(): boolean {
-    return this.screen.isFullscreen;
+    return this.screen.isFullScreen;
   }
 
   /**
@@ -608,7 +608,6 @@ O|===|* >________________>\n\
    * @param x          x game coordinate to play the animation
    * @param y          y game coordinate to play the animation
    */
-  @obsolete()
   public playAnimation(animation: Animation, x: number, y: number) {
     this._animations.push(new AnimationNode(animation, x, y));
   }
@@ -1118,7 +1117,6 @@ O|===|* >________________>\n\
     this.screen.pushResolutionAndViewport();
     this.screen.resolution = this.screen.viewport;
     this.screen.applyResolutionAndViewport();
-    this.graphicsContext.updateViewport();
     let loadingComplete: Promise<any>;
     if (loader) {
       this._loader = loader;
@@ -1132,7 +1130,6 @@ O|===|* >________________>\n\
     loadingComplete.then(() => {
       this.screen.popResolutionAndViewport();
       this.screen.applyResolutionAndViewport();
-      this.graphicsContext.updateViewport();
       this.emit('start', new GameStartEvent(this));
     });
 
@@ -1261,7 +1258,6 @@ O|===|* >________________>\n\
 /**
  * @internal
  */
-@obsolete()
 class AnimationNode {
   constructor(public animation: Animation, public x: number, public y: number) {}
 }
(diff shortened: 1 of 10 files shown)


src/engine/Engine.ts
```diff
@@ -81,32 +58,25 @@ export enum ScrollPreventionMode {
...
-  bottom?: number | string;
-}
-
 /**
  * Defines the available options to configure the Excalibur engine at constructor time.
  */
 export interface EngineOptions {
   /**
-   * Optionally configure the native canvas width of the game
+   * Optionally configure the native logical pixel width of the game (resolution)
```
kamranayub, line 66: Might be nice to include symbol link to `Resolution`

```suggestion
   * Optionally configure the native logical pixel width of the game. See [[Resolution]] for common presets.
```

src/engine/Engine.ts
```diff
@@ -81,32 +58,25 @@ export enum ScrollPreventionMode {
...
    */
   width?: number;
 
   /**
-   * Optionally configure the native canvas height of the game
+   * Optionally configure the native logical pixel height of the game (resolution)
```
kamranayub, line 71:
```suggestion
   * Optionally configure the native logical pixel height of the game. See [[Resolution]] for common presets.
```

src/engine/Engine.ts
```diff
@@ -81,32 +58,25 @@ export enum ScrollPreventionMode {
...
    */
   height?: number;
 
+  /**
+   * Optionally specify the size the viewport should take up onscreen
+   */
+  viewport?: { width: number; height: number };
```
kamranayub, line 78: My only concern is that I think this represents a change in the way `width` and `height` is interpreted from previous versions. So, if someone had set their WxH to 1024x768, uptaking this will "work" but technically the meaning of `width` and `height` is different than what they had previously thought.

I wonder if it would be a smoother upgrade path to flip this. Instead of `width` and `height` being resolution, maybe `viewport` should change to `resolution`.

That way, using my example, if I had code like this:

```js
const game = new Engine({
  width: 1024,
  height: 768
});
```

Then after uptaking this next version, I could adjust the resolution of the game using:

```js
const game = new Engine({
  width: 1024,
  height: 768,
  resolution: Resolution.Gameboy
});
```

It also works well with the new `Resolution` presets.

Thoughts?

src/engine/Screen.ts
```diff
@@ -0,0 +1,500 @@
...
+    return this._resolution.height * this.pixelRatio;
+  }
+
+  public setCurrentCamera(camera: Camera) {
+    this._camera = camera;
+  }
+
+  private _onPixelRatioChange(_event: 'pixelRatioChange', handler: () => any) {
+    const mqString = `(resolution: ${window.devicePixelRatio}dppx)`;
+    matchMedia(mqString).addEventListener('change', handler);
```
kamranayub, line 217: We should consider cleaning up event listeners. Each time a new engine is created, we'd be adding these listeners, right?

Also should this be:

```suggestion
    this._browser.window.matchMedia(mqString).addEventListener('change', handler);
```

src/engine/Screen.ts
```diff
@@ -0,0 +1,500 @@
...
+    return this.drawHeight / 2;
+  }
+
+  private _applyDisplayMode() {
+    if (this.displayMode === DisplayMode.FullScreen || this.displayMode === DisplayMode.Container) {
+      const parent = <any>(this.displayMode === DisplayMode.Container ? <any>(this.canvas.parentElement || document.body) : <any>window);
+
+      this._setHeightByDisplayMode(parent);
+
+      this._browser.window.on('resize', () => {
```
kamranayub, line 396: Same here, we should clean this up somehow...

version.js
```diff
@@ -19,7 +19,8 @@ function getCurrentCommit() {
 }
 
 function getLatestTag(commit) {
-  execSync('git fetch');
+  execSync('git fetch --depth=1000');
```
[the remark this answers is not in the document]
eonarheim: Temporary fix, Travis does a shallow clone :(

src/engine/Engine.ts
eonarheim: We spoke offline, this makes a lot of sense.

version.js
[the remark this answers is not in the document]
djcsdy: You can get a deep clone on Travis by adding `git fetch --unshallow` to your `before_script`:

```travis.yml
before_script:
  - git fetch --unshallow
```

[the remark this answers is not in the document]
eonarheim: Thanks, I might switch to this. Seems far less brittle

PR #1670, opened 2020-10-03
feat: [#1670] Add key code enums and KeyboardEvent.code support

### LINKED ISSUE #1654: Add full ASCII printable set of keys and switch to KeyboardEvent.code
### Context

Currently Excalibur does not have every printable ASCII character in it's constants and is using the deprecated `KeyboardEvent.keyCode`. We are no longer supporting IE11 so no need to support `keyCode`

### Proposal

* Add a constant for every ~~printable ASCII~~ physical key in Excalibur.
  * Current constants using deprecated keyCode property
        https://github.com/excaliburjs/Excalibur/blob/main/src/engine/Input/Keyboard.ts#L8-L59

* Update logic to use `KeyboardEvent.code` https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code
https://github.com/excaliburjs/Excalibur/blob/main/src/engine/Input/Keyboard.ts#L132

See ascii table https://theasciicode.com.ar/ascii-control-characters/horizontal-tab-ascii-code-9.html

rledford:
Closes #1654

## Changes
- Converted `Keys` enum value type from _number_ to _string_
- Changed `Keyboard` _keys, _keysUp, and _keysDown type from _number[]_ to _Keys[]_
- Updated `Keyboard` event handlers to use updated `Keys` enum 
- Added more key codes to `Keys` enum
- Added comments for key definition groups and sorted


src/engine/Input/Keyboard.ts
```diff
@@ -208,4 +266,11 @@ export class Keyboard extends Class {
         return code;
     }
   }
+
+  /**
+   * Gets the key code from KeyboardEvent.code if supported otherwise from KeyboardEvent.keyCode
+   */
+  private _getKeyCode(ev: KeyboardEvent): number {
+    return ev.code !== undefined ? parseInt(ev.code, 10) : ev.keyCode;
```
eonarheim, line 274: @rledford I don't think that `parseInt` will work on `ev.code` as expected.  `ev.code` is a string constant of the form "KeyA" 

image

See sandbox https://codesandbox.io/s/key-code-y6994?file=/src/index.ts

src/engine/Input/Keyboard.ts
```diff
@@ -6,6 +6,7 @@ import * as Events from '../Events';
  * Enum representing input key codes
  */
 export enum Keys {
```
eonarheim, line 8: Since we are switching to `KeyboardEvent.code` we should switch this enum to be an enum of strings

```typescript
export enum Keys {
   Num1 = "Numpad1",
   A = "KeyA"
...
```

src/engine/Input/Keyboard.ts
eonarheim: We are dropping support for IE so we don't need to worry about the the case where `ev.keyCode` is present and `ev.code` is not. Any objections to that @jedeen @kamranayub?

eonarheim (changes requested):
@rledford Thanks for adding the missing enum values, request a change to the enum and key handling logic

src/engine/Input/Keyboard.ts
rledford: the keyboard events in the chromium browser in the karma tests only have keyCode and I didn’t want to push anything that didn’t pass. I’ll change it to only use the KeyboardEvemt.code though.

src/engine/Input/Keyboard.ts
rledford: you’re right. i’ll get this sorted out.

src/engine/Input/Keyboard.ts
eonarheim: We'll likely need to update some tests :+1: thanks!

rledford:
@eonarheim, on my local branch I've converted the Keys enum to use the appropriate string values. In order to resolve a linter error I've cast the `KeyboardEvent.code` from a _string_ to _Keys_ type like this.

```js
// in keydown/keyup event listeners
const code = ev.code
// ...
const keyEvent = new KeyEvent((<any>Keys)[code]);
```

Everyone ok with this?

rledford:
In order to allow dependencies of `KeyEvent` and `Keys` to reference keys using bracket notation, in files like `/src/stories/Keyboard.stories.ts`, the `Keys` enum keys must match the string value. This could be a breaking change since I have to change ```A = 'KeyA'``` to ```KeyA = 'KeyA'```. Thoughts?

eonarheim:
@rledford One way to avoid the breaking change may be to treat the `KeyboardEvent.code` as a `Keys` type, which is technically true given the recent enum change. 

It is unfortunate that typescript enums don't have a convenient reverse lookup unless the name matches the value. 

Something like this?

```typescript
const enum Keys {
  A = "KeyA"
}

class KeyEvent {
  constructor(public key: Keys) {}
}

document.addEventListener("keydown", (ev) => {
  const key: Keys = ev.code as Keys;
  const keyEvent = new KeyEvent(key);
  console.log(key);
  console.log(Keys.A);
});
```

Example code sandbox
https://codesandbox.io/s/key-casting-deogv

rledford:
ok thanks.

rledford:
@eonarheim, the Keys enum has been changed to use string values. I updated KeyInputSpec to mock key events with a `code` property. All tests pass now. I also updated the Keyboard.stories to use the new Keys enum implementation.

rledford:
I'm investigating the build failure caused by Keys in src/engine/Engine.ts

rledford:
I'm going to have to rework some things again.

rledford:
I've updated `Keys` to include multiple keys for the same value (e.g. `A` and `KeyA`). The reason being that, for example, when calling `getKeys()` it will return an array of the Keys values which for `A` is `KeyA`, but when trying to reference `Keyboard.Keys` using bracket notation with values returned from `getKeys()`, there would not be a `KeyA`.

I'm having an issue with _sandbox/tests/zoom/zoom.ts_ when checking for plus and minus keys.
```js
if (ev.key === 107 /* + */) {
  game.currentScene.camera.zoom((currentZoom += 0.03));
}
```

When changed to the following:

```js
if (ev.key === ex.Input.Keys.NumAdd /* + */) {
  game.currentScene.camera.zoom((currentZoom += 0.03));
}
```

The following pre-commit git error is raised and I'm not sure how to resolve this issue.

```The file does not match your project config: sandbox/tests/zoom/zoom.ts.```

eonarheim:
@rledford I'll take a look today, we probably have something odd in our commit hooks that needs to be addressed

kamranayub:
@rledford if needed, you can bypass pre-commit checks with `-n` e.g. `git commit -nm "my commit"`

rledford:
ok thanks

eonarheim:
Okay I'm pretty sure this is a misconfiguration of our eslint in regards to `./sandbox` code, the gist is eslint will not perform any linting unless the file is configured in in a tsconfig that eslint knows about explicitly (which it isn't unfortunately). We must not have noticed since we don't change that area of test code very frequently, especially with the move to storybook style tests.

@rledford Go ahead and bypass the pre-commits with the `-n` flag for these issues as @kamranayub suggested

rledford:
@eonarheim and @kamranayub Thanks for helping me get that sorted out. Looks like everything passed this time.

eonarheim (approved):

eonarheim:
Thanks for the hard work on this @rledford!

rledford:
Glad to help. I’ll see what else I can help with.

Issue #1363, opened 2019-12-26
Scene constructor documentation is confusing/inconsistent


eonarheim:
### Steps to Reproduce
https://codesandbox.io/s/excaliburjs-reproduction-01-pezj9

```typescript
import * as ex from 'excalibur'

const game = new ex.Engine({
  width: 800,
  height: 600,
  displayMode: ex.DisplayMode.FullScreen,
  pointerScope: ex.Input.PointerScope.Document,
})

const Scene1 = new ex.Scene()
const label = new ex.Label('Hello', 10, 10, '48px Arial')

Scene1.add(label)

game.addScene('Scene1', Scene1)
game.goToScene('Scene1')
game.start()
```

### Expected Result

The typings for new ex.Scene indicate that it requires an Engine object to be passed in, yet it actually does not. Why is this? I see that the argument is labeled in the Typescript idiom which indicates an unused variable (leading underscore).

### Actual Result

Engine is not actually required 

### Environment
- operating system: n/a
- Excalibur versions: master

### Current Workaround
Use ex.Scene with or without passing an Engine argument

### Suggested fix
Either update the documentation to disambiguate proper usage, or correct the typing on ex.Scene

PR #1677, opened 2020-10-04
fix: Make engine parameter optional for scene constructor

miqh:
Hello,

This minor change should address #1363.

As additional information, I noticed there aren't any usages within the codebase which invoke the parameter-less constructor for `Scene`. However, there are usages with the `Engine` argument supplied scattered throughout spec and test files. The usages of `new Scene()` appear to be confined to the documentation.

I also noticed there had been a change made in the past (#1068) which caused the `Engine` argument to be mandatory in the constructor signature to begin with. It's not clear to me exactly why that was done given the context available.

---

==================


eonarheim (approved):
@miqh Thanks for your contribution!

eonarheim:
@miqh I think that the change to make it mandatory introduced in #1068 was an oversight on our part.

PR #1679, opened 2020-10-04
fix: [#1687] Actor z-index regression

### LINKED ISSUE #888: Actor z-index throws an error if set on a child actor or on an Actor not part of a scene
This issue was brought up on the forum by Maxime, https://groups.google.com/forum/#!topic/excaliburjs/NPdFFSPDdwE

This may take some work to fix, probably by generalizing draw code from scene and extracting them into a separate class that can be used both by scenes, child actors, and any other actor based data structrue meant for drawing.

### Steps to Reproduce
Repro codepen https://codepen.io/anon/pen/YrVwOe?editors=0010#0

image

Setting the z-index on a child actor does not work because the scene drawTree relies on a scene to order drawing, but child actors do not have a scene they are a member of.

Excalibur Actor.ts line 620
```typescript

   /**
    * Sets the z-index of an actor and updates it in the drawing list for the scene. 
    * The z-index determines the relative order an actor is drawn in.
    * Actors with a higher z-index are drawn on top of actors with a lower z-index
    * @param newIndex new z-index to assign
    */
   public setZIndex(newIndex: number) {
      this.scene.cleanupDrawTree(this);
      this._zIndex = newIndex;
      this.scene.updateDrawTree(this);
   }
```

### Expected Result
Excalibur should allow z-index operations in child actors

### Actual Result
Errors are thrown and the game does not load.

### Environment
- browsers and versions: Chrome 60+
- operating system: Windows 10
- Excalibur versions: v0.12.0

### LINKED ISSUE #1678: Z-Index is broken since v0.24.5
### Steps to Reproduce

1. Create some actors.
2. Set a different z-index on each actor.
3. Position the actors to overlap.

### Expected Result

Actors with a higher z-index should be displayed in front of actors with a lower z-index.

### Actual Result

Actors are displayed in no particular order.

### Environment

- browsers and versions: Firefox 80
- operating system: Windows 10
- Excalibur versions: 0.24.5

The problem is commit 855ce516.

In this commit calls to `SortedList.getComparable` have been changed from `getComparable.call(element)` to `getComparable(element)`. In other words, `element` was previously passed as `this`, but now it is passed as an argument. However the actual implementation passed as `getComparable` still expects `element` to be passed as `this`.

eonarheim:

==================

Closes #1678 
Also Closes #888 

## Changes:

- Fix z-index regression in sorted list
- Add safety around z-index setting that fixes #888
- Adds a unit test to ensure this isn't broken in the future


kamranayub (approved):
Lgtm!

PR #1707, opened 2020-11-10
refactor: [#994] Switch to browser promises

### LINKED ISSUE #994: Replace Excalibur promise implementation with browser implementation

eonarheim:

==================

Closes #994

## Changes:

- Retarget to es2015 from es5 for browser promises (without polyfill)
- Replace Excalibur `ex.Promise` with built in browser promise
- Refactor loader
- Fix tests, subtle changes needed to be made because callbacks are now on the microtask queue


src/spec/PromiseSpec.ts
```diff
@@ -1,6 +1,6 @@
 import * as ex from '@excalibur';
 
-describe('A promise', () => {
+xdescribe('A promise', () => {
```
jedeen, line 3: Were these tests failing as a result of the deprecation?

eonarheim: I'll figure this out before we merge, seems to be something related to the `@obsolete()` some sort of dependency issue

src/engine/Drawing/Sprite.ts
```diff
@@ -51,7 +51,7 @@ export class SpriteImpl implements Drawable {
   private _spriteCtx: CanvasRenderingContext2D = null;
   private _pixelData: ImageData = null;
   private _pixelsLoaded: boolean = false;
-  private _dirtyEffect: boolean = false;
+  private _dirtyEffect: boolean = true;
```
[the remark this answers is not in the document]
eonarheim: I'll look at this a little closer, but that is what it means. Perhaps there is a better way to handle the new load semantics and initializing the sprite for first use

src/spec/PromiseSpec.ts
eonarheim: @jedeen Thanks for asking about this, when retargeting to es2015 classes the decorators no longer worked!!

src/engine/Drawing/Sprite.ts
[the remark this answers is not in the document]
eonarheim: I think this change is desirable, in the past the Excalibur promise implementation was synchronous and would immediately flip this to true at construction time. now that there is a microtask from the browser promise this state should start at true otherwise the initial pixel load will never happen.

https://github.com/excaliburjs/Excalibur/pull/1707/files/e161b5d92043a575dd3506213fe8089ab6919545#diff-40fef6d9412ae9640b0d53676193a416542b9b1c6df55ce3ccf7d0ba64b5f4a1L92

PR #1752, opened 2021-01-02
chore: Switch to GH actions

eonarheim:

==================

## Changes:

- Switch to GH actions from Travis


kamranayub:
@eonarheim I see this in typedoc output:

> Using tsc v3.8.3

This was an issue before, I had written a small script (maybe in Travis?) to remove typescript or node_modules from the typedoc-default-theme directory or something like that. We will need that to fix this error, so it uses the correct TSC from the top level.

kamranayub:
Done!

kamranayub (approved):

PR #1779, opened 2021-02-03
feat: Add support for custom SpriteSheetSpacingDimensions

### LINKED ISSUE #1778: enhancement: Custom spritesheet spacing dimensions
### Context

Originally `spacing` was added in #1058

The issue is that for some sprites like ones from https://kenny.nl, the spacing is not on the left, right, top or bottom of the image, it's in-between sprites so SpriteSheet doesn't properly splice it up.

### Proposal

Use overloaded setter:

```js
spriteSheet.spacing = 1; // current behavior
spriteSheet.spacing = { // SpritesheetSpacing
  left: 0,
  right: 0,
  top: 0,
  bottom: 0,
  margin: 1
};
```

kamranayub:

==================

Closes #1778 

## Changes:

- **BREAKING CHANGE:** Make `SpriteSheet` immutable to reduce chance of state errors
  - Made public fields private and added getters
- Update `spacing` to accept a `SpriteSheetSpacingDimensions` structure
- Add `getSpacingDimensions` public method
- Add `spacing` as optional argument to `SpriteSheet` constructor
- Update backing sprites for sheet when adjusting spacing after construction
- Remove `Configurable` from sheet/font after discussion with Erik


src/engine/Drawing/SpriteSheet.ts
```diff
@@ -84,12 +85,14 @@ export class SpriteSheetImpl {
...
+      const spacing = this.getSpacingDimensions();
+
+      for (let row = 0; row < this.rows; row++) {
+        for (let col = 0; col < this.columns; col++) {
+          this.sprites[col + row * this.columns] = new Sprite(
             this.image,
-            j * this.spWidth + this.spacing * j + this.spacing,
-            i * this.spHeight + this.spacing * i + this.spacing,
+            col * this.spWidth + spacing.margin * col + spacing.left,
+            row * this.spHeight + spacing.margin * row + spacing.top,
```
kamranayub, line 95: @eonarheim tests pass but check my math 😅

kamranayub:
I am seeing an issue when testing locally with Kenny.nl's spritesheet.

kamranayub:
Figured it out, `spacing` was a public member but didn't take effect because sprites array is built only during construction. Changed to getter/setter pattern.

src/engine/Drawing/SpriteSheet.ts
```diff
@@ -186,6 +213,25 @@ export class SpriteSheetImpl {
...
+
+  /**
+   * Builds sprites array with dimensions and spacing for loaded image
+   */
+  private _buildSpritesArray() {
+    const spacing = this.getSpacingDimensions();
+
+    for (let row = 0; row < this.rows; row++) {
+      for (let col = 0; col < this.columns; col++) {
+        this.sprites[col + row * this.columns] = new Sprite(
```
kamranayub, line 225: @eonarheim Any implications to leaving old Sprites around from the array?

eonarheim: I think this is fine, legacy sprites do have some canvas overhead internally but this shouldn't be on a hot path like rebuilding every frame or something.

eonarheim: In the future, zero potential problems at all, sprites will be a "view" into a source image

kamranayub: What about references to these sprites from the other public methods? Should it actually set the properties of the sprite instead of overwriting with new instance?

eonarheim: @kamranayub Perhaps, although I'm not sure I see a use case for modifying sprites underneath users. You mentioned on another thread, but SpriteSheet's should probably be immutable after creation

src/engine/Drawing/SpriteSheet.ts
```diff
@@ -34,28 +35,28 @@ export class SpriteSheetImpl {
     rows?: number,
     spWidth?: number,
     spHeight?: number,
-    spacing?: number
+    spacing?: number | SpriteSheetSpacingDimensions
   ) {
     let loadFromImage: boolean = false;
     if (imageOrConfigOrSprites instanceof Array) {
```
kamranayub, line 41: This ctor is something else. Candidate for rewrite/refactor later 😓

src/engine/Drawing/SpriteSheet.ts
```diff
@@ -34,28 +35,28 @@ export class SpriteSheetImpl {
...
+        this._columns = columns;
+        this._rows = rows;
+        this._spWidth = spWidth;
+        this._spHeight = spHeight;
+        this._spacing = spacing || 0;
       }
-      this.sprites = new Array(this.columns * this.rows);
+      this._sprites = new Array(this._columns * this._rows);
       loadFromImage = true;
     }
```
kamranayub, line 61: After this initialization routine, the rest of the code in ctor uses public getters for consistency.

src/engine/Drawing/SpriteSheet.ts
```diff
@@ -98,6 +101,75 @@ export class SpriteSheetImpl {
...
+   */
+  public get spacing() {
+    return this._spacing;
+  }
+
+  /**
+   * Get a copy of the backing sprite array
+   */
+  public get sprites() {
+    return [...this._sprites];
```
kamranayub, line 115: @eonarheim should I do this instead?

```suggestion
    return this._sprites.map(sprite => sprite.clone());
```

src/engine/Drawing/SpriteSheet.ts
```diff
@@ -222,7 +307,7 @@ export class SpriteSheet extends Configurable(SpriteSheetImpl) {
 export class SpriteFontImpl extends SpriteSheet {
   private _currentColor: Color = Color.Black;
   private _currentOpacity: Number = 1.0;
-  private _sprites: { [key: string]: Sprite } = {};
+  private _spriteRecord: Record<string, Sprite> = {};
```
kamranayub, line 310: 21st century TS 💯

kamranayub: We could consider a `Map` instead at some point depending on performance test.

eonarheim: One cool this with a map, is you aren't restricted to just string keys

eonarheim (approved):
Spoke offline, approved

Issue #1161, opened 2019-06-08
Tentative Excalibur v1.0.0 Roadmap


eonarheim:
## v1 Roadmap

Roadmap is subject to change and not 100% complete

### v0.23.0 - Collision Refactor and Enhancements
* [x] Collision Refactoring #1119 
* [x] Within/closestLine implementation #1071 
* [x] Re-implement collision groups #862 #1091 
      * [Collision Group Sample Platformer](https://github.com/excaliburjs/sample-platformer)
* [x] Bugs, bugs, bugs

### v0.24.0 

* [ ] ECS foundations
* [ ] Drawing API redesign
   * [ ] Remove child actors in favor of heirarchical drawing
* [ ] Switch polyfills to CoreJS
* [ ] TypeScript
   * [ ] Switch to ESLint over TSLint
   * [ ] TypeScript 3.5
   * [ ] `strict: true`
* [ ] Build speed improvement
* [ ] Bugs, bugs, bugs

### v0.25.0

* [ ] ECS Everywhere
* [ ] Action Redo async/await
* [ ] Multi-contact collisions
* [ ] Incremental loading
* [ ] Shared vs Scene scoped resources to save memory
* [ ] Bugs, bugs, bugs

### v0.26.0

* [ ] API standardization
* [ ] Style guide finalization
* [ ] Game state management strategy
* [ ] Bugs, bugs, bugs

### v0.27.0
* [ ] Bugs, bugs, bugs

PR #1786, opened 2021-02-21
feat: Implement Child Entities w/ Collision Support

eonarheim:

==================

This change changes the Entity `TransformComponent` to be backed by an affine matrix that is aware of parent transforms. Child entities are added to the same ECS world as the every other entities, this allows children to behave like top level entities without custom code.

child-collision

~~There is a breaking change with `Vector` making x/y readonly, this was made because mutating those values on vector really is a foot gun if you are not careful causing bugs that are very hard to find. This is probably a controversial change, and could maybe be separated into a separate feature/PR? Open to thoughts here, one option to easy into this is to not take a hard-line `readonly` just yet.~~ After further discussion, we've decided this change is too drastic, instead we've created a `VectorView` that present a `Vector` interface to code but be backed by other data, like an affine matrix.

Related to child actors on roadmap https://github.com/excaliburjs/Excalibur/issues/1161

## Changes:

- `VectorView` type that can create a Vector compatible type 
- ECS Entity update to add parent/child relationships, it is up to components to interpret what that means
- ECS Entity update to add `entity.get(TypeOfComponent)` for strongly typed component access
- Matrix based transform component
- Child Entities exist in the ECS world


src/engine/EntityComponentSystem/Components/TransformComponent.ts
```diff
@@ -20,15 +74,110 @@ export enum CoordPlane {
...
+      setY: (source, y) => {
+        if (this.parent) {
+          const [, newY ] = this.parent?.getGlobalMatrix().getAffineInverse().multv([source.data[12], y]);
+          this.matrix.data[13] = newY;
+        } else {
+          this.matrix.data[13] = y;
+        }
+      }
+    });
+  }
```
kamranayub, line 171: Yeah, with these numbers appearing all over, they should be made into constants.

src/engine/Math/vector-view.ts
```diff
@@ -0,0 +1,39 @@
...
+
+
+export interface VectorViewOptions<T> {
+  data: T;
+  getX: (source: T) => number;
+  getY: (source: T) => number;
+  setX: (source: T, x: number) => void;
+  setY: (source: T, y: number) => void;
+}
+export class VectorView<T> extends Vector {
```
kamranayub, line 11: I don't think I understand the purpose of having to pass a data source. When you create a VectorView, implementing the getters and setters will let you close over any variables you'd need to access anyway.

If this is to help with closures, have you measured the impact? Otherwise I'd suggest keeping it simpler and not having to pass a data source (and therefore, don't need first arg of any of these methods).

src/engine/EntityComponentSystem/Components/TransformComponent.ts
kamranayub: Would it make sense to add getters for whatever this data array/value is? Should something outside the matrix be mutating it? It might be safer in the long run to get/set so consuming code doesn't have to know where `12` or `13` are in the data.

src/engine/Math/vector-view.ts
eonarheim: Ah true, I could definitely close over the source. I'll give that a shot :+1:

jedeen (approved):

PR #1818, opened 2021-03-19
feat: ECS based Motion + Multiple Collider Support + Improved Collision Simulation

### LINKED ISSUE #626: TileMaps should participate in Rigid Body Physics
### Context

TileMap collision is currently implemented as an actor trait, which seems like undue coupling between the Actor and TileMap. It should be folded into the new collision system with a collision area implementation, since TileMaps fit the definition of a collision area 

>  A collision area is a region of space that can detect when other collision areas intersect for the purposes of colliding 2 objects in excalibur.
### Proposal

* Carve a TileMap into a set of convex polygons and create a set of collision areas for them (our collision algorithm SAT only works on convex shapes)
    https://github.com/excaliburjs/Excalibur/blob/master/src/engine/Collision/PolygonArea.ts 
* Refactor Body to not be coupled with actor, but to instead allow interfaces around these properties so that physics Body can be applied to anything
* Refactor Body to allow multiple CollisionAreas at a time or make a meta CollisionArea that is composed other collision areas (THIS WOULD BE A HUGE WIN), this will require a change in CollisionContact & DynamicTreeCollisionBroadphase to support treating multiple areas as 1 collidable object
* If CollisionPairs and CollisonContacts are generated correctly, it should just work.

### LINKED ISSUE #664: Add PolyEdge physics body
### Context

In my desert golf clone, I create a series of points representing the terrain. I then manually create an Actor for each line segment and use edge collisions. This makes it tough because I need to subscribe to collision events for each edge _and_ remove each edge when I'm done with them.

It would be great to create a single actor but pass in an array of points representing line segments to create a poly edge.
### Proposal
``` ts
this.body.usePolyEdgeCollision(points, center?);
```

### LINKED ISSUE #712: Actors sometimes fall through edges that they collide with when stacked high
### Steps to Reproduce
1. Open the fastphysics visual test
2. Shoot rockets rapidly in a direction
3. Additionally, let rockets pile up in the box

### Expected Result
Actors should not pass through the edges they collide with.

### Actual Result
When enough Actors pile up, some of them fall through the box.
fastmovingphysics_1

### Environment

- browser: Google Chrome
- browser version: 54.0.2840.99 m
- operating system: Microsoft Windows 10

### LINKED ISSUE #1070: RigidBody Collision allows large masses to be pushed around by small mass collisions
### Steps to Reproduce
Click to move red block to the right
https://codepen.io/eonarheim/pen/MzLOeX?editors=0010 

I believe this is because we split MTV evenly which causes this odd behavior

https://github.com/excaliburjs/Excalibur/blob/e8239816c6e0c9a3f53022174746983821be806a/src/engine/Collision/CollisionContact.ts#L180-L183
image.png

### Expected Result
Large mass objects should not be pushed around by low mass objects when masses are disparate

### Actual Result
Because of mtv spliting they can be pushed 1 pixel or so at a time

### Environment
- Excalibur versions: master

### Current Workaround
Set massive objects to be collisionType Fixed to prevent motion

or

Use collisionType Passive and handle resolution manually in a precollision event handler.
See diagram
https://excaliburjs.com/docs/api/edge/classes/_physics_.physics.html#enabled

### LINKED ISSUE #1072: Infinite collisionstart called with physics when objects are level
Edit:Originally said this was only occurring when using rigid body physics, but even without it this still occurs. So when using any physics with simulated gravity(acc), having an object fall on 2 level objects results in continual collisionstart calls.

### Steps to Reproduce
Codepen example: https://codepen.io/anon/pen/eQxwmM?editors=0011

### Expected Result
2 collision start events are called one for each of the blocks that the falling falling has hit

### Actual Result
Collision events are continually called for each block

### LINKED ISSUE #1093: Implement Multi-Contact Style Collision Resolution
### Context
Currently Excalibur uses only single-contact style collision resolution, this comes with known stability problems such as collision oscillation yeilding confusing results (#1072 ) or large amounts of wobbling. This makes it difficult to use excalibur rigid body collision to do any kind of stacking game.

### Proposal
Implement multi-contact using a single contact memory strategy
   * Remember previous single contacts for a period of X frames and treat them as additional contact points if certain criteria are met
       - Resolution motion was small
       - Contact points are recent in time (allow for adjustability)
       - Shapes have a new contact point this frame

### LINKED ISSUE #1094: Implement Rigid Body sleeping in Collision System
### Context
Currently no rigid bodies in Excalibur sleep when collision resolution motion is very small. This wastes a lot of cycles computing collision information.

### Proposal
* Implement rigid body sleeping in excalibur
  - Have a way to prevent objects from sleeping for certain games
  - Implement collision resolution chaining, where the order of resolution is determined to produce the most stable output, rather than have it be random
  - If an body has had little motion (below a certain threshold constant) over X number of frames put it to sleep
  - Sleeping bodies will participate in the broadphase or narrowphase
  - Other bodies will wake up sleeping bodies if in their broadphase sweep detect a possible collision and if a contact is generated in their narrowphase
  - Sleeping bodies must wake if in a stack and resolve the collision in an ordered chain

eonarheim:

==================

This PR moves both the integration step to an ECS `MotionSystem` and the collision logic to an ECS `CollisionSystem`.

In addition this PR adds the following:
* Breaking: Colliders are no longer part of the Body component
* Multiple `Collider`s per `Entity` are now supported with the `CompositeCollider`
* Introduces `ColliderComponent` to hold Entity colliders and refactors colliders and the collision system
* `TileMap` now use built in `Collider`s instead of custom collision code
   * `TileMap`s collapse tiles into bigger convex shapes (Reverse tessellation?)
* Breaking: Renames Box and RigidBody colllsion resolution types to Arcade and Realistic respectively
* Improved collision simulation
   * Rigid body sleeping for improved performance
   * Multiple contact points possible now
   * Iterative solver for improved stability ([see article on the subject](https://erikonarheim.com/posts/understanding-collision-constraint-solvers/))
* Refactor Engine/Scene add/remove to simplify make use of new ECS World

Closes #1818
Closes #1094
Closes #1093
Closes #1072
Closes #1070
Closes #712
Closes #626
Closes #664


DrSensor:
​

eonarheim:
@excaliburjs/core-contributors I think I'm ready to call this ready, there are a few debug related items I'll address in a separate PR.

PR #1819, opened 2021-03-27
fix: Offscreen Culling Events in new graphics system

eonarheim:

==================

Offscreen culling events were not firing in the new graphics system, this updates entities with an "offscreen" tag if they are offscreen

## Changes:

- "offscreen" tag
- enter and exit viewport events fire once more for the new graphics system
- Bug in ECS component removal :boom: where entities were being removed from queries they still qualified for

### DIFF
--- a/src/engine/Camera.ts
+++ b/src/engine/Camera.ts
@@ -246,7 +246,16 @@ export class Camera extends Class implements CanUpdate, CanInitialize {
   /**
    * Get or set current zoom of the camera, defaults to 1
    */
-  public z: number = 1;
+  private _z = 1;
+  public get z(): number {
+    return this._z;
+  }
+
+  public set z(val: number) {
+    this._z = val;
+    this._halfWidth = this._engine.halfDrawWidth;
+    this._halfHeight = this._engine.halfDrawHeight;
+  }
   /**
    * Get or set rate of change in zoom, defaults to 0
    */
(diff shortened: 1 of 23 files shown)


src/engine/Camera.ts
```diff
@@ -246,7 +246,18 @@ export class Camera extends Class implements CanUpdate, CanInitialize {
   /**
    * Get or set current zoom of the camera, defaults to 1
    */
-  public z: number = 1;
+  private _z = 1;
+  public get z(): number {
```
jedeen, line 250: Would it be more readable to name this `zoom()`?

src/engine/EntityComponentSystem/Entity.ts
```diff
@@ -121,6 +121,15 @@ export class Entity<KnownComponents extends Component = never> extends Class imp
     return this.tags.includes(tag);
   }
 
+  /**
+   * Adds a tag to an entitie
```
jedeen, line 125:
```suggestion
   * Adds a tag to an entity
```

src/engine/EntityComponentSystem/Entity.ts
```diff
@@ -396,6 +405,8 @@ export class Entity<KnownComponents extends Component = never> extends Class imp
 
   /**
    * Get a component by type with typecheck
+   *
+   * (Does not work on tag components use .hasTag("mytag"))
```
jedeen, line 409:
```suggestion
   * (Does not work on tag components, use .hasTag("mytag") instead)
```

PR #1844, opened 2021-04-07
fix: [#874] Improve confusing behaviour of Actor.opacity

### LINKED ISSUE #874: Actor should not have its own opacity property
### Context
`Actor` should not have its own opacity property. Currently, it's used to facilitate the `Fade` Action, and to override the opacity of the Actor's `color` property (if it has one). This has the potential to cause confusion, over which property is actually used (see #837).

### Proposal
`Actor.opacity` should be deprecated in favor of using the opacity of `Actor.color.a` or `Actor.currentDrawing.opacity` (if the `IDrawable` has an opacity property, such as if it is a `Sprite`)

djcsdy:

- [x] :microscope: existing tests still pass **(after being updated for new behaviour)**

==================

Closes #874 (maybe)

## Changes:

This is a proposal to:

 * Make the behaviour of `Actor.opacity` more logical, so they behaves as developers would probably expect.
   * `Actor.opacity` and `Actor.color.a` are now independent values and both are honoured (wherever relevant) by multiplying them to determine the overall opacity.
   * Corresponding changes to the `Drawable` types (`Sprite`, `Animation`, `Polygon`) that multiply any `opacity` passed to `draw` as an option with the `opacity` field of the object.
 * Deprecate `Actor.opacity` in favour of `GraphicsComponent.opacity`, which already behaves in a sensible way.
 * Delete unused `Actor.previousOpacity` field.
 * Fix `Actor` turning "permanently" invisible after setting `Actor.opacity = 0`.

This changeset is technically backwards incompatible because it removes the interaction between `Actor.color.a` and `Actor.opacity`. Instead, `Actor.color.a` and `Actor.opacity` are treated as independent values and both are honored by multiplying them. I think this should maintain compatibility with any sensible usage of the API, but it's perfectly possible that there are some games that rely on the old behaviour.

Some examples of behaviour (that I would consider dubious) that would be incompatible with this change:

 * Constructing an `Actor` with an initial color and then expecting to be able to read that color's alpha value back by reading `Actor.opacity`.
 * Setting `Actor.opacity` and then expecting to be able to read that value back by reading `Actor.color.a`.
 * Setting `Actor.opacity = 0`, waiting at least one frame, and then expecting that Actor to stay invisible even if the opacity is subsequently set to a value > 0.
 * Setting the `opacity` field of a `Drawable` to a value other than `1` or `undefined` and then expecting that value to be overridden by `opacity` specified as an argument to `draw` (instead they will be multiplied).

I think any other usage of `Actor.color`, `Actor.opacity`, or `Drawable.opacity` should continue to work as before.

There is precedent for making this backwards incompatible change in that a similar change has already been applied to `Label`.

This changeset also removes an optimisation which would skip the `Actor.draw` call altogether for Actors with `opacity === 0`, but it's hard to retain that optimisation without also keeping confusing behaviour. I think it's a fair compromise to remove this optimization and deprecate `Actor.opacity` in favor of `GraphicsComponent.opacity`

It's not quite ready to merge (see incomplete checklist items above), but I'm submitting it as a PR early for discussion. I'll address the remaining items if Excalibur developers are broadly happy with the thrust of this changeset.

### DIFF
--- a/src/engine/Actor.ts
+++ b/src/engine/Actor.ts
@@ -327,7 +327,13 @@ export class ActorImpl extends Entity implements Actionable, Eventable, PointerE
   /**
    * The opacity of an actor. Passing in a color in the [[constructor]] will use the
    * color's opacity.
+   *
+   * @obsolete Actor.opacity will be removed in v0.26.0, use [[GraphicsComponent.opacity|Actor.graphics.opacity]].
    */
+  @obsolete({
+    message: 'Actor.opacity will be removed in v0.26.0',
+    alternateMethod: 'Use Actor.graphics.opacity'
+  })
   public opacity: number = 1;
 
   /**
--- a/src/engine/Actor.ts
+++ b/src/engine/Actor.ts
@@ -1114,10 +1114,6 @@ export class ActorImpl extends Entity implements Actionable, Eventable, PointerE
       this.color.a = this.opacity;
     }
 
-    if (this.opacity === 0) {
-      this.visible = false;
-    }
-
     // capture old transform
     this.body.captureOldTransform();
 
--- a/src/engine/Actor.ts
+++ b/src/engine/Actor.ts
@@ -85,9 +85,7 @@ export interface ActorDefaults {
  * @hidden
  */
 
-export class ActorImpl
-  extends Entity
-  implements Actionable, Eventable, PointerEvents, CanInitialize, CanUpdate, CanDraw, CanBeKilled {
+export class ActorImpl extends Entity implements Actionable, Eventable, PointerEvents, CanInitialize, CanUpdate, CanDraw, CanBeKilled {
   // #region Properties
 
   /**
@@ -331,7 +329,6 @@ export class ActorImpl
    * color's opacity.
    */
   public opacity: number = 1;
-  public previousOpacity: number = 1;
 
   /**
    * Direct access to the actor's [[ActionQueue]]. Useful if you are building custom actions.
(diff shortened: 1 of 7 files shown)


eonarheim:
@djcsdy Thanks for digging in to this issue, I'm 100% behind this change and your reasoning. This will be a great improvement.

@kamranayub Anything else to add?

kamranayub:
I will carve out some time to take a look this week but on the face of it I love everything about this 💪

djcsdy:
Thanks @eonarheim @kamranayub!

Based on the reception I'm going to finish this PR off - fix the tests, make sure everything still works, and finish the other checklist items. So you might want to hold off reviewing this change until I'm done, but don't let me stop you if you want to :-).

kamranayub:
I believe we have a team meeting today so we'll have some eyes on it I think and can leave any feedback 👍

djcsdy:
I've changed `Label.opacity` to be a property accessor so that it can set `text.opacity` internally. Without that change, `Label.opacity` doesn't do anything, which I felt was confusing (especially since `Label.opacity` did work in Excalibur 0.24.x).

However in order to do that I had to make `Actor.opacity` and `Particles.opacity` be property accessors too, because TypeScript 4.x+ doesn't allow property accessors and fields to override each other.

I'm not overly keen on this aspect of the change, so if anyone has any better ideas please let me know.

djcsdy:
AFAIC this change is more or less done and I'm just waiting for feedback. Anyone have any comments?

eonarheim:
@djcsdy I think the property accessors make sense in this case, I think this is a solid change.

note: looks like coveralls was struggling at the time to the tests do pass

eonarheim:
My squase/merge chops need work locally, closed in 3ec9698fe6694d3e42bec95e7aac6cf363fcef3a

Thanks for the fix @djcsdy

PR #1849, opened 2021-04-26
fix: [#1848] Add polyfill for decodeAudioData on Safari

### LINKED ISSUE #1848: Safari (on macos and iOS) fails to load audio resources
### Steps to Reproduce

In Safari (on iOS or macos) attempt to run the sandbox project, or any project that loads audio resources.

### Expected Result

The project should load and audio should play as normal.

### Actual Result

Excalibur's loader fails to complete and prints the below to the console:

```
Unable to decode this browser may not fully support this format, or the file may be corrupt, if this is an mp3 try removing id3 tags and album art from the file.
```

My initial attempt to use third party tools to remove the metadata from mp3 files as mentioned was not successful and produced the same error.

### Environment

- browsers and versions: Safari 14.0.2
- operating system:  macos
- Excalibur versions: main branch

alanag13:

- [X] :see_no_evil: code conforms to the [style guide](https://github.com/excaliburjs/Excalibur/blob/main/STYLEGUIDE.md)

==================

Closes #1848

## Changes:

`webkitAudioContext.decodeAudioData` does not support the newer Promise-style API offered by the standard and instead uses the older callback-based method:

https://stackoverflow.com/a/52536753

This PR adds a polyfill that emulates the Promise-style API in webkit browsers.


eonarheim (approved):
Thanks @alanag13, we spoke offline about this polyfill and confirmed fix

PR #2051, opened 2021-10-01
Adding property to get screen center

### LINKED ISSUE #2045: Add `.center` getter to Screen
### Context

A common operation that could be made quicker is a helper getter to get the center of the screen. Currently folks need to write
```typescript
const center = vec(game.halfDrawWidth, game.halfDrawHeight);
```

### Proposal

1. Add a new getter `center` to Screen that will return a `Vector` `vec(this.halfDrawWidth, this.halfDrawHeight)`;
2. New unit tests are required

SirPedr:

- [X] :see_no_evil: code conforms to the [style guide](https://github.com/excaliburjs/Excalibur/blob/main/STYLEGUIDE.md)
- [X] :triangular_ruler: new tests written and passing / old tests updated with new scenario(s)

==================
There's something I noticed while writing tests and don't know if I'm missing something or not. The properties `drawWidth` and `drawHeight` use `scaledWidth` and `scaleHeight` internally, and then multiply that by `pixelRatio`. The thing is that `scaledWidth` and `scaleHeight` are defined by `resolution.width/pixelRatio` and `resolution.height/pixelRatio` respectively. Therefore, wouldn't this cause `pixelRatio` to be ignored? Or am I not addressing something? 

Closes #2045

## Changes:

- Added new property to class `Screen` that encapsulates screen center coordinates calculations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- New property `center` to `Screen` to encapsulate screen center coordinates calculation considering zoom and device pixel ratio
 - New `ex.Shape.Capsule(width, height)` helper for defining capsule colliders, these are useful for ramps or jagged floor colliders.
 - New collision group constructor argument added to Actor`new Actor({collisionGroup: collisionGroup})`
 - `SpriteSheet.getSprite(x, y)` can retrieve a sprite from the SpriteSheet by x and y coordinate. For example, `getSprite(0, 0)` returns the top left sprite in the sheet.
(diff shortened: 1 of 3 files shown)


eonarheim:
@SirPedr Good question and observation about the `pixelRatio`! You are right, they do cancel each other out. 

It definitely looks like a simplification is possible using the screen `resolution` directly in the `drawWidth`/`drawHeight`. I'd support such a simplification in a separate PR 👍 I'll keep thinking if there was a good reason for the way the code is, but I think it may have been a product of different abstractions evolving over time and the simplification was missed originally.

Additional context perhaps unrelated: Some of the reason for some of the complexity in Screen is to have the canvas not look blurry on HiDPI screens when rendering. The internal canvas bitmap needs to be scaled up by pixel ratio, and then scaled back down to match the pixel density of the screen. The reasoning behind this is explained well here https://www.html5rocks.com/en/mobile/high-dpi/

eonarheim (approved):
@SirPedr well done! Fantastic tests

PR #2317, opened 2022-05-16
[#396] Add ability to select variable duration

### LINKED ISSUE #396: Add random timer range as a feature of timers

malitherl:

==================

Closes #396

## Changes:

-This would set a random interval after receiving a range  of numbers


eonarheim (changes requested):
@malitherl Thank you very much for the contribution!  

I want a couple of changes before accepting:
* Implement a new random interval for each invocation of the timer
* Add the ability to pass a `Random` instance into `TimerOptions`
* Update the tests to test the above

eonarheim:
@malitherl Let me know if you need any assistance with the requested changes 👍

malitherl:
> @malitherl Let me know if you need any assistance with the requested changes 👍

Hi there! I have made some changes to allow the user to pass ex.Random into TimerOptions, continue to produce random intervals after reset, but I am having some difficulty with allowing the user to change the interval on reset() and also produce Random intervals

eonarheim:
Hi @malitherl 

> Hi there! I have made some changes to allow the user to pass ex.Random into TimerOptions, continue to produce random intervals after reset, but I am having some difficulty with allowing the user to change the interval on reset() and also produce Random intervals. Any insight would be incredibly helpful, thank you

Is the issue that upon `reset(<newInterval>)` the random intervals return back to the `orignalInterval + randomRange`? 

If so I think it's because the original `interval` is closed over in this arrow function.

```typescript
this.on(()=> this.interval = interval + this.random.integer(this.randomRange[0], this.randomRange[1]))
//                              ^ JavaScript will hold onto this value from the time the () => {} was created
```

One way around that is to create a new member variable to story the "base interval" set by the constructor or the `reset()`

```typescript
private _baseInterval = 10;
private _generateRandomInterval = () => {
  return this._baseInterval + this.random.integer(this.randomRange[0], this.randomRange[1])
}

//...
// in constructor
this._callbacks = [];
this._baseInterval = this.interval = interval;
if (!!randomRange){
  if(randomRange[0] > randomRange[1]){
    throw new Error('min value must be lower than max value for range');
  }
  //We use the instance of ex.Random to generate the range
  this.random = random ?? new Random();
  this.randomRange = randomRange;

  this.interval = this._generateRandomInterval();
  this.on(() => {
    this.interval = this._generateRandomInterval();
  });
}
```

Then your `reset()` can alter the `this._baseInterval = newInterval;` 

That seems to work 👍 when I gave it a shot

image

src/engine/Timer.ts
```diff
@@ -127,7 +159,7 @@ export class Timer {
    */
   public reset(newInterval?: number, newNumberOfRepeats?: number) {
     if (!!newInterval && newInterval >= 0) {
-      this.interval = newInterval;
+        this.interval = newInterval;
```
eonarheim, line 162:
```suggestion
        this.interval = newInterval;
        this._baseInterval = newInterval;
```

eonarheim (commented):
Added some suggestions

src/engine/Timer.ts
```diff
@@ -59,14 +70,37 @@ export class Timer {
...
+        this.interval = interval + new Random().integer(randomRange[0], randomRange[1])
+      }
+      this.randomRange = randomRange;  
+    } else {
+      this.interval = interval;
+    }
 
+    this.repeats = repeats || this.repeats;
+    
     this._callbacks = [];
```
eonarheim, line 91:
```suggestion
    this._callbacks = [];
    this._baseInterval = this.interval = interval;
    if (!!randomRange){
      if(randomRange[0] > randomRange[1]){
        throw new Error('min value must be lower than max value for range');
      }
      //We use the instance of ex.Random to generate the range
      this.random = random ?? new Random();
      this.randomRange = randomRange;

      this.interval = this._generateRandomInterval();
      this.on(() => {
        this.interval = this._generateRandomInterval();
      });
    }
```

src/engine/Timer.ts
```diff
@@ -28,27 +33,33 @@ export class Timer {
   public interval: number = 10;
   public repeats: boolean = false;
   public maxNumberOfRepeats: number = -1;
-
+  public randomRange: [number, number] = [0,0];
+  public random: ex.Random;
```
eonarheim, line 37:
```suggestion
  public random: ex.Random;
  private _baseInterval = 10;
  private _generateRandomInterval = () => {
    return this._baseInterval + this.random.integer(this.randomRange[0], this.randomRange[1])
  }
```

src/engine/Timer.ts
```diff
@@ -59,14 +70,37 @@ export class Timer {
...
       this.on(fcn);
     }
+    if(!!randomRange){
+      if(!!random){
+        this.on(()=> this.interval = interval + this.random.integer(this.randomRange[0], this.randomRange[1]))
+      } else {
+        this.on(()=> this.interval = interval + new Random().integer(this.randomRange[0], this.randomRange[1]))
+      }
+
+    }
```
eonarheim, line 103:
```suggestion

```

eonarheim: This gets simplified and folded into the above suggestion

src/engine/Timer.ts
eonarheim: We can take advantage of `this.random = random ?? new Random()` (this means if random is not passed in use a new Random) and the `_generateRandomInterval()` helper to simplify a bit

malitherl:
Everything is working on my end too, and all tests are passing. Thank you again for the suggestions and help @eonarheim

PR #2345, opened 2022-06-07
fix: [#2344] [#2343] Arcade Solver sorts contacts by distance and ignores diverging contacts

### LINKED ISSUE #2343: Jumping against and over a wall sometimes causes a collision that resets velocity (Arcade Physics)
Kapture 2022-06-06 at 16 58 07

### Steps to Reproduce

The above gif can be reproduced using this [CodeSandbox example](https://codesandbox.io/s/excalibur-jump-along-wall-bug-xsym4w?file=/src/index.js). (CodeSandbox is finicky with capturing key events, best to open the preview in a separate window).

Try moving the player to the right against the wall by holding the right arrow key. While still holding right, hit up to jump. Sometimes the player will slide along & above the wall, but other times it will "snap" to the top of the wall.

### Expected Result

The player should slide along & above the wall

### Actual Result

The player snaps to the top of the wall instead and velocity is reset - sometimes.

### Environment

- browsers and versions: Chrome 102, Safari 15.5 
- operating system: macOS 12.4 
- Excalibur versions: 0.26, 0.25, didn't try anything earlier 
- Using a macbook pro with 120hz screen incase that has an impact on the physics tick rate

### Current Workaround

None that I know of. I'm still quite new to using Excalibur, but this movement logic was taken from the Sample Platformer, so let me know if there's a better way to be applying velocities/collisions that might remedy this.

### LINKED ISSUE #2344: Unexpected left/right side collisions when walking on ground made of consecutive objects (Arcade Physics)
When walking along a ground made of consecutive objects with colliders, sometimes a left/right collision will occur. 

I remember seeing an issue/discussion somewhere (I cannot find it again for the life of me) where someone mentioned it's best to have larger, single bodies rather than consecutive. However, I am using the [tiled plugin](https://github.com/excaliburjs/excalibur-tiled#excalibur-wiring) with a "solid" layer for the ground, and I think it is laying out the colliders in similar fashion, leading to this behaviour.

### Steps to Reproduce

See this [CodeSandbox example](https://codesandbox.io/s/excalibur-side-collision-while-on-ground-8wf42u?file=/src/index.js). Move character left/right with arrow keys, jump with up. 

Open the dev console and walk the player along the lower ground. It will occasionally log "Left" or "Right" when the respective collision occurs. The higher ground is a single, large collider, while the bottom floor is individual 16x16 blocks. The behaviour should only happen on the lower ground.

### Expected Result

It occasionally triggers a collision on the left/right side

### Actual Result

It should not trigger any left/side collisions

### Environment

- browsers and versions: Chrome 102, Safari 15.5 
- operating system: macOS 12.4 
- Excalibur versions: 0.26, 0.25, didn't try anything earlier 
- Using a macbook pro with 120hz screen incase that has an impact on the physics tick rate

eonarheim:

==================

Closes #2344 #2343

This PR changes how the `ArcadeSolver` solves collisions to avoid errant collisions on seams in geometry. The idea behind this fix is to sort the collision contacts by distance. Additionally if a contact has no more overlap during processing they are now discarded so errant collisions are even evaluated.

Note there are no more left right collisions in the console compared to the [sandbox](https://8wf42u.csb.app/)
sorted-collisions

This PR also fixes #2343 by ignoring "divergent contacts" which are when the underlying body's are already moving away from the contact, so no velocity adjustment is needed. This was noticeable when clipping corners, but the object was already moving out of the collision.

cancel-velocity-fix

TODO: 
* [x] Look into #2343 to see if this helps
* [x] Also sort by collider proximity
* [x] Tests

## Changes:

- Change the `Solver` from abstract class to an interface to be more flexible about solvers
- Sorts contacts by body distance from each other
- Cancels "divergent contacts" when the bodies are moving away from the contact

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fixed issue with `ex.ArcadeSolver` based collisions where colliders were catching on seams when sliding along a floor of multiple colliders. This was by sorting contacts by distance between bodies.
+  ![sorted-collisions](https://user-images.githubusercontent.com/612071/172401390-9e9c3490-3566-47bf-b258-6a7da86a3464.gif)
+
+- Fixed issue with `ex.ArcadeSolver` where corner contacts would zero out velocity even if the bodies were already moving away from the contact "divergent contacts".
+  ![cancel-velocity-fix](https://user-images.githubusercontent.com/612071/172500318-539f3a36-31ae-4efc-b6ab-c4524b297adb.gif)
+
 - Fixed issue where `ex.Sound` wasn't being paused when the browser window lost focus
 
 ### Updates
--- a/sandbox/tests/side-collision/index.html
+++ b/sandbox/tests/side-collision/index.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Side Collision</title>
+</head>
+<body>
+  <script src="../../lib/excalibur.js"></script>
+  <script src="index.js"></script>
+</body>
+</html>
--- a/sandbox/tests/side-collision/index.ts
+++ b/sandbox/tests/side-collision/index.ts
@@ -1,0 +1,89 @@
+
+var game = new ex.Engine({
+  width: 400,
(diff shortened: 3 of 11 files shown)


eonarheim:
@mattjennings Give the latest alpha `excalibur@0.27.0-alpha.477` a try! Let me know if you run into any other issues. Theoretically #2344 and #2343 should be fixed.

Turns out sorting contacts by distance was enough to sort out the issue generally, no configurable bias was needed 😄

mattjennings:
@eonarheim fantastic! Thanks a ton. I haven't had a chance to take a look yet but I super appreciate the quick and great work.

PR #2361, opened 2022-06-23
fix: [#2359] work around image decode limits in chromium

### LINKED ISSUE #2359: Chromium's Image#decode has limited calls
It's a Chromium specific issue. In Firefox & Safari works as expected.

[codesandbox](https://codesandbox.io/s/happy-gagarin-j1vjr) from [this](https://bugs.chromium.org/p/chromium/issues/detail?id=1055828#c7) issue on chromium.org
Another issue is [here](https://bugs.chromium.org/p/chromium/issues/detail?id=606319).

Possible solutions:
1. Add option to load images through DOM
2. Set up a limit for Image#decode promises (only 256 images could be decoded simultaneously)

I just check second solution by myself, and it worked like a charm.

Error:
`excalibur.js:2996 [Error] :  [Tiled] Error loading tileset images
tiled-map-resource.ts:392 Error loading ImageSource from path '/assets/clothes/dress.png' with error [The source image cannot be decoded.]`

Implemented in ImageSource: https://github.com/excaliburjs/Excalibur/blob/b285ede60652ff8273f78a37a1de2bc6a77db485/src/engine/Graphics/ImageSource.ts#L92

eonarheim:

==================

Closes #2359

This PR switches away from `Image.decode()` to use the older method of `Image.onload` this method is more reliable for large or numerous images due to a chromium bug around decode.

@HxShard provided a great codesandbox illustrating the issue https://codesandbox.io/s/happy-gagarin-j1vjr

## Changes:

- Creates a new Future type for working with native Promises' resolve/reject at any time
- Creates a new Semaphore type that can limit async calls in between `enter` and `exit`
- Fixes a small clock schedule bug as well

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,6 +108,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Add target element id to `ex.Screen.goFullScreen('some-element-id')` to influence the fullscreen element in the fullscreen browser API.
 
 ### Fixed
+- Fixed bug in `Clock.schedule` where callbacks would not fire at the correct time, this was because it was scheduling using browser time and not the clock's internal time.
 - Fixed issue in Chromium browsers where Excalibur crashes if more than 256 `Image.decode()` calls are happening in the same frame.
 - Fixed issue where `ex.EdgeCollider` were not working properly in `ex.CompositeCollider` for `ex.TileMap`'s
 - Fixed issue where `ex.BoundingBox` overlap return false due to floating point rounding error causing multiple collisions to be evaluated sometimes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,24 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 -
 
 ### Added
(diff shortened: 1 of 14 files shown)


HxShard:
@eonarheim I just did some checks and noticed one not very cool thing. In my case with this error, Image#decode was fired only 20 times.
I think, buffer size (or something similar) matters there too.

eonarheim:
@HxShard Good call out, I'll do some experimentation to see how the buffer size plays into this.

Roughly how large were your images? Dimensions and disk size?

HxShard:
@eonarheim https://gist.github.com/HxShard/376b82855f076e4f6e8b51fc40e85f7e

eonarheim:
@HxShard With some experimentation, looks like using `Image.decode()` will be problematic for just a handful of large textures in chrome, tuning down the semaphore super low to 2-3 "works" but really kills load speed.

I checked using the old school `Image.onload` and that seems to work (and fast) just fine without issue or image flickering in webgl, I think because the GPU needs to decode the image into whatever internal format it needs for graphics. I think `Image.decode()` may only be necessary for DOM related images... more research is required but experimentally seems to be the case. 

I'm going to explore removing the semaphore and switching to the `Image.onload` method instead 👍

PR #2381, opened 2022-06-30
feat: pass data to scene activation from goToScene

### LINKED ISSUE #1548: Implement Scene activation context and data passing
### Context

Thinking about usage of scenes as navigation structures, you go between scenes almost like a page router. Typically with navigation APIs, you can pass data (like a URL or route data). In Excalibur, when you call `Engine.goToScene(sceneKey)` it might be nice to pass data to the `Scene.onActivate` hook.

### Proposal

We'd need to figure out a way to indicate to TypeScript how to type this but something like this:

```ts
class Level extends Scene {
  onActivate(context) {
    console.log(context.data); // "overworld"
  }
}
game.add('level', new Level());
game.goToScene('level', 'overworld');
```

The data passed can be arbitrary, no reason to restrict it necessarily. By default, `undefined` is used.

```ts
interface SceneActivationContext<TData = undefined> {
  data: TData;
  previousScene: Scene;
  engine: Engine;
}
```

Maybe some type helpers on `Scene`?

```ts
class Level extends Scene<string> {
  onActivate(context /* type: SceneActivationContext<string> */) {
    console.log(context.data); // "overworld"
  }
}
```

Or can we override with optional generic?

```ts
class Scene {
  // ...
  public onActivate<TData = undefined>(context: SceneActivationContext<TData>) {
    // override me
  }
}

class Level extends Scene {
  onActivate<string>(context) {
    console.log(context.data);
                        ^ string
  }
```

mattjennings:

==================

Closes #1548

## Changes:

- Added optional `data` parameter to `goToScene`, which gets passed to the target scene's `onActivate` method.
  ```typescript
  class SceneA extends ex.Scene {
    /* ... */

    onActivate(context: ex.SceneActivationContext<{ foo: string }>) {
      console.log(context.data.foo); // bar
    }
  }

  engine.goToScene('sceneA', { foo: 'bar' })
  ```
- Scene `onActivate` and `onDeactivate` methods have been changed to receive a single parameter, an object containing the `previousScene`, `nextScene`, and optional `data` passed in from `goToScene()`

### DIFF
--- a/src/engine/Interfaces/LifecycleEvents.ts
+++ b/src/engine/Interfaces/LifecycleEvents.ts
@@ -88,11 +88,18 @@ export interface CanInitialize {
   off(eventName: Events.initialize, handler?: (event: Events.InitializeEvent<any>) => void): void;
 }
 
-export interface CanActivate {
+export interface SceneActivationContext<TData = undefined> {
+  data?: TData;
+  previousScene: Scene;
+  nextScene: Scene;
+  engine: Engine;
+}
+
+export interface CanActivate<TData = undefined> {
   /**
    * Overridable implementation
    */
-  onActivate(oldScene: Scene, newScene: Scene): void;
+  onActivate(context: SceneActivationContext<TData>): void;
 
   /**
    * Event signatures
@@ -106,7 +113,7 @@ export interface CanDeactivate {
   /**
    * Overridable implementation
    */
-  onDeactivate(oldScene: Scene, newScene: Scene): void;
+  onDeactivate(context: SceneActivationContext<never>): void;
 
   /**
    * Event signature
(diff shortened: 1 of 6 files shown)


src/engine/Interfaces/LifecycleEvents.ts
```diff
@@ -88,11 +88,18 @@ export interface CanInitialize {
   off(eventName: Events.initialize, handler?: (event: Events.InitializeEvent<any>) => void): void;
 }
 
-export interface CanActivate {
+export interface SceneActivationContext<TData = undefined> {
```
mattjennings, line 91: @eonarheim I named this TData as per the proposal in #1548, but I don't think we follow the `<TSomething>` naming convention with generics anywhere else (it seems to usually be `<T>`). I feel like either `<Data>` or `<T>` would be more consistent here.

src/engine/Interfaces/LifecycleEvents.ts
```diff
@@ -88,11 +88,18 @@ export interface CanInitialize {
...
+  data?: TData;
+  previousScene: Scene;
```
mattjennings, line 93: we could alternatively call this `prevScene` for convenience

eonarheim: `previousScene` is good, we try to avoid too many abreviations 👍

eonarheim (approved):
@mattjennings Thanks again for all the hard work! This will be a super useful addition!

PR #2393, opened 2022-07-01
fix: use generic on Scene for better onActivate data type

mattjennings:

==================

Addresses [follow up](https://github.com/excaliburjs/Excalibur/issues/1548#issuecomment-1172504724) in #1548 

## Changes:

- removes the generic in `onActivation` and adds it to the Scene instead. This allows it to be inferred by doing `onActivation(ctx: SceneActivationContext<Data>)` in Scene subclasses.


eonarheim:
@mattjennings Thanks for the update!

PR #2412, opened 2022-07-13
fix: [#2409] engine 'visible' event not firing

### LINKED ISSUE #2409: Engine VisibleEvent is not firing on visibility change
### Steps to Reproduce

This is reproducible with the [sample platformer](https://codesandbox.io/s/github/excaliburjs/sample-platformer)

- Start game
- Switch tabs or unfocus window
- Go back to game and it remains paused
- console log only shows "pause", no "start", which means the visible event never fired

### Expected Result

Engine should emit the "visible" event when window is focused

### Actual Result

It does not emit the "visible" event

### Environment

- browsers and versions: Chrome 103.0.5060.114 , FireFox 102.0.1, Safari 15.5
- operating system: macOS 12.4
- Excalibur versions: 0.27

### Current Workaround

I see that we have some browser-specific behaviour here https://github.com/excaliburjs/Excalibur/blob/88a30ceb7fd2986e2d13f4d2fa7c8be87a1def09/src/engine/Engine.ts#L1112-L1133 but I am currently doing this instead:

```js
document.addEventListener('visibilitychange', () => {
  if (document.visibilityState === 'hidden') {
    engine.stop()
  } else if (document.visibilityState === 'visible') {
    engine.start()
  }
})
```

which seems to have [good browser support](https://caniuse.com/mdn-api_document_visibilitystate)

mattjennings:
Fixes `engine.on('visible')` not firing by using `document.visibilityState` API to determine current state

==================

Closes #2409

## Changes:

- Fixed `engine.on('visible')` event not firing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
--
+- Fixed `engine.on('visible')` event not firing
 
 ### Updates
 
--- a/src/engine/Engine.ts
+++ b/src/engine/Engine.ts
@@ -1117,7 +1117,7 @@ O|===|* >________________>\n\
         this.eventDispatcher.emit('visible', new VisibleEvent(this));
         this._logger.debug('Window visible');
       }
-    })
+    });
 
     if (!this.canvasElementId && !options.canvasElement) {
       document.body.appendChild(this.canvas);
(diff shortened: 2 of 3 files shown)


eonarheim (changes requested):
@mattjennings Looks great!

Just needs some lint clean up

`npm run eslint -- --fix && npm run eslint:spec -- --fix`

eonarheim (approved):

eonarheim:
@mattjennings Thanks again for the contribution!

PR #2422, opened 2022-07-17
fix: [#2418] falsey event values being converted to GameEvent

### LINKED ISSUE #2418: Emitting an event with a value of 0 produces a GameEvent instead
When emitting a custom event on an object, if the value is 0 it gets overwritten by a GameEvent due to this line:

https://github.com/excaliburjs/Excalibur/blob/730d621805d32104c18dc7ca8d9e97c7f212bb2b/src/engine/EventDispatcher.ts#L36-L38

### Steps to Reproduce

Emit an event with a value of 0:

```js
actor.emit('something', 0)

actor.on('something', (value) => {
    console.log(value) // GameEvent
})

actor.emit('something', 1)

actor.on('something', (value) => {
    console.log(value) // 1
})
```

### Expected Result

Value to be persisted

### Actual Result

Value gets overwriten with GameEvent

### Environment

- Excalibur versions: 0.27

### Current Workaround

Create a custom GameEvent instance and emit that instead. Perhaps that is the intention, that every event should be an instance of GameEvent? If so, should `emit` throw an error if it receives something other than `undefined` or a `GameEvent`?

mattjennings:

==================

Closes #2418

## Changes:

- Fixed `EventDispatcher.emit` converting falsy values to `ex.GameEvent`. It will only convert `undefined` or `null` values now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fixed issue where `ex.Font` would become corrupted when re-used by multiple `ex.Text` instances
 - Fixed `engine.on('visible')` event not firing
+- Fixed `EventDispatcher.emit` converting falsy values to `ex.GameEvent`. It will only convert `undefined` or `null` values now.
 
 ### Updates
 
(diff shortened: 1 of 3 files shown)


eonarheim:
@mattjennings Thanks again for the contribution!

PR #2471, opened 2022-08-26
Fix x-axis bug in `actions.scaleTo`

### LINKED ISSUE #2470: `actions.scaleTo` never completes when scaling exclusively on the x-axis
Hello! :wave:

I'm pretty sure I found a bug in the `ScaleTo` action; the `isComplete` method checks `tx.scale.y` against the x-axis, rather than the `tx.scale.x` against the x-axis, and so the action will never complete if, for example, you're only scaling the x-axis. [The bug is on this line](https://github.com/excaliburjs/Excalibur/blob/fd2fa241da53558ebbebb300463d06ed465713a2/src/engine/Actions/Action/ScaleTo.ts#L64); I also cloned the repo, wrote a test, watched it fail, and then fixed the `isComplete` method; I'm opening this issue first since the PR checklist asks you to open an issue first. :smile:

### Steps to Reproduce

```javascript
actor
  .scaleTo(vec(2, 1), vec(1, 1))
  .scaleTo(vec(3, 1), vec(1, 1)); // This will never run
```

### Expected Result

The first scale would complete, and then the second scale would run and complete.

### Actual Result

The first scale executes, but never believes it's completed, and so the second scale never runs.

### Environment

- browsers and versions: Chrome, but this is a problem in any browser; the line of code that checks the x-axis for completion actually checks the y-axis.
- operating system: Windows
- Excalibur versions: main branch

### Current Workaround

Manually do the scaling yourself in an pre-/post-/`update` function instead of using the action.

reissbaker:
Previously if scaling only on the x-axis, the action would never complete.

- [x] :page_facing_up: changelog entry added (or not needed) __I think it's not needed for a small bugfix?__

==================

Closes #2470 

## Changes:

Fixes a bug I found in the `ScaleTo` action, where scaling on the x-axis exclusively may never complete. The `isComplete` method accidentally tested `tx.scale.y` against the x-axis, rather than testing `tx.scale.x` against the x-axis. I wrote the test first, watched it fail, and then fixed the bug and watched the test pass.

It looks like since I'm a first-time contributor I'll need approval from a maintainer to actually run the CI checks, but I pinky swear I ran `npm run all` locally and it all passed. :smile:

### DIFF
--- a/src/engine/Actions/Action/ScaleTo.ts
+++ b/src/engine/Actions/Action/ScaleTo.ts
@@ -61,7 +61,7 @@ export class ScaleTo implements Action {
   public isComplete(): boolean {
     return (
       this._stopped ||
-      (Math.abs(this._tx.scale.y - this._startX) >= (this._distanceX - 0.01) &&
+      (Math.abs(this._tx.scale.x - this._startX) >= (this._distanceX - 0.01) &&
         Math.abs(this._tx.scale.y - this._startY) >= (this._distanceY - 0.01))
     );
   }
--- a/src/spec/ActionSpec.ts
+++ b/src/spec/ActionSpec.ts
@@ -1051,6 +1051,19 @@ describe('Action', () => {
       expect(actor.scale.x).toBe(1.5);
       expect(actor.scale.y).toBe(1.5);
     });
+
+    it('completes when scaling on the x-axis', () => {
+      expect(actor.scale.x).toBe(1);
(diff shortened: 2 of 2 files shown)


eonarheim (approved):
@reissbaker Thanks for the contribution! This is fantastic!

reissbaker:
Thanks for the quick merge! :)

PR #2576, opened 2023-02-13
[#2575] fix actor clone

### LINKED ISSUE #2575: Weird behavior on actor clone
(Wasn't there a template for bug reports somewhere? I couldn't find it this time around.)

Excited as I was for the clone bug fix, I continued working on a side project of mine, but ran into some weirdness. After a lot of second guessing myself, I've reached the conclusion that the current method for actor cloning has a bug in it somewhere.

This example is being run on Excalibur v0.28.0-alpha.628+017953a.

## Minimal Working Example

```
import "regenerator-runtime/runtime";

import * as ex from 'excalibur';
import {vec, Actor, Color, CollisionType, Engine} from 'excalibur';

const game = new Engine({
  width: 800,
  height: 600,
});

game.start();

const paddle = new Actor({
x: 150, y:  40,
width: 200, height: 20,
color: Color.Chartreuse,
collisionType: CollisionType.Active,
});
game.add(paddle);

const paddle2 = paddle.clone();

paddle2.pos.x = 400;
paddle2.color = Color.Red;
paddle2.on('collisionstart', (ev)=>{console.log(ev);});
game.add(paddle2);

ex.Physics.useRealisticPhysics();
ex.Physics.gravity = vec(0,60);

const wall = new Actor({
x: 100, y:  400,
width: 200, height: 20,
color: Color.Black,
collisionType: CollisionType.Fixed
});
game.add(wall);
```

## Expected Behavior

One should see two falling physics objects, `paddle` and `paddle2`, and a `wall` below `paddle`. `paddle2` should be a copy of `paddle`, but with a different x coordinate and a different color. There is a `collisionstart` event assigned to `paddle2`, which should never occur because it never collides with `wall`.

## Actual Behavior

While `paddle2` does indeed have a distinct x coordinate from `paddle`, the change in color to `paddle2` also applies to `paddle`, which becomes red instead of chartreuse. Moreover, the `collisionstart` event is triggered when `paddle` collides with the wall.

## Probable Cause

The deep copy which should be happening when `paddle.clone()` is called is probably not deep copying some of the object's components.

eonarheim:

==================

Closes #2575 

## Changes:

- Update Graphics layer cloning to clone individual graphics
- Update Collider Component cloning to clone stored colliders

## TODO:

* [x] Investigate ColliderComponent cloning to determine eventing issue
* [x] Look into other components closely

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,9 @@ are returned
 
 ### Fixed
 
+- Fixed issue where `ex.ColliderComponent` was not deeply cloning the stored `ex.Collider` causing them to be shared across clones.
+- Fixed issue where `ex.GraphicsComponent` was not deeploy cloning the
+stored `ex.Graphics` causing them to be shared across clones.
 - Fixed issue where `Actor.clone()` and `Entity.clone()` crashed.
 - Fixed issue where zero mtv collisions cause erroneous precollision events to be fired in the `ArcadeSolver` and `RealisticSolver`
 - Fixed issue where calling `.kill()` on a child entity would not remove it from the parent `Entity`
--- a/sandbox/index.html
+++ b/sandbox/index.html
@@ -10,6 +10,7 @@
     <ul>
       <li><a href="html/index.html">Sandbox Platformer</a></li>
       <li><a href="tests/sprite-tint/">Sprite Tint</a></li>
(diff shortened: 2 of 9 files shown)


eonarheim:
Validated against the example repro in the bug 👍 

No logs appear where they shouldn't

clone-behavior

PR #2764, opened 2023-09-07
TileMap: Draw all tiles if screen coordinate plane is used

### LINKED ISSUE #2763: TileMap draw method does not support screen coordinate plane
Related to discussion https://github.com/excaliburjs/Excalibur/discussions/2761.

The current implementation of the draw method in the TileMap class only checks whether tiles are visible in the world coordinate plane. This means that if a developer uses the TileMap in the screen coordinate plane, the visibility calculations do not work correctly, potentially leading to tiles not being drawn when they should be.

### Steps to Reproduce

* Create a TileMap instance and populate it with tiles.
* Set up the tilemap such that the TileMap is used in the screen coordinate plane rather than the world coordinate plane, e.g. like this:

```ts
    // Set the tilemap coordinate plane to the screen
    let trans = tilemap.get(TransformComponent);
    if (trans) {
      trans.coordPlane = CoordPlane.Screen;
    }
```

* Notice that the draw method's visibility calculations are incorrect, leading to incorrect tile rendering.

### Expected Result

The draw method should correctly determine tile visibility and draw tiles correctly when the TileMap is used in the screen coordinate plane.

### Actual Result

The draw method only correctly determines tile visibility in the world coordinate plane, leading to incorrect behavior when used in the screen coordinate plane.

### Environment

- browsers and versions: Firefox (117.0)
- operating system: Fedora Linux 38
- Excalibur versions: v0.29.0-alpha.0+6bd8b51

### Possible Solution
Modify the draw method to include an option or mechanism to correctly handle tile visibility calculations in both the world and screen coordinate planes. This could potentially involve 
getting the TransformComponent`s coordPlane property to detecting the current coordinate plane.

A simpler and faster solution (which would currently be sufficient for me) would be to include a way to disable the overlap check. This would have the positive side effect that this could be used for small maps to improve performance. This is possibly also be a permanent solution, as in the screen coordinate plane everything should usually be in the visible area.

JumpLink:
This partially fixes #2763, by drawing all tiles if the tilemap uses the screen coordinate plane.

### DIFF
--- a/src/engine/TileMap/TileMap.ts
+++ b/src/engine/TileMap/TileMap.ts
@@ -17,7 +17,7 @@ import { DebugGraphicsComponent } from '../Graphics/DebugGraphicsComponent';
 import { Collider } from '../Collision/Colliders/Collider';
 import { PostDrawEvent, PostUpdateEvent, PreDrawEvent, PreUpdateEvent } from '../Events';
 import { EventEmitter, EventKey, Handler, Subscription } from '../EventEmitter';
-import { CoordPlane } from '../Math/coord-plane'
+import { CoordPlane } from '../Math/coord-plane';
 
 export interface TileMapOptions {
   /**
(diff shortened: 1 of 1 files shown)


eonarheim:
@JumpLink I think this is a good enough change for now, I'm planning on updating the data structure to improve TileMap's offscreen calculation performance anyways.

Quick note, build failed on some lint

JumpLink:
@eonarheim fixed 👍

eonarheim:
@JumpLink Thanks for the find! I have a couple things to include, but I'll cut a new patch this weekend

PR #2771, opened 2023-09-27
fix: [#2763] TileMap Improve perf and draw tiles using QuadTree

### LINKED ISSUE #2763: TileMap draw method does not support screen coordinate plane

eonarheim:

==================

Closes #2763

## Changes:

- Updates TileMap to use a QuadTree spatial data structure to identify the tiles that need to be drawn
- Accidental perf improvement! It is now much more efficient to locate the proper tiles
  - This is especially noticeable on FireFox with bigger tile maps (test this PR with npm link against https://github.com/excaliburjs/excalibur-tiled/)


eonarheim:
Need to update the TileMap debug draw test, I've changed it :) but otherwise looks good 👍

PR #2776, opened 2023-09-30
feat: Implement async onInitialize

eonarheim:

==================

This PR changes up the onInitialize to be async, this is currently and experiment to see if this is the direction we want to take.

This is inline with our roadmap desire to allow loading scene specific resources

We will hold this type of breaking change until v0.29.0


eonarheim:
Implemented in Transition api pr

PR #2790, opened 2023-10-10
feat: Excalibur Scene Transitions & Loaders

eonarheim:

==================

This PR  implements an API inspired by https://github.com/mattjennings/excalibur-router/ with permission from Matt.

The idea behind this refactoring is to add transitions between scenes and scene specific loaders!
* Add or remove scenes by constructor
* Add loaders by constructor
* New `DefaultLoader` type that allows for easier custom loader creation
* New `Transition` type for building custom transitions
* New scene lifecycle to allow scene specific resource loading
    * `onTransition(direction: "in" | "out") {...}`
    * `onPreLoad(loader: DefaultLoader) {...}`
* New async goto API that allows overriding loaders/transitions between scenes
* Scenes now can have `async onInitialize` and `async onActivate`!
* New scenes director API that allows upfront definition of scenes/transitions/loaders

View [this file](https://github.com/excaliburjs/Excalibur/blob/feat/async-init-ex-router/sandbox/tests/router/index.ts) for a current example of what the API looks like 

Defining scenes upfront
```typescript
const game = new ex.Engine({
  scenes: {
    scene1: {
      scene: scene1,
      transitions: {
        out: new ex.FadeInOut({duration: 1000, direction: 'out', color: ex.Color.Black}),
        in: new ex.FadeInOut({duration: 1000, direction: 'in'})
      }
    },
    scene2: {
      scene: scene2,
      loader: ex.DefaultLoader, // Constructor only option!
      transitions: {
        out: new ex.FadeInOut({duration: 1000, direction: 'out'}),
        in: new ex.FadeInOut({duration: 1000, direction: 'in', color: ex.Color.Black })
      }
    },
   scene3: ex.Scene // Constructor only option!
  } 
})

game.start('scene1',
{
  inTransition: new ex.FadeInOut({duration: 500, direction: 'in', color: ex.Color.ExcaliburBlue})
  loader: boot,
});
```

scene-transitions-loader


mattjennings:
@eonarheim would it make sense to define the routes at the engine definition, which could clean up some of the `game.start` API?

```ts
const game = new ex.Engine({
  backgroundColor: ex.Color.fromHex("#5fcde4"),
  width: 600,
  height: 400,
  routes: {
    level1: {
      scene: scene1,
      loader: new BaseLoader(),
      // grouped transitions here so `out` and `in` are more explicit
      transitions: {
         out: new ex.FadeInOut({duration: 1000, direction: 'up', color: ex.Color.Black}),
         in: new ex.FadeInOut({duration: 1000, direction: 'down'})
      }
    }
  }
})

game.start('level1', {
  // optional overrides for the route
  loader: boot
});

// goToScene would/could have the same api
game.goToScene('level2', {
     transitions: {
          out: new ex.FadeInOut({...}),
          in: new ex.FadeInOut({...})
     }
})
```

eonarheim:
I love this idea, 100% I think it belongs there now that you point it out. It definitely feels like setup type config, as opposed to start!

eonarheim:
@mattjennings many thanks!

mattjennings:
Spitballing other ideas:

```ts
const game = new ex.Engine({
  scenes: {
    level1: Level1Scene
  }
})

class Level1Scene extends Scene {
  loader = new BaseLoader()

  transitions() {
    return {
      out: new ex.FadeInOut({duration: 1000, direction: 'up', color: ex.Color.Black}),
      in: new ex.FadeInOut({duration: 1000, direction: 'down'})
    }
  }
}
```

Pros:

- Gets rid of `routes` terminology which is probably too web dev-y anyway. I was also concerned it might cause confusion of what a route is vs a scene
- Scenes are just registered as key/value properties on Engine instantiation (note: no `new Scene()`)
- Move loader/transitions as properties onto Scene. `transitions()` can be a getter to allow for the Scene to change the transition, but maybe that's unnecessary. Could just be an object property instead, or a static if we want to know them before instantiating the Scene.


Downsides:
- If there was intention to support dynamic scenes, e.g. `level1: () => import('./level1.ts')`, the loader/transition is not known until the dynamic import has loaded. However, if the transition can be set on navigation e.g. `goToScene(name, { transitions: ... })`, that's a fair workaround. One might decide to always specify transitions on the navigation instead of by Scene if they want to lazy load Scene files

mattjennings:
One other thing to consider if if you'd ever want to get the current scene's "name". In the above example, it'd be `"level1"` since that's the key assigned to it. I suppose you could do something like this:

```ts
const game = new ex.Engine({
  scenes: [Level1Scene]
})

class Level1Scene extends Scene {
  name = 'level1'
}
```

this way the Scene class explicitly defines this. But I'm not a fan of this, personally. It would also prevent you from having the same scene for multiple names (though maybe that could be seen as an anti-pattern anyway with this new routing).

eonarheim:
> Gets rid of routes terminology which is probably too web dev-y anyway. I was also concerned it might cause confusion of what a route is vs a scene

I like this idea! That is a plus for sure, folks might not know what "route" is unless they come from webdev. And you are forced to give them an explicit unique name.

> Move loader/transitions as properties onto Scene. transitions() can be a getter to allow for the Scene to change the transition, but maybe that's unnecessary. Could just be an object property instead, or a static if we want to know them before instantiating the Scene.

Let me noodle, I agree these do feel like Scene concerns, but also like Engine/configuration? Maybe they feel like both? Allow ahead of time definitions but scenes can choose to override by returning a non-falsey transition perhaps?


> If there was intention to support dynamic scenes, e.g. level1: () => import('./level1.ts'), the loader/transition is not known until the dynamic import has loaded. However, if the transition can be set on navigation e.g. goToScene(name, { transitions: ... }), that's a fair workaround. One might decide to always specify transitions on the navigation instead of by Scene if they want to lazy load Scene files

Good point, I do want to support lazy loading. I feel like this might be a necessary thing for bigger games eventually. Maybe this is a reason to also support ahead of time configuration in addition to a dynamic property/method for transitions?

mattjennings:
> Let me noodle, I agree these do feel like Scene concerns, but also like Engine/configuration? Maybe they feel like both? Allow ahead of time definitions but scenes can choose to override by returning a non-falsey transition perhaps?

Yeah I think both are valid to have concerns for this. To me, it makes the most sense to have the defaults defined as the Engine/configuration level, but I wonder what the common user would intuitively think. Probably worth taking a look at how Unity/Godot/Unreal approach this. A web dev might expect the router configuration, but a dev coming from another game engine might expect differently.

In an effort to remove the `route` terminology, it'd be nice to get rid of the `scene` redundancy

```ts
scenes: {
   level1: {
     // bleh
     scene: MyScene
  }
} 
```

Spitballing:

You _could_ do something like this but it's not my favorite

```ts
scenes: {
   level1: MyScene,
   level2: [MyScene, { ...options }]
}
```

Alternatively.... double down on the Route terminology. This does strengthen the line between what lives on a Scene vs what doesn't, and what a "route" is, but it's more verbose.

```ts
routes: {
   level1: new Route(MyScene, {... options })
}
```

eonarheim:
Woooo! Finally!

PR #2793, opened 2023-10-26
feat: Docs vNext (Docusaurus Migration)

kamranayub:

==================

Closes https://github.com/excaliburjs/excaliburjs.github.io/issues/383

## Preview Site

👉 https://excaliburjs.pages.dev/

## Local Dev

```
cd site
npm install
npm start
```

## Changes:

- Added `site/` directory to hold Docusaurus site

### Foundations Phase

_Everything that will enable the full migration to work_

- [x] Add Typedoc API integration and API reference section
- [x] Test linking from MDX to APIs
  - [x] **fix:** initial build does not link symbols (affects first local build + deployed site)
- [x] Migrate and test our custom MDX components
  - [x] `Note` replaced with Admonitions
  - [x] `CodeSandboxEmbed` 
  - [x] `IFrameEmbed`
  - [x] `Example` storybook embed
  - [x] `embed:` snippets (Live Code?)
- [x] Move over all the previous MD/X files
- [x] Scaffold the other static pages
- [x] ~Integrate Tailwind~
  - Tutorial: https://github.com/swyxio/swyxdotio/issues/456
  - Do we "really" need it, or could we make do with Infima until official preset/theming is in?
- [x] [Upgrade to Docusaurus v3](https://docusaurus.io/docs/next/migration/v3)
  - [x] TypeScript 5.x
  - [x] Typedoc 0.25.3
- [x] Make sure all the docs are working and links are updated
- [x] Figure out more integrated way to provide _compiled_, interactive side-by-side examples
- [x] Update community showcase
- [x] Update donation/support page
- [x] Migrate blog posts and come up with a workflow that will work well (+ newsletter?)
- [ ] Replace Storybook examples with new Playgrounds
  - [x] Upgrade to Storybook 7.5.x
  - [ ] Migrate examples
- [x] Investigate [social cards](https://github.com/facebook/docusaurus/issues/2968)
- [ ] Add product analytics (Posthog)
- [ ] Make new homepage

### Docs Buildout Phase

- [ ] Organize old docs into new unified docs system
- [ ] Redo and update the Quickstart
- [ ] Update or make new diagrams (Mermaid?)
- [ ] Bring in community tutorials/resources
- [ ] Figure out old API versioning build flow
- [ ] Provide at least one recipe/learning path beyond the Quickstart

## Notes

- **Separate tooling:** Right now the `site` directory is managed separately from the rest of the engine package building. I may explore unified tooling using a monorepo approach, but it will require moving folders around + redoing all the build tooling. If we move Docusaurus to run/build from the root of the repo, there are some conflicts with the `build` temp directory, as well as `docs` directory and `src` directories. We'd have to change those and I haven't looked into it yet.

- **Cloudflare root directory:** Since the Docusaurus build needs to reach into other parts of the repo (hello Typedoc), I was getting errors when I set the Pages build root to `site`. I had to remove that and then modify the build command to `cd site && npm ci && npm run build` so the build takes longer. It'd be great not to have to (but we may need to do that if we cannot access the repo files outside the `site` dir).

- ~**Initial build doesn't link symbols:**~ **Fixed**. Since the docusaurus-plugin-typedoc-api runs as a plugin lifecycle, and we need the `.json` Typedoc file as input into the `docs.remarkPlugins` config, it has to wait until the first successful build to get the JSON. I am not sure if there's a way to fix this without modifying the other plugin. 🤔 

- **Custom symbol link builder:** I had to patch my https://github.com/kamranayub/remark-typedoc-symbol-links package to support custom link building since it was hardcoded to Typedoc paths. I could extract this into a npm package perhaps. It would be ideal to bake it into docusaurus-plugin-typedoc-api though 🤔 Not sure how to handle the remarkPlugin piece with `@mdx/loader`.

- **Integrated Storybook build:** Previously we had to run `npm run visual` or `npm run build-storybook` to work with Storybook examples. I have a POC that adds direct dev server + build integration with Docusaurus so everything happens in one "experience." This will move SB to the `site` directory instead of the root, keeping things tidy. It will also allow us to move the stories _into_ the `examples/` folders within `docs`, making story creation even more seamless. There is some uncertainty with how the SB dev server will interact with the Docusaurus dev server, and process file handles, exiting gracefully, etc.


kamranayub:
## Notes for live code research

**Ideal DX:** write TS code files that are compiled alongside MDX files, and can be embedded, with live code + preview.

I believe Storybook will get us that but the challenge is that the build pipeline is separate. However, it feels _possible_ to integrate Storybook's build as a Docusaurus plugin.

- Builder API: https://storybook.js.org/docs/html/builders/builder-api
- Custom framework: https://storybook.js.org/docs/html/api/new-frameworks
- Live code add-on: https://github.com/hirohe/storybook-addon-jarle-monaco

kamranayub:
## Notes on `GameCodeBlock`

I wrapped the live codeblock with our own custom MDX component, which helps set up the code block with different modes (Live vs. Static).

Right now, the usage is like this:

```mdx
import AddingActorsExample from '!!raw-loader!./examples/adding-actors.ts';

<GameCodeBlock>{AddingActorsExample}</GameCodeBlock>
```

This feels a little "boilerplate-y", so I will probably explore this instead:

```mdx
<GameCodeBlock src="./examples/adding-actors.ts" />
```

This I think feels a bit more natural. Note that you can still include an inline code block, but that will not get typechecked.

There are a few other critical enhancements to support I think before the MVP is ok:

- Start the live result paused/at start screen otherwise users will miss any action
- Have a restart/replay button

I still expect that there will be an issue here with multiple live editors/engines on the screen. I could workaround it by "fake isolating" them; i.e. only allowing one to be active at any time.

eonarheim:
@kamranayub MERGED! 🎉 🎉 🎉

PR #2857, opened 2023-12-29
feat: [#2856] Emit actionstart and actioncomplete events on actors

### LINKED ISSUE #2856: Emit "actionstart" and "actioncomplete" events
It would be useful the ActionsComponent emitted events when starting/completing actions

### Context

I have a moving platform that has its path defined by actions

```ts
    const platform = new MovingPlatform(
      {
        x: 300,
        y: 50,
        width: 100,
        height: 16,
      },
      (actions) =>
        actions.repeatForever((builder) =>
          builder.moveBy(0, -50, 40).moveBy(0, 50, 40)
        )
    )
```

I have some particular logic that I need to run whenever the platform changes direction that should happen immediately after the velocity is changed. Hooking into an event would be a very nice way to solve this.

### Proposal

Action component the following events containing the relevant action:

- `actionstart`
- `actioncomplete`

These can be either on the actor or the actions component, or both? This would also fire for each action in `repeatForever` etc.

```ts
type ActionStartEvent {
    action: ex.Action
}

type ActionCompleteEvent {
   action: ex.Action
}

this.on('actioncomplete', (ev) => {
   
})
```

mattjennings:

==================

Closes #2856

## Changes:

- Added `actionstart` and `actioncomplete` events to the Actor that are fired when an action starts and completes

### DIFF
--- a/src/engine/Events.ts
+++ b/src/engine/Events.ts
@@ -13,6 +13,7 @@ import { OnInitialize, OnPreUpdate, OnPostUpdate, SceneActivationContext } from 
 import { BodyComponent } from './Collision/BodyComponent';
 import { ExcaliburGraphicsContext } from './Graphics';
 import { Axes, Buttons, Gamepad } from './Input/Gamepad';
+import { Action } from './Actions/Action';
 
 export enum EventTypes {
   Kill = 'kill',
@@ -80,7 +81,10 @@ export enum EventTypes {
   PointerDragEnd = 'pointerdragend',
   PointerDragEnter = 'pointerdragenter',
   PointerDragLeave = 'pointerdragleave',
-  PointerDragMove = 'pointerdragmove'
+  PointerDragMove = 'pointerdragmove',
+
+  ActionStart = 'actionstart',
+  ActionComplete = 'actioncomplete'
 }
 
 /* istanbul ignore next */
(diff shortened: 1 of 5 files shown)


src/engine/Events.ts
```diff
@@ -545,3 +549,21 @@ export class ExitTriggerEvent extends GameEvent<Actor> {
     super();
   }
 }
+
+/**
+ * Event thrown on an [[Actor]] when an action starts.
+ */
+export class ActionStartEvent extends GameEvent<Entity> {
```
mattjennings, line 556: @eonarheim I wasn't too sure if this should be `GameEvent<Entity>` or if you'd have wanted an interface to be made so I could do the `<T extends XYZ = Entity>` trick?

eonarheim: @mattjennings This is perfect for now! Fits with the current pattern. 

Eventually I want to rethink GameEvent entirely (maybe remove it?) but I haven't put enough effort into what the future would look like.

eonarheim (approved):

eonarheim:
Looks like a test failure, not sure if this is flakey yet

src/engine/Events.ts
mattjennings: sounds good! 

It's pretty nice having these to use as types when setting up handlers on a class:

```ts
onInitialize() {
     this.on('actionstart', this.onActionStart)
}

onActionStart(ev: ex.ActionStartEvent) { ... }
```

so I'd advocate for keeping them

mattjennings:
@eonarheim I think it was legitimate - just pushed a change and it's awaiting approval to run again

eonarheim (approved):

mattjennings:
@eonarheim I wonder if there should be an `actionupdate` event too?

eonarheim:
> @eonarheim I wonder if there should be an `actionupdate` event too?

I suppose it might be useful to know what action is running, although start and complete will give you that. Perhaps useful to peer into the current actions state?

eonarheim:
I'm okay without an update for now

mattjennings:
```ts
    this.actions.moveBy(0, 50, 40) 

    this.on('actionstart', (ev) => {
      console.log(this.vel)
    })

    this.on('actioncomplete', (ev) => {
      console.log(this.vel)
    })
```

In this example, `this.vel` will be 0,0 both times as it either hasn't run its first update yet or is completed already. For this use-case though I could just store the vel from the actor's update loop and then compare in the actioncomplete.

I think I'm fine without it too until convinced of a necessary use-case.

eonarheim:
@mattjennings Thanks again!

PR #2869, opened 2024-01-06
feat: Collision lifecycle convenience methods on Actor

eonarheim:

==================

This PR adds new collision lifecycle convenience methods to `Actor`! This was possible before by wiring up an event handler. This is a super common pattern so now it's moved to Actor!

```typescript
class MyActor extends ex.Actor {
  constructor(args: ex.ActorArgs) {
    super(args);
  }
  onPreCollisionResolve(self: ex.Collider, other: ex.Collider, side: ex.Side, contact: ex.CollisionContact): void {
    
  }
  onPostCollisionResolve(self: ex.Collider, other: ex.Collider, side: ex.Side, contact: ex.CollisionContact): void {
    
  }
  onCollisionStart(self: ex.Collider, other: ex.Collider, side: ex.Side, contact: ex.CollisionContact): void {
    
  }
  onCollisionEnd(self: ex.Collider, other: ex.Collider): void {
    
  }
}
```

### DIFF
--- a/sandbox/src/game.ts
+++ b/sandbox/src/game.ts
@@ -50,6 +50,7 @@ var game = new ex.Engine({
   pointerScope: ex.PointerScope.Canvas,
   displayMode: ex.DisplayMode.FitScreenAndFill,
   snapToPixel: false,
+  fixedUpdateFps: 60,
   maxFps: 60,
   configurePerformanceCanvas2DFallback: {
     allow: true,
@@ -552,6 +553,13 @@ follower.actions
     console.log('Player met!!');
   });
 
+player.onCollisionStart = (_a,_b,_c, contact) => {
+  console.log('start', contact);
+}
+player.onCollisionEnd = (_a,_b) => {
+  console.log('end');
+}
+
 // follow player
 
 player.rotation = 0;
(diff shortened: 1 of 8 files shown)


sandbox/src/game.ts
```diff
@@ -581,6 +589,29 @@ player.addChild(healthbar);
...
 
+class OtherActor extends ex.Actor {
+  constructor(args: ex.ActorArgs) {
+    super(args);
+  }
+  onCollisionStart(self: ex.Collider, other: ex.Collider, side: ex.Side, contact: ex.CollisionContact): void {
+    console.log('other collision start')
+  }
+  onCollisionEnd(self: ex.Collider, other: ex.Collider): void {
+    console.log('other collision start')
```
alanag13, line 600: Probably meant `other collision end`

src/engine/Actor.ts
```diff
@@ -878,6 +878,49 @@ export class Actor extends Entity implements Eventable, PointerEvents, CanInitia
...
+   * @param other
+   * @param side
+   * @param contact
+   */
+  public onPreCollisionResolve(self: Collider, other: Collider, side: Side, contact: CollisionContact) {
+    // Override me
+  }
+
+  /**
+   * Fires every after collision resolution for a confirmed contact.
```
alanag13, line 893: Missing word, probably `frame`, after `Fires every`

eonarheim: Oh yup, should read "fires for every collision resolution for a confirmed contact"

Issue #2981, opened 2024-03-25
Code examples on website are shown in light and dark mode always


uvd540:
### Steps to Reproduce
1. Open Firefox (maybe other browsers too?).
2. View any of the documentation pages. For example, the [Hello World](https://excaliburjs.com/docs/getting-started) example.
3. Toggle dark mode on/off. In either case, each code example is shown in both light and dark modes.
![image](https://github.com/excaliburjs/Excalibur/assets/19933462/65a877bc-8bb6-4c35-a110-444a6d4167b2)

### Expected Result
Only see the light/dark code examples based on dark mode preference

### Actual Result
Seeing both code examples no matter what mode is selected

### Environment
- Firefox 115.9.1esr (64-bit)
- macOS 10.14
- Excalibur versions: Current

### Current Workaround
None

eonarheim:
@uvd540 Thanks for the bug! Very interesting I'm not able to repro on FF 124.0.1 on windows.

I'll do a little digging on what's going on, possibly a bug in FF?

What does Chrome/Safari show for you?

uvd540:
I'm seeing the same issue on Safari as well.
![image](https://github.com/excaliburjs/Excalibur/assets/19933462/4de4b2d7-8281-4aa5-bbce-2d9e22b6207a)

eonarheim:
I can reproduce it on browserstack with:
 
* Mojave/Safari 12
* Catalina/Safari 13
* BigSur/Safari 14

![image](https://github.com/excaliburjs/Excalibur/assets/612071/f1513d3a-8b74-47dd-862c-db9f89dee131)

Things only appear to work as intended at Monterey/Safari 15

I think it relates to this pseudo class selector using a `:has()` which isn't supported until Safari 15.4 and Firefox 121 https://caniuse.com/css-has.
![image](https://github.com/excaliburjs/Excalibur/assets/612071/7de566a3-5a51-489d-834a-c6b44f29504e)

This feature comes pre-bundled with Docusaurus v3 which we use for documentation, I'll see if their issue tracker has anything relating to this.

@kamranayub any ideas?

eonarheim:
It looks like we've fallen off the older safari support by default in browserlist

https://docusaurus.io/docs/browser-support

```
❯ npx browserslist --env="production"
Need to install the following packages:
  browserslist@4.23.0
Ok to proceed? (y) y
and_chr 122
and_ff 123
and_qq 14.9
and_uc 15.5
android 122
[...]

PR #3019, opened 2024-04-07
docs: polyfill css :has with postcss plugin

mattjennings:

==================

Polyfills CSS :has using the [css-has-pseudo](https://www.npmjs.com/package/css-has-pseudo) postcss plugin. It also involves a JS polyfill (~12kb). Theoretically solves #2981 ... but I haven't tested this on an affected browser. 

There's potential for it to not be perfect (see "Known Shortcomings" section in its readme), but initially it seems OK, but given the CSS+JS size increase + potential issues with docusaurus pages, we'll have to consider if it's worth it or not.

Highlighted section is the polyfill I believe

CleanShot 2024-04-07 at 18 32 17@2x

### DIFF
--- a/site/docusaurus.config.ts
+++ b/site/docusaurus.config.ts
@@ -173,7 +173,11 @@ const config: Config = {
               })
             ]
           };
-        }
+        },
+        configurePostCss(postcssOptions) {
+          postcssOptions.plugins.push(require('css-has-pseudo'));
+          return postcssOptions;
+        },
       };
     },
     [
--- a/site/docusaurus.config.ts
+++ b/site/docusaurus.config.ts
@@ -311,7 +311,11 @@ const config: Config = {
        */
       playgroundPosition: 'bottom'
     }
-  } as ClassicPresetThemeConfig
+  } as ClassicPresetThemeConfig,
+
+  clientModules: [
+    './src/polyfills/css-has.js'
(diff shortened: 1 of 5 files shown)


eonarheim:
@mattjennings Awesome I had no idea this was a thing!

eonarheim:
Bummer looks like safari 13 doesn't like it still in browserstack, I'll try it locally with browserstack to be sure

image

mattjennings:
Ahh shoot. I was surprised such a thing existed as well, as I remember this being one of those things that you couldn't polyfill for the longest time.

Well, IMO this is a docusaurus issue, and this was probably the best shot we had at doing anything about it save for migrating to an entirely different docs framework. So we may have to close #2981 as a can't/won't fix, unfortunately.

eonarheim:
> Ahh shoot. I was surprised such a thing existed as well, as I remember this being one of those things that you couldn't polyfill for the longest time.
> 
> Well, IMO this is a docusaurus issue, and this was probably the best shot we had at doing anything about it save for migrating to an entirely different docs framework. So we may have to close #2981 as a can't/won't fix, unfortunately.

Agreed, I'll open a bug against docusaurus and close ours

eonarheim:
Closing for now, opened an issue on docusaurus

PR #3032, opened 2024-04-27
Feat: provide development build for bundlers, warn on unadded entity

mattjennings:

==================

This adds a "development" build of Excalibur that [bundlers can use](https://webpack.js.org/guides/package-exports/#providing-devtools-or-production-optimizations). It allows us to include code that can be used to help the user during development but still keep it out of production builds as to not affect performance / size. 

This requires removing the `module` field in package.json in favour of `exports`, which is OK (`exports` essentially supersedes `module`). **There is potential for breaking change** here, as any nested imports e.g. `'excalibur/dist/abc'` would now need to be added to this exports map. But I don't believe there was ever a reason to do that, since everything is exported from the excalibur.js file. (Types are not affected)

I've tested this in a Vite project and it works, haven't tested it with Webpack yet.

## Changes:

- create development builds of excalibur that bundlers can use in dev mode
- show warning in development when Entity hasn't been added to a scene after a few seconds

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - `actor.oldGlobalPos` returns the globalPosition from the previous frame
 - Built in actions now have a unique `id` property
+- create development builds of excalibur that bundlers can use in dev mode
+- show warning in development when Entity hasn't been added to a scene after a few seconds
 
 ### Fixed
 
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -104,7 +104,8 @@ module.exports = (config) => {
       },
       plugins: [
         new webpack.DefinePlugin({
-          'process.env.__EX_VERSION': '\'test-runner\''
+          'process.env.__EX_VERSION': '\'test-runner\'',
+          'process.env.NODE_ENV': JSON.stringify('test')
         }),
       ],
       module: {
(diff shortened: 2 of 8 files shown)


eonarheim:
@mattjennings This looks very reasonable (Also I'm excited for the DX opportunities!!) I'm will to try it out in the alpha build ASAP.

I'm not as concerned with the deep imports `'excalibur/dist/abc'` we never supported that officially and I'm willing to document that as a break for v0.30

mattjennings:
I pushed a change to remove the the tsc .js output. I believe the `excalibur-version.js` script now is not necessary, and may not even be doing anything, so we could remove that too while we're here.

I was also thinking of removing the `excalibur.min.development.js` output to speed up the build a bit as it won't be used by any of this export config, but is there a chance you might want this available for those using it via unpkg CDN etc?

eonarheim:
> I pushed a change to remove the the tsc .js output. I believe the `excalibur-version.js` script now is not necessary, and may not even be doing anything, so we could remove that too while we're here.

Awesome, yes we spoke offline about this. I believe you are correct, we'll let it cook in alpha for a while.

> I was also thinking of removing the `excalibur.min.development.js` output to speed up the build a bit as it won't be used by any of this export config, but is there a chance you might want this available for those using it via unpkg CDN etc?

100% it might be a nice feature to have the minified dev version available via CDN

Issue #2923, opened 2024-02-03
Support anchor in pixels 


mattjennings:
### Context

Sometimes it's helpful to be able to define an Actor's anchor in pixels instead of %. e.g. if your sprite isn't perfectly centered on the frame and the anchor point would be 8px down from the middle, trying to find the % for that exact offset is  tricky.

### Proposal

Two proposals:

## an Anchor class that defines the unit

My initial idea when writing this issue was an `Anchor` class that could be used as the `anchor` argument.

```ts
new ex.Actor({
  anchor: new ex.Anchor(x,y, 'px' | 'relative')
})
``` 

However it's ambiguous what this offset would be relative to. I suppose top-left is what it would be, but there are times you might want it to be offset from the center (but that just means some extra math based on width/height, so maybe it's not a big deal).

## an anchorOffset argument

If there were an additional argument `anchorOffset` that would offset in px from the `anchor` value, that would give you the flexibility to easily shift things around as needed. I just don't like having 2 different properties for this. It could cause confusion.

eonarheim:
@mattjennings I agree! I think a pixel offset is useful in this case. Let me play around with the first one, a `ex.Anchor` type that extends vector would be a relatively easy drop in.

We do have a way to do this (but not super visible on actor). On the graphics component we have a `actor.graphics.offset` property that will shift graphics by pixels.

```typescript
const actor = new ex.Actor({...});
actor.graphics.anchor = ex.vec(0, 0); // zero out the anchor
actor.graphics.offset = ex.vec(-8, -8); // offset 
```

Internally this is the drawing code in GraphicsSystem

```typescript
    let anchor = graphicsComponent.anchor;
    let offset = graphicsComponent.offset;

    // handle specific overrides
    if (options?.anchor) {
      anchor = options.anchor;
    }
    if (options?.offset) {
      offset = options.offset;
    }
    // See https://github.com/excaliburjs/Excalibur/pull/619 for discussion on this formula
    const offsetX = -graphic.width * anchor.x + offset.x;
    const offsetY = -graphic.height * anchor.y + offset.y;

    const oldFlipHorizontal = graphic.flipHorizontal;
    const oldFlipVertical = graphic.flipVertical;
    if (flipHorizontal || flipVertical) {
      // flip any currently flipped graphics
      graphic.flipHorizontal = flipHorizontal ? !oldFlipHorizontal : oldFlipHorizontal;
[...]

PR #3054, opened 2024-05-09
feat: add Anchor class

SteveSuv:
## Changes:

- add commonly used anchor vector as Anchor Class, no need to write `ex.vec(x, y)` but use `ex.Anchor.TopLeft`, it makes code more stable.

### DIFF
--- a/src/engine/Anchor.ts
+++ b/src/engine/Anchor.ts
@@ -1,0 +1,58 @@
+import { Vector, vec } from "./excalibur";
+
+export class Anchor {
+  /**
+   * get TopLeft Anchor Vector
+   */
+  public static get TopLeft(): Vector {
+    return vec(0, 0)
+  }
+  /**
+   * get TopCenter Anchor Vector
+   */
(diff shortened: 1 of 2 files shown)


src/engine/Anchor.ts
```diff
@@ -0,0 +1,58 @@
+import { Vector, vec } from "./excalibur";
+
+export class Anchor {
+  /**
+   * get TopLeft Anchor Vector
+   */
+  public static get TopLeft(): Vector {
```
Autsider666, line 7: The Vector class already has a set of these static functions, so why should these be in a new class? https://github.com/excaliburjs/Excalibur/blob/main/src/engine/Math/vector.ts#L16

Vector doesn't contain getters for diagonal vectors yet, so I'd love to see yours added to it.

eonarheim:
Closing for now I don't think we have a strong direction here yet, I'm willing to chat about this more in a new issue, and we have some discussion about anchoring changes #2923 that are somewhat related

PR #3056, opened 2024-05-10
fix: Cirular dependency on Actor

### LINKED ISSUE #3055: Bug with custom build: Circular Dependency Causes `Actor` to be `undefined` in `ParticleEmitter`
### Description
I am using TypeScript source files directly from Excalibur.js instead of the bundle for several advantages:
- Better source maps as my project directly uses the sources, it can generate better source maps.
- I can decide for myself which target (e.g. ESNext, ESM, ...) I want to use
- Ability to make changes without needing to rebuild Excalibur with fast page reload.
- Direct access to the source code in my IDE.
- Faster builds with esbuild (or possibly vite, which also utilizes esbuild) which is significantly faster than Excalibur’s webpack.

This approach has been working well until a recent issue where `Actor` is `undefined` when referenced in the `ParticleEmitter` class definition. This seems to be due to circular dependencies and import order issues, causing `Actor` to be undefined when `ParticleEmitter` tries to extend it in my own bundled file.

I understand if this is not recognised as a bug as this is a very specific problem, I am happy to fix it myself and hope it is accepted.

### Steps to Reproduce

Build Excalibur with Vite or ESBuild with a small sample game, this needs configurations so that this also works with the imported `*.css` and `.glsl` files because vite needs to import them using a `?raw` prefix by default like this:

```ts
import lineVertexSource from './line-vertex.glsl?raw';
```

I can create an working example using esbuild or vite if there is a need for it.

### Expected Result

`ParticleEmitter` should successfully extend `Actor` without encountering an undefined issue.

### Actual Result

The `Actor` class is `undefined` when `ParticleEmitter` attempts to extend it, due to circular dependencies and import order issues within the TypeScript source files.

### Environment

- Excalibur versions: latest main branch
- Build tools: esbuild/vite (latest versions)

### Current Workaround

Just use the official bundle instead of the sources directly as intended.

JumpLink:
- [x] :page_facing_up: changelog not needed

==================

This PR resolves a circular dependency issue in Excalibur.js when using TypeScript source files directly. It refactors `ParticleEmitter` by isolating it into a separate file and adjusting the dependency management in `Particles.ts`.

This PR has the advantage that you should not encounter this problem if Webpack should ever be replaced.

Closes  #3055

## Changes
* Isolated `ParticleEmitter`: Moved `ParticleEmitter` to its own file to manage dependencies more effectively.
* Updated `Particles.ts`: Replaced direct dependency with a type import of `ParticleEmitter`.
* Instance Check Adjustment: Replaced the `instanceof` check with a `constructor` presence check.


eonarheim:
@JumpLink Looks good locally! We added a new prettier to format files on stage, please run prettier on the files changed! 

`npm run format:fix`

JumpLink:
> @JumpLink Looks good locally! We added a new prettier to format files on stage, please run prettier on the files changed!
> 
> `npm run format:fix`

Also just realised ;)

eonarheim:
@JumpLink Thanks again! Much appreciated!

PR #3057, opened 2024-05-11
feat: YSortComponent

mattjennings:

==================

## Changes:

- YSortComponent that adjusts z-index based on its global Y position. This can be used on an
Actor with the new `ySort` configuration in the constructor, or added as a component to an Entity.

  ```ts
  new ex.Actor({
    ySort: true,
    // or
    ySort: {
      offset: 0, // apply an offset to the resulting z-index
      order: 1 // 1 will add the pos.y to the z-index, -1 will subtract it
    }
  });

  // or
  entity.addComponent(
    new ex.YSortComponent({
      offset: 0,
      order: 1
    })
  );
  ```

https://github.com/excaliburjs/Excalibur/assets/8703090/eae21125-eca7-497f-a913-43a75d41007b

Open to any feedback or suggestions on functionality / API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ Actor with the new `ySort` configuration in the constructor, or added as a compo
 
   ```ts
   new ex.Actor({
+    ySort: true,
+    
+    // or
+
     ySort: {
       offset: 0, // optionally apply an offset to the resulting z-index
       order: 1 // 1 will add the pos.y to the z-index, -1 will subtract it
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,11 +27,11 @@ Actor with the new `ySort` configuration in the constructor, or added as a compo
   ```ts
   new ex.Actor({
     ySort: true,
-    
+
     // or
 
     ySort: {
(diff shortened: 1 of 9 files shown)


eonarheim (approved):
@mattjennings Looks completely reasonable to me, no notes! Let me know if you're ready to merge

mattjennings:
After discussion on discord, we decided:

- Scene should have a config that will enable ysorting for _all_ entities.
- YSortComponent should take optional min/max z values as a way to enable layering.

Further ideas from myself since discussion:

- The Scene config should only apply to entities that contain a YSortComponent
- YSortComponent is included in all actors by default, and uses the Scene's ySort config unless overridden. In other words, if it's enabled by Scene, then every Actor sorts.
- Caveat: would mean we need to add YSortComponent to tilemap plugins etc. May be easy to miss it when composing custom entities. Maybe the YSort logic should be in the TransformComponent?

```ts
class MyScene extends ex.Scene {
  constructor() {
      super({ 
        ySort: true,
        // and/or ?
       ySort: { enabled: true, ...otherConfig }
      })
   }
}

class MyActor {
   constructor() {
      super({
          // ySort config defaults to scene's ySort
      })
   }
}
```

PR #3070, opened 2024-05-20
Feat: z-indexes are relative to their parent

mattjennings:

==================

## Changes:

- Z-indexes are now relative to the parent's Z-index. You can get the global Z-index with the `globalZ` property on the Actor or TransformComponent.

### DIFF
--- a/src/engine/Graphics/GraphicsSystem.ts
+++ b/src/engine/Graphics/GraphicsSystem.ts
@@ -65,7 +65,7 @@ export class GraphicsSystem extends System {
     this._graphicsContext = this._engine.graphicsContext;
     if (this._zHasChanged) {
       this._sortedTransforms.sort((a, b) => {
-        return a.z - b.z;
+        return a.globalZ - b.globalZ;
       });
       this._zHasChanged = false;
     }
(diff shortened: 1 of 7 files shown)


src/engine/Graphics/GraphicsSystem.ts
```diff
@@ -265,7 +265,7 @@ export class GraphicsSystem extends System {
       }
 
       if (transform) {
-        this._graphicsContext.z = transform.z;
+        this._graphicsContext.z = transform.globalZ;
         this._graphicsContext.translate(tx.pos.x, tx.pos.y);
```
mattjennings, line 269: @eonarheim unsure if this is the right approach here. the other values aren't using the global here, so this leads me to think `translate` should maybe be updated to take into account the z?

eonarheim: I think that global z is right, the graphics context uses the z to sort so this seems right

mattjennings: gotcha. well, it works like this, so i'm fine to leave it like so if you are.

eonarheim (approved):
@mattjennings Looks great, thanks again!

PR #3091, opened 2024-06-01
feat: added getters width and height to Tilemap

jyoung4242:
see title, added a public getter for width and height to the tilemap class


eonarheim:
@jyoung4242 I think this branch accidentally includes your wfc blog post changes

PR #3094, opened 2024-06-10
feat: Added default value to Canvas constructor

Autsider666:
The combined type for the constructor property has no required fields. This way you don't have to add an empty object to the superclass's constructor when, for example, creating a Canvas subclass


Autsider666:
Making `height` and `width` required could potentially be an even better option if the warning in [the docs](https://excaliburjs.com/docs/canvas) is to be believed, but I'm sure there are people that are able to do some magic this way by working around it.

eonarheim (approved):
@Autsider666 Good find! Many thanks

PR #3095, opened 2024-06-11
feat: removed invalid null as value from Random.shuffle()

Autsider666:
Setting `null` on a variable of type `T` isn't valid in the eyes of TS


Autsider666:
Ran into this one because I'm copy-pasting some often used classes into my own codebase (while referencing EX ofc) so I can experiment with workers

eonarheim (approved):

eonarheim:
@Autsider666 Thanks again!

PR #3112, opened 2024-06-29
fix: getTile return type should include null

Autsider666:
I thought EX had checks in the pipeline to prevent these types of TS inconsistencies.


Autsider666:
Is there a reason why null is returned instead of undefined by the way?

eonarheim (approved):
@Autsider666 This should be the type signature! Much appreciated

eonarheim:
@Autsider666 There are various levels of strictness throughout the TS confgs unfortunately due to history, the desired end state is strict TS

PR #3113, opened 2024-06-30
feat: expose all actor exports

Autsider666:
I originally only intended to add `ActorEvents` to the exports, so it's possible to extend it without a dirty import:
 `import {ActorEvents} from "excalibur/build/dist/Actor";`

But I don't see any reason to not export all of the Actor exports, because `isActor` could also be really useful for people extending core EX functionality.


eonarheim (approved):
@Autsider666 makes sense to me!

PR #3176, opened 2024-08-23
Feat: [#3157] return target on trigger

### LINKED ISSUE #3157: feat: Trigger - access to action's actor
### Discussed in https://github.com/excaliburjs/Excalibur/discussions/3099

<div type='discussions-op-text'>

<sup>Originally posted by **KostarSf** June 17, 2024</sup>
Hi! It would be very useful to have access to the Actor, which calls Trigger's action callback. Something like this:

```ts
const trigger = new Trigger({
  // rest
  action: (actor) => {
    // do action
  }
})
```

I want to use triggers as areas of view for AI enemies, so it's important for me to know which Actor called action. Now I have to do my own implementation of the trigger, which takes this into account.

It would be great if it appeared in Excalibur!</div>

Autsider666:

==================

Couldn't get the [suggested setup](https://github.com/excaliburjs/Excalibur/issues/3157#issuecomment-2266842536) to work without adding a lot of extra (code) complexity. I'd recommend dropping that requirement, because it's a lot easier to implement by the developers themselves by adding the check in their own `filter` callback to assure an entity is of a specific type (and use `as ....` to placate TS if needed).

I'll finish the last 2 points on the checklist after there's a green light on my current implementation.

Closes #3157

## Changes:

- The action callback of Trigger now returns the entity that triggered it.
- General cleanup of Trigger class
  - `target` is now 100% separate from `filter`, because that's what the functionality described in the documentation implied (to me).
  - Converted all the different "default value fallbacks" to a single setup.
  - Trigger used `Entity` and `Actor` interchangeably. Rewrote it to be 100% Entity (Although Entity doesn't have any collision mechanics build in, so an argument could be made to go for 100% Actor)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Breaking Changes
 
+- `Trigger` API has been slightly changed:
+  - `action` now returns the triggering entity: `(entity: Entity) => void`
+  - `target` now works in conjunction with `filter` instead of overwriting it.
+  - `EnterTriggerEvent` and `ExitTriggerEvent` now contain a `entity: Entity` property instead of `actor: Actor`
 - `ex.Vector.normalize()` return zero-vector (`(0,0)`) instead of `(0,1)` when normalizing a vector with a magnitude of 0
 - `ex.Gif` transparent color constructor arg is removed in favor of the built in Gif file mechanism
 - Remove core-js dependency, it is no longer necessary in modern browsers. Technically a breaking change for older browsers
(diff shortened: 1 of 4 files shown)


Autsider666:
Ah, those tests have to be fixed.
Is there any magic ritual to get the sandbox command to run locally? I was hope sacrificing a chicken was enough, but it'll take a bit to find a goat...
image

eonarheim:
@Autsider666 there is some flakey unit tests on main I haven't nailed down yet either :/

eonarheim (approved):
@Autsider666 looks great!

Autsider666:
@eonarheim Updated the changelog, so PR is ready to be merged

eonarheim:
@Autsider666 Many thanks!

PR #3179, opened 2024-08-24
Feat: [#3171] normalize zero vector

### LINKED ISSUE #3171: Unexpected result of .normalize() on already Zero vector
### Steps to Reproduce

Attempt to normalize a vector of length zero

image

Other prior art is that Godot/Unity/Unreal all return a zero vector

* https://github.com/godotengine/godot/blob/master/core/math/vector2.cpp
* https://docs.unity3d.com/ScriptReference/Vector2.Normalize.html
* https://dev.epicgames.com/documentation/en-us/unreal-engine/BlueprintAPI/Math/Vector/Normalize?application_version=4.27 

### Expected Result

Return a vector copy that is also zero

### Actual Result

Arbitrary direction is returned

### Environment

Latest

### Current Workaround

Check for zero vectors before normalizing

Autsider666:

==================

Closes #3171

## Changes:

- `Vector.normalize()` return zero-vector (`(0,0)`) instead of `(0,1)` when normalizing a vector with a magnitude of 0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Breaking Changes
 
+- `ex.Vector.normalize()` return zero-vector (`(0,0)`) instead of `(0,1)` when normalizing a vector with a magnitude of 0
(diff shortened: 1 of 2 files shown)


eonarheim (approved):
@Autsider666 Thanks for the fix! Much appreciated

PR #3275, opened 2024-11-09
docs: Update tutorial collision

dca123:
I was following along the tutorial and noticed that I couldn't get the collision working. It seems like owner isn't available in `ev.other`. I removed owner, since `ev.other` is an actor instance and called kill on it to get the tutorial working.


eonarheim:
@dca123 Sorry I'm late getting back to you! I appreciate your PR and you reaching, you are correct! This is incorrect in excalibur v0.29.3!

We are about to ship the new version of excalibur where the tutorial will change completely to a new setup. See the code (and the readme that will be to tutorial) https://github.com/excaliburjs/sample-excalibird/


Thanks again! I'll close this, please feel free to reach out with anything else you find!

PR #3285, opened 2024-11-23
[#3284] perf: reduce allocations for GraphicsGroup draw calls

### LINKED ISSUE #3284: Reduce Allocations for GraphicsGroup Draw Calls
### Context

I am trying to render thousands of tiles to the screen. I use a `GraphicsGroup` to "chunk" draw calls. If I profile a static scene, the JS heap will grow before performing a GC that impacts frame rate:

image

I did a memory allocation profile over 15 seconds, and [`get localBounds`](https://github.com/excaliburjs/Excalibur/blob/9176e5b705cc8b851050d5e0395e58bdb29c13c0/src/engine/Graphics/GraphicsGroup.ts#L60) allocated about 60 MB.

Looking at that method, I notice a few allocations:

- `let bb = new BoundingBox();`
- `bb = member.localBounds.combine(bb);`
- `bb = graphic.localBounds.translate(pos).combine(bb);`

### Proposal

Is it possible to reduce the number of allocations of `get localBounds` by `this.members.length` by reusing `bb` in [`combine`](https://github.com/excaliburjs/Excalibur/blob/9176e5b705cc8b851050d5e0395e58bdb29c13c0/src/engine/Collision/BoundingBox.ts#L344)?

```typescript
public get localBounds(): BoundingBox {
  let bb = new BoundingBox();
  for (const member of this.members) {
    if (member instanceof Graphic) {
      member.localBounds.combine(bb, bb); // use bb as dest to avoid allocation
    } else {
      const { graphic, offset: pos, useBounds } = member;
      const shouldUseBounds = useBounds === undefined ? true : useBounds;
      if (graphic) {
        if (shouldUseBounds) {
          graphic.localBounds.translate(pos).combine(bb, bb); // use bb as dest to avoid allocation
        }
      } else {
        this._logger.warnOnce(`Graphics group member has an null or undefined graphic, member definition: ${JSON.stringify(member)}.`);
      }
    }
  }
  return bb;
}
```

blakearoberts:

==================

Closes #3284 

## Changes:

- reuse single cumulative bounding box for local bounds calculation of `GraphicsGroup` to reduce per-draw allocations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,7 +121,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - `ex.Engine.timeScale` values of 0 are now supported
 - `ex.Trigger` now supports all valid actor constructor parameters from `ex.ActorArgs` in addition to `ex.TriggerOptions`
 - `ex.Gif` can now handle default embedded GIF frame timings
-- New `ex.Screen.worldToPagePixelRatio` API that will return the ratio between excalibur pixels and the HTML pixels. 
+- New `ex.Screen.worldToPagePixelRatio` API that will return the ratio between excalibur pixels and the HTML pixels.
   * Additionally excalibur will now decorate the document root with this same value as a CSS variable `--ex-pixel-ratio`
   * Useful for scaling HTML UIs to match your game
     ```css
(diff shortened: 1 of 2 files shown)


blakearoberts:
I pulled this change into the same scene I was working with in #3284, and the heap looks much healthier!

image

CHANGELOG.md
```diff
@@ -215,6 +215,7 @@ are doing mtv adjustments during precollision.
   * `EventEmitter`s
   * `GraphicsSystem` entity iteration
   * `PointerSystem` entity iteration
+- Perf improvements to `GraphicsGroup` by reducing per draw allocations in bounds calculations
```
eonarheim, line 218: This is the only change we should keep on the changelog, we should revert the rest

eonarheim (changes requested):
@blakearoberts This looks great! I love that heap graph

2 asks:
* revert extra whitespace changes in changelog
* fix lint (the linter want's a `const bb`)

eonarheim (approved):

eonarheim:
@blakearoberts Thanks again! this was a great add to excalibur! ❤️

PR #3291, opened 2024-11-29
feat: Make RotateTo+RotationType functionality more available

### LINKED ISSUE #3288: Make RotateTo+RotationType functionality more available
### Context

I find  `RotateTo` action to be very useful, I often have the situation that objects have to orient themselves toward some place on the screen, and  is not a trivial task to determine the correct angle by which to rotate an object. For example often I need objects to turn towards the destination in the optimal way, using minimal possible rotation.  

The problem is that this handy functionality is only available with Actions. I believe this functionality should be generalized and moved to e.g. `Vector`. The `RotateTo` Action could then use the functionality from `Vector`.  

### Proposal

I would like to implement the following: 
* Copy `RotationType` enum to `Math` package, and use that one from now on. 
* Deprecate `RotationType` enum that is in `Actions` package, to be removed at next major release
* Inroduce `angleBetween` method to `Vector` class. 
* Use `Vector.angleBetween` also in `RotateTo` action. 

Here would be the example use case: 

```typescript
    const point = vec(100, 100)
    const destinationDirection = Math.PI / 4
    const angleToRotate = point.angleBetween(destinationDirection, RotationType.ShortestPath)
    expect(point.rotate(angleToRotate).toAngle()).toEqual(destinationDirection)
```

The idea for the method would be to return the angle that could be used directly as a parameter of `Vector.rotate()`, of course by applying some sort of gradual transition, otherwise the user could just set the rotation to `destinationDirection`.

isokissa:

==================

Closes #3288 

## Changes:

- Introduced `angleBetween` metdhod to `Vector` class. 
- fixed `canonicalizeAngle` to never return 2PI
- `RotationType` moved to `Math` 
- small documentation improvements
- added/modified tests to `Vector.rotate()` to illustrate the conventions


isokissa:
@eonarheim this is ready, waiting for your review and blessing :) Locally all tests pass, don't know how to trigger the check here on github.

eonarheim:
@isokissa Apologies I've been busy with other things, I'll give this a proper look today

isokissa:
Still stying problems in vector.ts, but it won't tell me what is wrong. 

And yeah, I ended up not using `lerpAngle`'s logic, I wrote `angleBetween` a bit simpler, IMHO.

isokissa:
Fixed formatting, now it should be 100% ready for review. @eonarheim

eonarheim:
@isokissa Thanks a ton! This is a great addition, much appreciated!

PR #3350, opened 2025-01-30
fix: Spritesheet Spacing Options, allowing Vectors

jyoung4242:
### SpriteSheet.ts
- modified  SpriteSheetSpacingDimensions originOffset and margin to accept optional Vectors
- modified fromImageSource to test for instance of Vector and handle it properly, while maintaining backward compatibility

### SpriteSheetSpec.ts
- added test: `can be created from with spacing using vectors`

### Updated docs entry on Spritesheets to note that Vectors are now allowed

Ran locally and verified functionality using both vec() and new Vector


eonarheim:
@jyoung4242 Thanks again!

Issue #3076, opened 2024-05-27
Pause/unpause Actors and Entities feature


eonarheim:
### Context

Sometimes it is useful to pause the update and/or draw of an actor without removing it from the scene to save on performance.

One examples might be NPCs that exist offscreen and don't need to run their logic until they are onscreen.

### Proposal

Add a new method (open to different names) `.pause()` or `.unpause()` that should leverage the `entity.active` property to suspend updates and/or drawing.

`.kill()` should be updated to use a different property internally.

eonarheim:
#3656

PR #3380, opened 2025-03-02
reat: support any, all and not component/tag filters on Query

mattjennings:
⚠️ Open to any feedback. I went back and forth on a few API designs, ultimately felt this was the simplest & safest without breaking existing behavior.

Improves `ex.Query` to support "any" and "not" filters for components, as well as all/any/not tags filters. The intention is for this to replace `ex.TagQuery`, but that is not (yet) in this PR.

With this change, this would much easily allow systems to exclude entities with tags. For example, GraphicsSystem could now exclude any Entity with an `ex.offscreen` tag instead of manually checking in its update. This could also be used for #3076 by adding a tag to an Entity while paused that excludes it from systems.

I also removed the restriction of throwing on empty component queries, as I think it could make sense to have a system that applies for every Entity. I can change this, though.

==================

Closes #

## Changes:

- Queries can now take additional options to filter in/out by components or tags.

```ts
const query = new Query({
  // all fields are optional
  components: {
    all: [ComponentA, ComponentB] as const, // important for type safety!
    any: [ComponentC, ComponentD] as const, // important for type safety!
    not: [ComponentE]
  },
  tags: {
    all: ['tagA', 'tagB'],
    any: ['tagC', 'tagD'],
    not: ['tagE']
  }
})

// previous constructor type still works and is shorthand for components.all
new Query([ComponentA, ComponentB] as const)
```

- Queries can now match all entities by specifying no filters

```ts
const query = new Query({})
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -83,6 +83,13 @@
         "allowTernary": true
       }
     ],
+    "no-restricted-syntax": [
+      "error",
+      {
+        "message": "Do not commit tests with 'fit'",
+        "selector": "CallExpression[callee.name='fit']"
+      }
+    ],
     "jsdoc/require-param": 0,
     "jsdoc/require-param-description": 0,
     "jsdoc/require-param-type": 0,
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -87,7 +87,7 @@
       "error",
       {
         "message": "Do not commit tests with 'fit'",
-        "selector": "CallExpression[callee.name='fit']"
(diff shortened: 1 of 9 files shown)


mattjennings:
This was partially inspired by [Unity's EntityQueryBuilder](https://docs.unity3d.com/Packages/com.unity.entities@1.0/api/Unity.Entities.EntityQueryBuilder.html) although I didn't include an actual builder API.

eonarheim (approved):
@mattjennings No notes! this looks awesome

PR #3381, opened 2025-03-03
chore: migrate tests to vitest

mattjennings:
TODO:

- migrate jasmine memory usage reporters to vitest

==================

Closes #


src/spec/MaterialRendererSpec.vitest.ts
```diff
@@ -439,7 +436,10 @@ describe('A Material', () => {
 
     expect(material).toBeDefined();
     expect(logger.warn).toHaveBeenCalledWith(
-      'Max number texture slots 31 have been reached for material "test", no more textures will be uploaded due to hardware constraints.'
+      expect.stringMatching(
+        // this was 31 before, but since moving to vite it's 15. Is there a difference in max texture slots from chromium vs old karma browser?
+        /Max number texture slots (\d+) have been reached for material "test", no more textures will be uploaded due to hardware constraints\./
+      )
```
mattjennings, line 442: @eonarheim i'm unsure why max texture slots would have changed from 31 to 15 from karma chrome -> playwright chromium

eonarheim: It might be down to chrome's configuration to chromiums? It also could be based on the driver/hardware on your system

mattjennings:
@eonarheim at long last the tests are passing. let me know what you think of how i handled the TextSpec tests.

Also IIRC this still lacks the memory tracking and other custom karma reporting stuff. would you want that done in this PR?

mattjennings:
managed to migrate the engine leak & memory reporters, although it took a bit of rigamarole. Mainly because, as opposed to karma, the reporter runs in node and not the browser. So I have to track the data needed separately in some global hooks _within_ the browser environment, which the reporter then reads and creates the logs if needed.

However, it seems the memory tracking is regularly reporting >1 mb, so I'm not sure if it's tracking properly or if things have changed with the new browsers. My only theory is the `afterEach` hooks where the memory tracking is setup to record is ran before any of the specific test's `afterEach`, which could be doing cleanup / releasing memory. Implementation is done in `src/spec/vitest/__reporters__/memory.ts` and `src/spec/vitest/__reporters/memory.setup.ts`

mattjennings:
There appears to be some flakiness with the windows TextSpec tests 😵‍💫 https://github.com/excaliburjs/Excalibur/actions/runs/14814873651/job/41595236051

On the first run the text was just rendered differently, and on the retry they passed but a WebGL test timed out. I wonder if we're hitting some WebGL context limit similar to what we saw on macOS Chrome by opening up a ton of contexts...

PR #3404, opened 2025-04-02
feat: added Graph module to Math 

jyoung4242:
Open to feedback on the API here.   

Features added, Graphs, Nodes, Edges, Vertices...
Pathfinding and graph traversal functions
QOL methods for using graphs...

Special reqeust for feedback on the Astar vs Dijkstra's pathfinding.... the API is subtle different where Astar uses spatial PositionNodes and Dijkstra's uses weighted edges.

Open for ideas on this.

3 Files added 
- Graph.ts in /Math
- GraphSpec.ts in /src/spec
- 07-graph.mdx in /docs


src/engine/Math/Graph.ts
```diff
@@ -0,0 +1,805 @@
...
+export type Vertex<T> = Node<T>;
+
+class GraphUUId {
+  static generateUUID(): G_UUID {
+    const uuid = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx';
+    const generatedUuid = uuid.replace(/[xy]/g, function (c) {
+      const r = (Math.random() * 16) | 0;
+      const v = c === 'x' ? r : (r & 0x3) | 0x8;
+      return v.toString(16);
+    });
```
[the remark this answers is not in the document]
jyoung4242: Replacing Math.random() wth Ex' Random class

package-lock.json
```diff
@@ -5,6 +5,79 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "excalibur",
```
eonarheim, line 8: I think this file will need to be reverted too, makes reference to your local Meet/Excalibur

site/docs/09-math/07-graph.mdx
```diff
@@ -0,0 +1,143 @@
...
+const nodeB = graph.addNode("B", new Vector(5, 10));
+const nodeC = graph.addNode("C", new Vector(10, 5));
+```
+
+### Edge Properties
+
+Edges connect nodes and can have properties:
+
+weight: Numeric value representing distance or cost (default: 0)
+directed: Whether the edge is one-way or bidirectional (default: true)
```
eonarheim, line 83: I'm curious if directed should default to false? No strong data, but I generally use bidirectional edges when I'm building nav mesh type situations.

Also is the direction A->B in `.addEdge(nodeA, nodeB, ...)`?

src/engine/Math/graph.ts
```diff
@@ -0,0 +1,809 @@
...
+    super(data);
+    this.pos = pos;
+  }
+}
+
+/**
+ * Alias for a node in a graph, representing a vertex in a geometric context.
+ * @template T The type of data stored in this vertex.
+ */
+export type Vertex<T> = Node<T>;
```
eonarheim, line 792: Should we stick to either node or vertex? I'm a little worried the synonym could be confusing.

eonarheim: It seems there are duplicate methods with different names

eonarheim (changes requested):
Looks really awesome, just a couple notes.

Question, do we want to deprecate the pathfinding plugin since we are moving this functionality into core? https://github.com/excaliburjs/excalibur-pathfinding/

site/docs/09-math/07-graph.mdx
jyoung4242: changing default is easy, and i can do some parameter renaming to accommodate direction

src/engine/Math/graph.ts
jyoung4242: I would actually prefer to keep it to just nodes, and remove vertex

package-lock.json
jyoung4242: might need some guidance on this, i've never touched a package-lock.json

it's usually managed by node, right?

jyoung4242:
@eonarheim i'm considering modifying how the positions work, and use it to dictate edge weight
that way if you set weights on an edge OR position, astar will still work

package-lock.json
jyoung4242: i've deleted my package-lock and redid a `npm install` and it keeps coming up with local references in my package-lock...
@eonarheim

jyoung4242:
@eonarheim we should deprecate the pathfinding plugin in, but I would think we should recreate the sample application using the core native Graph library before deprecating

site/docs/09-math/07-graph.mdx
jyoung4242: last push established default as bi-directional, and two edges are created A-> B and B->A

src/engine/Math/Graph.ts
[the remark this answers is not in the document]
jyoung4242: This is resolve in update.

jyoung4242:
We are going to renew this PR after the vitest migration and a fresh branch of Ex

PR #3425, opened 2025-05-03
#3424 feat: Add maxVel to MotionComponent

### LINKED ISSUE #3424: MotionComponent for Arcade physics should include `maxVel`
### Context

Arcade physics solving currently has no native way to provide a terminal velocity in either direction. The Phaser framework provides this feature, and it's fairly useful as a way to have more control over player motion and game feel. It also has some precedent as a low-cost solution for the 'tunneling' problem that can occur with fast-moving objects.

### Proposal

1. Modify `MotionComponent` to include a `maxVel` Vector that can independently constrain motion on the X and Y axes
    1. this Vector would be optional, and default to `Infinity` on both axes
3. Modify the `EulerIntegrator` so that immediately after computing acceleration's effect on velocity, and before applying the velocity to the position, the velocity is clamped based on the provided `motion.maxVel` x and y components

Potential follow ups:
- a `maxSpeed` scalar-based value that does not split the components out- it only clamps the resulting velocity vector's magnitude while preserving direction
    - TBD how this property will work if `maxVel` is allowed at the same time

kfalicov:

==================

Closes #3424 

## Changes:

- Adds `maxVel` attribute to `MotionComponent`, which clamps velocity on separated X and Y axes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fixed issue where the Actor.color did not respect being set
 - Fixed offscreen culling issue when using parallax on TileMaps
 - Fixed division by 0 when timescale is 0 in actions
 - Fixed `onTransition` on the initial scene transition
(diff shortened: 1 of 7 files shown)


mattjennings (approved):
Thanks for the PR! Looks good!

PR #3434, opened 2025-05-31
Use vite for excalibur build

mattjennings:

==================

I've left the old build for now as reference while I work on this, but I've replaced `npm run build` with the vite build so I can see if CI passes. 

## Stats for nerds

results are taken from an M1 Max Macbook Pro running Linux

### Webpack

Build time 18.31s

```
800K	./esm/excalibur.min.js
1.1M	./esm/excalibur.js
800K	./esm/excalibur.min.development.js
1.1M	./esm/excalibur.development.js
536K	./dist/excalibur.min.js
1.4M	./dist/excalibur.js
1.8M	./dist/excalibur.min.development.js
1.8M	./dist/excalibur.development.js
```

### Vite

Build time: 7.67s

```
800K	./esm/excalibur.min.js
1.1M	./esm/excalibur.js
800K	./esm/excalibur.min.development.js
1.1M	./esm/excalibur.development.js
524K	./dist/excalibur.min.js
1.1M	./dist/excalibur.js
524K	./dist/excalibur.min.development.js
1.1M	./dist/excalibur.development.js
```


kamranayub:
@mattjennings this looks sweet. Just curious but did you take a look at [tsdown](https://tsdown.dev/) since it's meant for libraries? Maybe the config is not quite there yet where we need.

[Declarations](https://tsdown.dev/options/dts) are also supported in tsdown through rolldown-dts.

Also I know TSGO will be integrated into tsdown as soon as it's compatible 🚀 I'm sure it'll be in Vite too.

mattjennings:
I haven't, mainly because I wanted to unify the config that we already need to use with Vitest, and I use vite's lib mode to bundle libraries myself. What particular parts make you feel Vite's config isn't there yet?

I know that rolldown is going to be integrated into Vite when it's ready, but it looks like according to tsdown's landing page tsdown itself will too. So maybe some ergonomics will improve there.

Vite can output dts files with a plugin but it would be run for every different kind of build (development/production x minified/unminified). We only need to output the declarations once, so it was fine to just run tsc manually after, similar to what we were doing before.

mattjennings:
appveyor is failing with that error because it's using Node 20, seems we need to be at least on Node 22 because of something that vite-plugin-static-copy does

eonarheim:
@mattjennings I think I've finally killed the appveyor runner

sandbox/vite.config.ts
```diff
@@ -0,0 +1,34 @@
...
+      }
+    },
+    plugins: [
+      {
+        // brutally inject every js/ts file with an ex import at the top
+        name: 'load-excalibur',
+        transform(code, id) {
+          const isJsFile = id.endsWith('.ts') || id.endsWith('.js');
+          if (isJsFile) {
+            code = `import * as ex from '@excalibur'\n${code}`;
```
eonarheim, line 23: OH THIS IS COOL, had no idea you could do this

vite.config.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  // eslint-disable-next-line no-console
+  console.log(`[version] ${version}.js`);
+
+  return mergeConfig(commonConfig, {
+    build: {
+      target: 'es2018',
+      outDir: path.resolve(__dirname, 'build'),
+      emptyOutDir: false,
+      lib: {
```
eonarheim, line 27: chef's kiss

PR #3446, opened 2025-06-02
feat(Clock): Add API to clearSchedule

kamranayub:

==================

Closes #

## Changes:

- feat(Clock): Add `clearSchedule(scheduleId: number)` API to clear a scheduled callback

### DIFF
--- a/src/engine/Util/Clock.ts
+++ b/src/engine/Util/Clock.ts
@@ -3,6 +3,11 @@ import { FpsSampler } from './Fps';
 
 export type ScheduledCallbackTiming = 'preframe' | 'postframe' | 'preupdate' | 'postupdate' | 'predraw' | 'postdraw';
 
+/**
+ * Unique identifier for a scheduled callback
+ */
+export type ScheduleId = number;
+
 export interface ClockOptions {
   /**
    * Define the function you'd like the clock to tick when it is started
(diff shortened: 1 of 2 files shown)


src/engine/Util/Clock.ts
```diff
@@ -94,13 +99,26 @@ export abstract class Clock {
...
+    this._scheduledCbs.push([id, cb, scheduledTime, timing]);
+    return id;
   }
 
+  /**
+   * Clears a scheduled callback using the ID returned from {@apilink schedule}
+   * @param id The ID of the scheduled callback to clear
+   */
+  public clearSchedule(id: ScheduleId): void {
+    const index = this._scheduledCbs.findIndex(([scheduleId]) => scheduleId === id);
```
kamranayub, line 117: only concern: tight loop performance? if user clears every frame or something...

to fix would need new data structure?

kamranayub:
I should add corresponding docs

kamranayub:
Appveyor failures are just text graphics which are known to be a bit flaky (they failed for me locally even on the first test run but then passed)

src/engine/Util/Clock.ts
eonarheim: Truth, if there were a LOT of scheduled callbacks this could start to be bad. I vote we keep this as is until it's a problem, my intuition says we'll be in the 10's of callbacks which will be low overhead to scan

src/engine/Util/Clock.ts
```diff
@@ -109,8 +127,9 @@ export abstract class Clock {
   public __runScheduledCbs(timing: ScheduledCallbackTiming = 'preframe') {
     // walk backwards to delete items as we loop
     for (let i = this._scheduledCbs.length - 1; i > -1; i--) {
-      if (timing === this._scheduledCbs[i][2] && this._scheduledCbs[i][1] <= this._totalElapsed) {
-        this._scheduledCbs[i][0](this._elapsed);
+      const [_, callback, scheduledTime, callbackTiming] = this._scheduledCbs[i];
```
eonarheim, line 130: Love this change, past Erik really didn't want to read this 🤦‍♂️

eonarheim (approved):

eonarheim:
@kamranayub Thanks again! this is a valuable addition to the clock api

PR #3470, opened 2025-07-06
fix: ex.Random improve unique seeding on constructor w/o provided seed

jyoung4242:
3 files changed

- Random.ts
- 07-random.mdx
- RandomSpec.ts

# Random.ts
added static intialSeed form Date.now() and an incrementing offset.
also, this ensures that the actual seed used is available property
moved public seed  to private _seed
added getter for seed
in constructor(), if no seed provided it pulls from the 'next' seed in the incrementing list

# 07-random.mdx
updated documentation to reflect new functionality

# RandomSpec.ts
-added 2 tests
- will have a seed even if not passed into constructor
- can have sequential random instances with unique seeds

# Checklist for PR
- added tests
- all tests pass locally
- created test project to exercise new logic with favorable results
- updated docs

### DIFF
--- a/src/engine/Math/Random.ts
+++ b/src/engine/Math/Random.ts
@@ -59,12 +59,7 @@ export class Random {
     this._mt = new Array<number>(this._n);
     // need to mask to support higher bit machines
 
-    const dynamicSeed = () => {
-      seedOffset++;
-      return initialSeed + seedOffset;
-    };
-
-    this._mt[0] = (seed || dynamicSeed()) >>> 0;
+    this._mt[0] = (seed || initialSeed + seedOffset++) >>> 0;
     this._seed = this._mt[0];
 
     for (let i = 1; i < this._n; i++) {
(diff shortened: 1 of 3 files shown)


src/engine/Math/Random.ts
```diff
@@ -48,14 +50,22 @@ export class Random {
...
     this._mt = new Array<number>(this._n);
     // need to mask to support higher bit machines
-    this._mt[0] = (seed || Date.now()) >>> 0;
+
+    const dynamicSeed = () => {
+      seedOffset++;
+      return initialSeed + seedOffset;
+    };
+
+    this._mt[0] = (seed || dynamicSeed()) >>> 0;
```
mattjennings, line 67: this is a nitpick but creating an anonymous function here could be detrimental to performance if we're in a hot path. Would rather do `(seed || (initialSeed + seedOffset++))`.

jyoung4242: Excellent, will fix

eonarheim (approved):

PR #3474, opened 2025-07-29
Improved flappy bird tutorial

Clemix37:
Whene reading Flappy Bird tutorial, on the 12th step, no explanation and no code are added to show the use of animation when bird is going down or up.

## Changes:

- Added new code to put everything together
- Explained and displayed where to use an animation and why resetting some is important


eonarheim (approved):
@Clemix37 Great addition! Many thanks

Issue #3413, opened 2025-04-29
Consistency: Adding optional destination/output arguments to all applicable Vector functions


Autsider666:
### Context

I ran into a [great presentation about increasing Vector math performance](https://media.tojicode.com/sfjs-vectors/) that advises to add output arguments to functions to prevent the creation of unneeded objects. 

### Proposal

`EX.Vector` already has some functions (`scale`, `add`, `sub`, `rotate` and `clone`) that have an optional `dest` argument, so I'm suggesting to consistently add an optional `dest`argument to the following functions:
-  `min`
- `max`
- `normalize`
- `cross`
- `perpendicular`

#### Optional
- Add an extra argument for [the fallback anchor used in `rotate`](https://github.com/excaliburjs/Excalibur/blob/main/src/engine/Math/vector.ts#L427)
- Use `destination` or `output` instead of the abbreviated `dest` currently used.
- Update the documentation to inform developers of the potential performance gains from using this argument when using `Vector`. (Even more optional: Add code examples to show the differences)

mattjennings:
Hm I didn't realize this was a thing. My personal thought on this is I don't like polluting all the methods with a param like this because it can already be chained onto any vector with the `clone` method

```ts
vec.rotate(...).clone(dest)
```

I would argue this should be the convention instead - or, implement all these methods as a static that target vectors. (Then, the Vector instance methods simply call the static ones with itself as the target)

```ts
ex.Vector.rotate(vec, ...)
```

Autsider666:
> Hm I didn't realize this was a thing. My personal thought on this is I don't like polluting all the methods with a param like this because it can already be chained onto any vector with the `clone` method
> 
> vec.rotate(...).clone(dest)

Correct me if I'm wrong, but calling `rotate` without a `dest` argument would still create a new Vector object that would clutter up memory the moment it's returned (if called a lot of times every frame), even if you chain `clone` after it?

mattjennings:
You're right, certain methods mutate itself while others create a new one and return it. IMO the consistency should be improved here too, with every operation mutating itself and instantiating new ones should be opt in with `clone`.  This would be a big breaking change though.

Autsider666:
> I would argue this should be the convention and instead _remove_ any `dest` params - or, implement all these methods as a static that target vectors. (Then, the Vector instance methods simply call the static ones with itself as the target)
> 
> ex.Vector.rotate(vec, ...)
[...]

PR #3491, opened 2025-08-04
Feature: Added lerp method to vector class

jyoung4242:
Added lerp method to Vector class, with clamped t for time [0,1]

Made a comprehensive update to the Vector documentation, spelled out more about what's available with Vectors

added lerptests to Vector spec, 
can lerp ...
can clamp lerp time...

broke AlgebraSpec.ts into VectorSpect.ts, LinesSpect.ts, ProjectionsSpec.ts, and RaysSpec.ts

### DIFF
--- a/site/docs/09-math/07-vector.mdx
+++ b/site/docs/09-math/07-vector.mdx
@@ -61,3 +61,169 @@ console.log(anotherPoint.toString()) // "(50, 50)"
 ```
 
 Notice how both `point` and `samePoint` share the same vector reference, so using `setTo` mutates the vector. Use `clone` to ensure you are not changing vectors unexpectedly!
+
+### Static Methods Available
+
+#### `Vector.fromAngle(angle: number): Vector`
+Creates a unit vector from an angle in radians.
+
+```typescript
+const rightVector = Vector.fromAngle(0); // (1, 0)
+const upVector = Vector.fromAngle(-Math.PI / 2); // (0, -1)
+```
+
+#### `Vector.distance(vec1: Vector, vec2: Vector): number`
+Calculates the distance between two vectors.
+
+```typescript
+const distance = Vector.distance(vec(0, 0), vec(3, 4)); // 5
+```
+
+#### `Vector.isValid(vec: Vector): boolean`
+Checks if a vector is valid (not null, undefined, or containing NaN/Infinity).
+
+```typescript
+const isValid = Vector.isValid(new Vector(1, 2)); // true
+const isInvalid = Vector.isValid(new Vector(NaN, 1)); // false
+```
+
+#### `Vector.min(vec1: Vector, vec2: Vector): Vector`
+Returns a vector with the minimum x and y components.
+
+#### `Vector.max(vec1: Vector, vec2: Vector): Vector`
+Returns a vector with the maximum x and y components.
+
+### Instance Properties
+
+#### `x: number` and `y: number`
+Get or set the vector components. **Warning**: Setting these properties mutates the vector.
+
+#### `magnitude: number`
+Get or set the vector's magnitude (length). Setting this normalizes the vector and scales it to the new magnitude.
+
+### Core Operations
+
+#### `add(v: Vector, dest?: Vector): Vector`
+Adds another vector to this one.
+
+```typescript
+const result = vec(1, 2).add(vec(3, 4)); // (4, 6)
+```
+
+#### `sub(v: Vector, dest?: Vector): Vector`
+Subtracts another vector from this one.
+
+```typescript
+const result = vec(5, 3).sub(vec(2, 1)); // (3, 2)
+```
+
+#### `scale(size: number | Vector, dest?: Vector): Vector`
+Scales the vector by a scalar or component-wise by another vector.
+
+```typescript
+const scaled = vec(2, 3).scale(2); // (4, 6)
+const componentScaled = vec(2, 3).scale(vec(2, 0.5)); // (4, 1.5)
+```
+
+#### `normalize(): Vector`
+Returns a unit vector in the same direction.
+
+```typescript
+const unit = vec(3, 4).normalize(); // (0.6, 0.8)
+```
+
+### Distance and Magnitude
+
+#### `distance(v?: Vector): number`
+Returns the distance to another vector, or magnitude if no vector provided.
+
+```typescript
+const dist = vec(0, 0).distance(vec(3, 4)); // 5
+const magnitude = vec(3, 4).distance(); // 5
+```
+
+#### `squareDistance(v?: Vector): number`
+Returns the squared distance (faster than distance for comparisons).
+
+### Vector Math
+
+#### `dot(v: Vector): number`
+Calculates the dot product with another vector.
+
+```typescript
+const dotProduct = vec(1, 2).dot(vec(3, 4)); // 11
+```
+
+#### `cross(v: Vector | number): number | Vector`
+Calculates the cross product. With a vector, returns a scalar. With a scalar, returns a vector.
+
+```typescript
+const crossScalar = vec(1, 2).cross(vec(3, 4)); // -2
+const crossVector = vec(1, 2).cross(5); // (10, -5)
+```
+
+#### `perpendicular(): Vector`
+Returns a perpendicular vector (rotated 90 degrees).
+
+```typescript
+const perp = vec(1, 0).perpendicular(); // (0, -1)
+```
+
+#### `normal(): Vector`
+Returns the normalized perpendicular vector.
+
+### Rotation and Angles
+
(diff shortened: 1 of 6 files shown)


site/docs/09-math/07-vector.mdx
```diff
@@ -61,3 +61,169 @@ console.log(anotherPoint.toString()) // "(50, 50)"
...
+Clamps the vector's magnitude to a maximum value (mutates the vector).
+
+### Mutating Operations
+
+**⚠️ Warning**: These methods modify the original vector and should be used carefully.
+
+- `addEqual(v: Vector)`: Adds and mutates
+- `subEqual(v: Vector)`: Subtracts and mutates  
+- `scaleEqual(size: number)`: Scales and mutates
+- `setTo(x: number, y: number)`: Sets components and mutates
```
mattjennings, line 229: @eonarheim any thoughts to my [proposal from way back here](https://github.com/excaliburjs/Excalibur/issues/3413#issuecomment-2845493717)? The fact this section is probably needed maybe points to needing a more intuitive convention.

mattjennings (approved):

site/docs/09-math/07-vector.mdx
```diff
@@ -61,3 +61,169 @@ console.log(anotherPoint.toString()) // "(50, 50)"
 ```
 
 Notice how both `point` and `samePoint` share the same vector reference, so using `setTo` mutates the vector. Use `clone` to ensure you are not changing vectors unexpectedly!
+
+### Static Methods Available
+
+#### `Vector.fromAngle(angle: number): Vector`
+Creates a unit vector from an angle in radians.
+
+```typescript
```
kamranayub, line 70: Would you mind using Shiki for these snippets? I'd love for all the snippets in the docs to be migrated to Twoslash snippets for type definitions and so they get compiled/type-checked in CI. You can reference the Events or Actors docs for how to set it up.

kamranayub (changes requested):

site/docs/09-math/07-vector.mdx
```diff
@@ -61,3 +61,169 @@ console.log(anotherPoint.toString()) // "(50, 50)"
...
+
+#### `negate(): Vector`
+Returns the negated vector.
+
+#### `clampMagnitude(magnitude: number): Vector`
+Clamps the vector's magnitude to a maximum value (mutates the vector).
+
+### Mutating Operations
+
+**⚠️ Warning**: These methods modify the original vector and should be used carefully.
```
kamranayub, line 224: This should use the admonition syntax:

```suggestion
:::warning

 These methods modify the original vector and should be used carefully.

:::
```

site/docs/09-math/07-vector.mdx
```diff
@@ -61,3 +61,169 @@ console.log(anotherPoint.toString()) // "(50, 50)"
 ```
 
 Notice how both `point` and `samePoint` share the same vector reference, so using `setTo` mutates the vector. Use `clone` to ensure you are not changing vectors unexpectedly!
+
+### Static Methods Available
```
kamranayub, line 65: Including the APIs here in the docs is nice for seeing it inline, but I'm also concerned this duplicates effort/information with the Typedoc API docs. If the signatures change, people need to remember to update them here which is what we want to avoid.

1. Is the same information (with examples) in the JSDoc for each of the methods listed here? If not, I think it should be.
2. Could we instead maybe link to each of them versus including them all here?

It makes me think that it would be handy to have a way to "include" API docs in MDX files here.

kamranayub (commented):
Some other comments; I am not 100% sure we should duplicate signature information in MDX docs.

site/docs/09-math/07-vector.mdx
jyoung4242: i understand the why behind this... will update to links

site/docs/09-math/07-vector.mdx
jyoung4242: this makes perfect sense, will update

jyoung4242: done

eonarheim:
@jyoung4242 Cloudflare pages build error

<pre class="c_lz c_bw c_la c_ip c_ig c_ih c_ds c_pe c_pf c_pg c_ph c_hh c_hi c_mx c_by c_pi" style="box-sizing: border-box; border-radius: 0px; color: rgb(49, 49, 49); display: block; font-size: 12px; margin: 0px; overflow: auto; padding: 16px 32px; width: 1262px; background-color: rgb(242, 242, 242); border-width: 0px 0px 1px; border-style: solid; border-color: rgb(217, 217, 217); border-image: initial; font-family: monaco, courier, monospace; height: 500px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">
[ERROR] Client bundle compiled with errors therefore further build is impossible.
--
17:56:54.161 | Error: MDX compilation failed for file "/opt/buildhome/repo/site/docs/09-math/07-vector.mdx"
17:56:54.162 | Cause:
17:56:54.162 | ## Errors were thrown in the sample, but not included in an errors tag
17:56:54.162 |  
17:56:54.162 | These errors were not marked as being expected: 2339.
17:56:54.162 | Expected: // @errors: 2339
17:56:54.162 |  
17:56:54.162 | Compiler Errors:
17:56:54.162 |  
17:56:54.162 | index.ts
17:56:54.162 | [2339] 208 - Property 'toString' does not exist on type 'void'.
17:56:54.162 | [2339] 257 - Property 'toString' does not exist on type 'void'.
17:56:54.162 | Details:
17:56:54.163 | {
17:56:54.163 | "title": "Errors were thrown in the sample, but not included in an errors tag",
17:56:54.163 | "description": "These errors were not marked as being expected: 2339. \nExpected: // @errors: 2339",
17:56:54.163 | "recommendation": "Compiler Errors:\n\nindex.ts\n  [2339] 208 - Property 'toString' does not exist on type 'void'.\n  [2339] 257 - Property 'toString' does not exist on type 'void'.",
17:56:54.163 | "code": "## Code\n\n'''ts\nimport { vec } from 'excalibur'\n\nconst point = vec(0, 10)\nconst samePoint = point.setTo(8, 8)\nconst anotherPoint = point.clone().setTo(50, 50)\n\nconsole.log(point.toString()) // \"(8, 8)\"\nconsole.log(samePoint.toString()) // \"(8, 8)\"\nconsole.log(anotherPoint.toString()) // \"(50, 50)\"\n'''"
17:56:54.163 | }
17:56:54.163 | Error:
17:56:54.163 | ## Errors were thrown in the sample, but not included in an errors tag
17:56:54.163 |  
17:56:54.170 | These errors were not marked as being expected: 2339.
17:56:54.170 | Expected: // @errors: 2339
17:56:54.170 |  
17:56:54.170 | Compiler Errors:
17:56:54.170 |  
17:56:54.170 | index.ts
17:56:54.170 | [2339] 208 - Property 'toString' does not exist on type 'void'.
17:56:54.170 | [2339] 257 - Property 'toString' does not exist on type 'void'.
17:56:54.313 | Failed: Error while executing user command. Exited with error code: 1
17:56:54.322 | Failed: build command exited with code: 1
17:56:55.416 | Failed: error occurred while running build command

</pre>


Digging into the snippet, I think the issue is that `point.setTo(8, 8)` doesn't return a Vector

kamranayub:
Hey, that's why twoslash is great! 😄

site/docs/09-math/07-vector.mdx
```diff
@@ -62,168 +62,21 @@ console.log(anotherPoint.toString()) // "(50, 50)"
...
-const dotProduct = vec(1, 2).dot(vec(3, 4)); // 11
-```
-
-#### `cross(v: Vector | number): number | Vector`
-Calculates the cross product. With a vector, returns a scalar. With a scalar, returns a vector.
-
-```typescript
-const crossScalar = vec(1, 2).cross(vec(3, 4)); // -2
-const crossVector = vec(1, 2).cross(5); // (10, -5)
+```sh twoslash
```
kamranayub, line 71: If the snippet isn't meant to be copied or referenced (or isn't TypeScript), you don't need to make it a twoslash directive. 
```suggestion
```sh
```

kamranayub (approved):
Once build errors are addressed it LGTM

kamranayub:
@jyoung4242 fixed up the snippet(s) and wrapped the "Notice..." paragraph in a tip to call it out.

Bonus: twoslash not only caught a bug with the snippet, it let us intentionally call it out:

image

PR #3500, opened 2025-08-20
Add call to super.update

SanjoSolutions:


eonarheim (approved):
@SanjoSolutions ACK! good catch thanks for the fix

PR #3506, opened 2025-08-29
fix(Engine): Remove canvas only when it has been created

SanjoSolutions:
I have noticed this when I have embedded the game inside React with React strict mode on and using the canvasElement option.


eonarheim:
@SanjoSolutions good find! Definitely a bug

eonarheim:
@SanjoSolutions thanks again!

PR #3507, opened 2025-08-31
feat: Add Dev Container support

kamranayub:

- [x] :pushpin: ~~issue~~ Discord discussion exists in github for these changes

==================

Dev Containers are a VS Code feature that make it easier for contributors to set up a project in a dedicated environment using Docker.

## Changes

- Deps: Update to `playwright@1.55.1`
- Deps: Added Vitest UI
- Add `devcontainer.json` configuration
- Update README for contributors
- Remove crufty old Dockerfile/compose

## Dev Container Lifecycle

- Based on MS Playwright image (keep in sync with Playwright version, we will get an error if it's not)
- Install dependencies (`npm install`)
- Install Playwright browsers (if missing) (`npx playwright install`)
- Start virtual Xvfb server for headed tests

## Tested

- [x] Unit: headed
- [x] Unit: headless
- [x] Visual: headed
- [x] Visual: headless
- [x] Vitest UI
- [x] Sandbox
- [x] Storybook
- [x] Docs


mattjennings:
FWIW I ran our tests both headfully and headlessly on mac with chromium and only had the following 2 tests fail:

>  FAIL   chromium  src/spec/vitest/DebugSpec.ts > Debug draw static > can draw lines
Error: Expected image to match within 99%, but only matched 96.00%. 4 pixels different.

>  FAIL   chromium  src/spec/vitest/PolygonSpec.ts > A Polygon Graphic > can be drawn
Error: Expected image to match within 99.3%, but only matched 99.01%. 99 pixels different.

BaconSoap:
A small  point in favor - I had a bit of trouble with the setup/instructions (node version in the Dockerfile for Mac dev that the readme mentions is 16, vs LTS 22), and determining if those same two test fails were from my environment. Having a devcontainer would both make that initial setup for drive-by contributors a bit more straightforward/unified, remove the need for `nvm` shuffling, and I _think_ would make it easier to fully edit/test in codespaces on github.

And devcontainers work nicely in jetbrains tools, too 😸

kamranayub:
Thanks @BaconSoap 💯 

I also noticed @mattjennings that the Dockerfile sets up Node 16. That's way out of support. I'm wondering if that was on purpose or not? It would be much better to make the base image an official Node image if possible but wasn't sure if this was due to Playwright dependencies or something. 

The dev container will start with Node 16 and then uses nvm to set up LTS (22). So it "works" but setup would be much faster if the Dockerfile baked in LTS.

This PR can stay in draft until the Vitest suite is segmented, which Matt is doing (thanks Matt!).

Once that merges, I can update and retest this dev container setup.

eonarheim:
@kamranayub Definitely not on purpose, I'd say this is just a lapse over time
> I also noticed @mattjennings that the Dockerfile sets up Node 16. That's way out of support. I'm wondering if that was on purpose or not? It would be much better to make the base image an official Node image if possible but wasn't sure if this was due to Playwright dependencies or something.

kamranayub:
> @kamranayub Definitely not on purpose, I'd say this is just a lapse over time
> > I also noticed @mattjennings that the Dockerfile sets up Node 16. That's way out of support. I'm wondering if that was on purpose or not? It would be much better to make the base image an official Node image if possible but wasn't sure if this was due to Playwright dependencies or something.
> 

Okay, this is something I could probably take a look at streamlining too then

mattjennings:
I'd vote to just remove the dockerfile since none of us use it and dev containers seem to be the better fit for development.

mattjennings:
oops, didn't realize the dev container used that same docker file. Whatever we choose, there's no restriction I'm aware of to use anything older than LTS, and if there is I'd rather fix that so we _can_ use LTS.

The vitest changes have been merged in @kamranayub so let's see if that resolved the issues here

mattjennings:
@kamranayub I've pulled this PR locally, rebased on master and added a few changes to how the container is setup for playwright. I was able to get the tests running & passing on my M4 mac, however the tests need to run headlessly to run in a container. By default, unless CI env is true, vitest will open a page in a browser to display the running tests. This is really nice to have when developing, but will not work when executed inside of a container. So `npm run test` will fail unless you pass `-- --browser.headless=true`.

So I suppose if we want this to just work for container users, we'll need to default to headless. Then we could provide npm scripts for running with the UI mode.

--- 

edit: actually I think we can still run headlessly and just print the localhost URL to view the UI mode in the browser yourself... this would be the best of both worlds. taking a look...

mattjennings:
pushed my branch here on [`matt/dev-container`](https://github.com/excaliburjs/Excalibur/commit/4e6996c2aa613a962774e177957a6f9bd1822341)

let me know if it works out of box for you (may need to do `docker compose down` first to recreate the containers). There will be an xdg error printed when running `npm run test` but this is because vitest tries to launch the URL - it's harmless. It'll print the URL so you can view it manually instead.

if it looks good, it's targeting this branch so we can merge it in here & then merge into main.

eonarheim:
@kamranayub So excited for this :D

kamranayub:
Good news!

* No longer need crufty Node 16 `docker-compose` setup
* It works with headed and headless tests!
* It "just works" on my Mac
* Ports are all auto-forwarded when servers start

mattjennings (changes requested):
for `test:ci` script we should disable --ui, will probably give a slight speed boost and it has no purpose there. otherwise the rest LGTM!

mattjennings (approved):

PR #3509, opened 2025-09-01
feat: Add Camera methods to control strategy priority

chrisk-7777:

==================

Closes https://github.com/excaliburjs/excalibur-tiled/issues/725 (excalibur-tiled)

## Changes:

- Added two methods to Camera: `addStrategies` and `setStrategies`.
- Added getter property, `strategies`,  to expose current Camera strategies (so they can be merged via `setStrategies` like `camera.setStrategies([new CustomStrat(), ...camera.strategies])`
- Added additional documentation on custom camera strategies and multiple camera strategies


mattjennings (commented):
good stuff! my only thought is maybe we should just have `addStrategy` take multiple params instead of having two methods for single and multiple?

```ts
addStrategy<T>(strategy: T, ...strategies?: T[])
```

chrisk-7777:
> good stuff! my only thought is maybe we should just have `addStrategy` take multiple params instead of having two methods for single and multiple?
> 
> ```ts
> addStrategy<T>(strategy: T, ...strategies?: T[])
> ```

I dig it, I'll update this evening 👍

eonarheim:
Oh I like that idea of having a variadic amount of strategies!

1 method is better than 2? Opposite of 2 heads are better than 1?

eonarheim (approved):

PR #3511, opened 2025-09-01
fix: Clean up memory leaks from Entities in Maps [#3510]

### LINKED ISSUE #3510: Memory leaks from Entities tracked in Maps
### Steps to Reproduce

1. Add and remove Actors from a scene over the course of gameplay, or in a loop. Ensure many Actors are added and removed to remove profiling noise from pooled objects. It's easiest to subclass a new Actor, to enable better grouping.
2. Open a dev tools that enables memory snapshots (eg Chrome)
3. Create a memory snapshot
4. Follow the retained objects trail to see Maps

### Expected Result

If I've added and removed 1000 Actors from a scene, I expect there to be < 50 objects retained in memory by Excalibur's systems, and for the memory usage to remain stable.

Below is an image of what I would expect (and get after a patch :) ) after 10 minutes of churning objects through automated gameplay.

Expected results of only active-ish Actors retained in memory

### Actual Result

Every actor ever created is retained for the lifetime of the World, kept in internal `Map<Entity>`, some of which are not cleared when the entity is removed. Memory usage climbs til OOMs.

Actual results of all Actors retained in-memory

### Environment

- browsers and versions: Verified Chrome (139.0.7258.155) 
- operating system: MacOS 15.6.1 (24G90)
- Excalibur versions: 30.0.3 (main branch)

### Current Workaround

I have a PR ready and am just submitting as an issue to help keep a tidy repo 🖇️

BaconSoap:

==================

Excalibur makes use of `Map<Entity, TData>` for functionality (event dispatch) and performance, using the entity as a key to event subscriptions or other data. This stores the Entity by reference as a key, making both lookup and storage fairly efficient. Thanks to Excalibur's event system, it's easy for engine developers to maintain these lists in a safe manner (paired entity added and removed event handlers); here we are patching up a few that slipped through.

Fixes open memory leaks I was experiencing in my game (which churns Actors); it now maintains a stable ~80-100mb memory usage, vs climbing to ~2gb before my poor browser gives up. Previously, these maps would grow without bounds as Actors were created and removed.

I did a quick search for more, but wanted to focus on what I could verify.

Tests passed, but I wasn't able to find a place to add new tests to automate verification and left that unchecked for now - if you have a place to start I'd be happy to look!

Thanks for the great engine ⚔️ 

Closes #3510

## Changes:

- removes entities from internal `Map<Entity>` and `Map<object>` lists when on entity removal

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,6 +116,7 @@ const query = new Query({})
 - Fixed CollidePolygonPolygon crash with some defense against invalid separation
 - Fixed issue with PostProcessor where it would not run correctly if no actors present
 - Removed warning in development for unadded entities
+- Fixed memory leaks from retained entities in `Map<Entity>`
 
 ### Updates
 
--- a/src/engine/EntityComponentSystem/EntityManager.ts
+++ b/src/engine/EntityComponentSystem/EntityManager.ts
@@ -104,10 +104,12 @@ export class EntityManager {
       const childAddedHandler = this._childAddedHandlerMap.get(entity);
       if (childAddedHandler) {
         entity.childrenAdded$.unsubscribe(childAddedHandler);
+        this._childAddedHandlerMap.delete(entity);
       }
       const childRemovedHandler = this._childRemovedHandlerMap.get(entity);
       if (childRemovedHandler) {
         entity.childrenRemoved$.unsubscribe(childRemovedHandler);
+        this._childRemovedHandlerMap.delete(entity);
       }
 
       // stats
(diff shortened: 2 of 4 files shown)


mattjennings (approved):
Oh boy, thanks a ton for tracking this one down. This looks good to me @eonarheim but perhaps we want to add a test to ensure general mass entity add/remove doesn't cause leaks. I think this can be done when running on chrome with our memory reporter...

eonarheim (approved):
@BaconSoap WOW! Bravo on tracking this one down.

Looks great!

eonarheim:
@mattjennings I'll noodle, hopefully we could come up with a way to do this during vitest somehow... I had a Jasmine/Karma plugin that did some chrome heap analysis when the flags were turned on. I wonder if we could do something similar.

This map issue is a class Erik move of forgetting to do the remove after an add🤦

PR #3531, opened 2025-10-05
Site particle docs

### LINKED ISSUE #3528: Particle Tester and docs updates
### Context

Currently, the [particle docs](https://excaliburjs.com/docs/particles/) have a few missing elements. The [Particle Tester](https://excaliburjs.com/particle-tester/) is also out of date, as mentioned in the docs. 

#### Docs

Update the following in the docs:

 - Remove `minVel` and `maxVel` from the new structure sample. It isn't valid I'm pretty sure?
 - `vel` is part of the constructor signature for an emitter, but it is not used. I believe initial velocity is derived from speed? Should we `Omit<ParticleConfig, 'vel'>` for the emitter constructor args, to make it clearer?
 - Might need to make it clear that particles ignore gravity (esp. coming from Phaser)
 - Might need to make it clear that particles ignore physics/collisions (probably assumed knowledge)
 - Is the GPU emitter mentioned in the docs? 
   - Do we have a position on when a user should reach for Emitter v GPU Emitter? 
   - Is it purely performance v compatibility? 
  
Should we drop the reference to the old structure entirely? 
Remove the term "alpha", making it current (assuming the particle tester is updated below)? 

Or do we simply include a "Previous versions" section to a page like Particles? For cases where there is a significant change. We probably want to avoid "versioning" docs (dropdown with v0.29.0, v0.30.0, v0.31.0, etc), feels maintenance heavy. 

Either way, I think the "current" should be the first thing on the page. 

#### Particle Tester:

 - Probably should be updated to latest Excalibur version
 - Update output so it produces valid 0.31.0 output
 - Update sliders to only include relevant 0.31.0 values

[Discord discussion here](https://discord.com/channels/1195771303215513671/1422414456490561577/1422414456490561577)

chrisk-7777:

==================

Closes #3528

## Changes:

- Cleans up particle documentation to reflect current particle emitter
  - nested particle structure
  - replace vel for speed
  - additional comments explaining more of the properties
  - adds embedded examples
- Playground emebed
  - More notes [here](https://github.com/excaliburjs/Excalibur/issues/3528#issuecomment-3368702814)
  - It could be useful to have an "?embed" version of the playground where it hides the header, specifically for the docs
  - Open to _not_ having this
- Added initial docs style guide
  - Open to change, its more a reference for myself - nearly every time I modify the docs I do a quick search for "now, do we refer to it as Excalibur or Excalibur.js?"


STYLEGUIDE.md
```diff
@@ -134,3 +134,15 @@ export enum CollisionType {
...
+
+#### DO
+
+ - Use "Excalibur" (vs. "excalibur", "excalibur.js", "excaliburjs") when referring to the project or the broader ecosystem
+  - e.g. "Excalibur has built in types"
+  - e.g. "The Excalibur community can be found on Discord"
+
+#### DON'T
+
+ - Avoid referencing "new" code. Instead the docs should represent the current point in time. This minimizes the docs appearing stale over time
```
chrisk-7777, line 148: Might be a hot take, push back!

Reasoning: Throughout the docs I see Excalibur referred to differently (sometimes Excalibur.js, sometimes Excalibur, etc), and sometimes on the same page. Its minor, and some may not notice, but from a branding perspective I think the consistency is good. 

It does go against the word mark, which is Excalibur.js - but its not uncommon for the word mark / logo to be a bit different.

site/docs/02-fundamentals/06-loaders.mdx
```diff
@@ -6,11 +6,9 @@ section: Fundamentals
 
 Games typically will have assets like image files, sound files, level data, etc. that you want to make available to your code. Excalibur has built in types to help you load assets. Excalibur loaders can load anything that implement the [[Loadable]] interface.
 
-To build a custom designed loader check out the [custom loaders](/docs/custom-loaders) docs.
```
chrisk-7777, line 9: This is a dead link

STYLEGUIDE.md
eonarheim: I tend to write like I speak (terrible habbit)

eonarheim:
@chrisk-7777 this is awesome!! ❤️ 

> It could be useful to have an "?embed" version of the playground where it hides the header, specifically for the docs

Totally agree, I fully support this. I want to make it "iframe-able" as cursed as that phrase sounds. We are on the same excaliburjs.com domain so we can get away with a lot.

chrisk-7777:
Thanks for the review and merge Erik. 

Funnily enough I just saw your comment on Discord about more embedded samples, so it seems like this approach is relatively aligned long term. 

If no one gets to the embed version, I can take a look in the next week.

eonarheim:
@chrisk-7777 That would be absolutely stupendous

chrisk-7777:
> @chrisk-7777 That would be absolutely stupendous

Done and raised! https://github.com/excaliburjs/excalibur-playground/pull/2

PR #3534, opened 2025-10-13
[feat] Color.Lerp

jyoung4242:
Added static lerp method to Color class AND HSLColor class.
The Color.lerp() static method essentially manages the conversion to HSL and runs the HSL lerp, and converts back.

Added test to ColorSpec.ts
Updated documentation page for Color to add Lerp section at end

### DIFF
--- a/site/docs/04-graphics/04.1-color.mdx
+++ b/site/docs/04-graphics/04.1-color.mdx
@@ -4,6 +4,12 @@ slug: /color
 section: Graphics
 ---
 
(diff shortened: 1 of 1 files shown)


jyoung4242:
The Color.lerp example code isn't showing up in the cloudfare preview pages... :|    
is that normal?

jyoung4242:
I think this one is done, .... whew

eonarheim:
SWEEEEEET

PR #3549, opened 2025-10-26
hasChild

jyoung4242:
added hasChild method


eonarheim:
QQ, should we also add a `hasGrandChild` or `hasDescendant`? that would recurse thorugh the structure

eonarheim:
Merged in b4784f0dd8505ecc525a312e00cb9e7ee0e1b799

PR #3571, opened 2025-11-12
fix: Screen orientation recalcs viewport

eonarheim:

==================

Closes #

## Changes:

- Adds a hook in the `Screen.ts` to also listen to the browser ScreenOrientation API and treat that as a "resize" event.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fixed issue where sometimes the when device orientation changed the screen would not recalculate viewport.
 - Fixed issue where the Loader could run twice even if already loaded when included in the scene loader.
 - Fixed issue where pixel ratio was accidentally doubled during load if the loader was included in the scene loader.
 - Fixed issue where Slide trasition did not work properly when DisplayMode was FitScreenAndFill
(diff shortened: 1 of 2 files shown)


eonarheim:
Doing a test locally against sum monsters, this isn't fixed. So something else is going on in that resize

eonarheim:
It seems that window.innerWidth and window.innerHeight lie to us on orientation change :O

eonarheim:
Okay the plot thickens here. It seems that chrome is doing something odd wrt to screen & innerWidth/Height on mobile browser. 

I ported this current code to sum-monsters and only got things working IF I added the meta no user scalable. This is generally considered a "no no" for accessibilty because it prevents folks from zooming. Some related discussion about this on this SO post https://stackoverflow.com/questions/76026292/why-is-window-innerheight-incorrect-until-i-tap-chrome-android but I'm currently mystified

```html
    <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
```

https://github.com/user-attachments/assets/1bf87c02-9409-4584-b73d-68fa1b196808

eonarheim:
Further investigation reveals this was a Chrome change to layout behavior

https://developer.chrome.com/blog/viewport-resize-behavior?hl=en#opting_in_to_a_different_behavior

eonarheim:
Well, this is disappointing, I think the only way to fix this is to add "user-scalable=no" to the meta viewport tag :/

PR #3624, opened 2025-12-15
docs: add share tooltip indicator for visual feedback

chrisk-7777:

Discussed in Discord. This adds a tooltip indicator when the user clicks the share button.
See video (in the video it looks like on hover, but its actually on click):

https://github.com/user-attachments/assets/736d988c-6084-4f9d-b055-58ab6b72c87c

Also fixed up pathing for the playground templates, again. Its like I've never heard of relative links before..

### DIFF
--- a/playground/src/main.ts
+++ b/playground/src/main.ts
@@ -178,6 +178,9 @@ const shareCode = async (writeToClipboard?: boolean) => {
   const url = `${window.location}?${encoded}`;
   if (writeToClipboard) {
     await navigator.clipboard.writeText(url);
+    shareButtonEl.classList.remove("copied");
+    void shareButtonEl.offsetWidth; // trigger reflow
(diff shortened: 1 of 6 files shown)


playground/src/main.ts
```diff
@@ -178,6 +178,9 @@ const shareCode = async (writeToClipboard?: boolean) => {
   const url = `${window.location}?${encoded}`;
   if (writeToClipboard) {
     await navigator.clipboard.writeText(url);
+    shareButtonEl.classList.remove("copied");
+    void shareButtonEl.offsetWidth; // trigger reflow
+    shareButtonEl.classList.add("copied");
```
chrisk-7777: I don't love the "trigger reflow" approach, but seems to be the way

Source: https://stackoverflow.com/a/45036752/614112

eonarheim:
Good catch on those relative links

PR #3695, opened 2026-02-21
fix(site): patch /api page crash from read-only Function.length

KakkoiDev:
## Summary

Hi there, I just discovered this project and I really like it! Thank you very much for making it!
[I went on the docs and... it wasn't working](https://excaliburjs.com/api/).

image

## Fix

- Fix `/api` page crash: "Cannot assign to read only property 'length' of function 'function ApiPage'"
- Root cause: Docusaurus `ComponentCreator` copies webpack CJS interop properties onto function components, hitting read-only `Function.length`
- Two patch-package fixes:
  - `@docusaurus/core`: wrap property assignment in try-catch in `ComponentCreator.js`
  - `docusaurus-plugin-typedoc-api`: make `ApiPage.length` writable before CJS export

Before | After
--- | ---
bug-api-page-crashed | fix-api-page-crashed

## Test plan
- [x] Run `npm install` in `site/` (patches apply via postinstall)
- [x] Run `npm start` in `site/`
- [x] Navigate to [http://localhost:3000/api](http://localhost:3000/api). Page should load without crashing


eonarheim:
@KakkoiDev thanks for bringing this up! I was merging some dependabot earlier and must have busted the api page

PR #3743, opened 2026-05-08
fix: Add GamepadEvents export to enhance gamepad input handling

jyoung4242:
Add GamepadEvents to the public exports in the engine's main entry point. This makes the GamepadEvents class available to consumers of the Excalibur library, allowing them to access gamepad-related event types and utilities.

### DIFF
--- a/src/engine/index.ts
+++ b/src/engine/index.ts
@@ -83,7 +83,7 @@ export type {
   NavigatorGamepadEvent,
   GamepadConfiguration
 } from './input/gamepad';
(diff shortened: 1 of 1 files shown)


eonarheim (approved):

Nimgoble:
@eonarheim If its exporting GamePadEvents now, should the code now also export the other input events(KeyEvents, PointerEvents)?  Or should it go back to not exporting GamePadEvents and add your suggested fix to https://github.com/excaliburjs/Excalibur/blob/main/src/engine/input/gamepad.ts#L5 ?

PR #3747, opened 2026-05-17
fix: Enhance pointer event handling by adding coordPlane check for graphics tracking, also added world to screen space conversion on pointer event check

jyoung4242:
Fixed pointer event handling for screen-space graphics by ensuring only screen-space graphics are tracked in the pointer system and converting world coordinates to screen coordinates before querying the graphics hash grid.

What changed:
- Captured coordPlane from the entity transform during pointer registration.
- Added a guard so only graphics with CoordPlane.Screen are tracked by _graphicsHashGrid.
- Converted pointer event world position to screen space with _engine.worldToScreenCoordinates(...) before querying _graphicsHashGrid.

Files changed
pointer-system.ts

### DIFF
--- a/src/engine/input/pointer-system.ts
+++ b/src/engine/input/pointer-system.ts
@@ -54,7 +54,7 @@ export class PointerSystem extends System {
       const maybeGfx = e.get(GraphicsComponent);
       if (maybeGfx) {
         this._graphics.push(maybeGfx);
-        this._graphicsHashGrid.track(maybeGfx); // disambiguate screen and world space
+        this._graphicsHashGrid.track(maybeGfx);
       }
       this._sortedTransforms.push(tx);
       this._sortedEntities.push(tx.owner);
--- a/src/engine/input/pointer-system.ts
+++ b/src/engine/input/pointer-system.ts
@@ -37,7 +37,6 @@ export class PointerSystem extends System {
 
     this.query.entityAdded$.subscribe((e) => {
       const tx = e.get(TransformComponent);
-      const coordPlane = tx.coordPlane;
       const pointer = e.get(PointerComponent);
       this._pointerEventDispatcher.addObject(
         e,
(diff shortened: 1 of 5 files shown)


jyoung4242:
this is the tricky issue i've been hunting down all weekend.   The core failure mode is that ScreenElements are 'drawn' in screenspace, but their colliders remain in worldspace.  This results in pointer events no longer aligning with the drawn entities on the screen if the camera pans or zooms.

I'm worried about unintended side effects of my fix.  My line of thinking is that if you're using 'graphics' or screenspace items, then and only then should it be in the hashgrid query, then i can holistically convert the query check against the screen coords, not the world coords.  It fixed the bug in my testing, but maybe something else is dependent on it that i'm not seeing.

src/engine/input/pointer-system.ts
```diff
@@ -52,7 +53,8 @@ export class PointerSystem extends System {
       );
       this._entityToPointer.set(e, pointer);
       const maybeGfx = e.get(GraphicsComponent);
-      if (maybeGfx) {
+      const isInScreenSpace = coordPlane === CoordPlane.Screen;
+      if (maybeGfx && isInScreenSpace) {
```
eonarheim: I think we want to still track all graphics, non-screen space graphics wouldn't be tracked in this case. 

Let me noodle here, I think there is a way to make this work for all.

src/engine/input/pointer-system.ts
```diff
@@ -142,7 +144,7 @@ export class PointerSystem extends System {
         }
       }
 
-      const graphics = this._graphicsHashGrid.query(pos.worldPos);
+      const graphics = this._graphicsHashGrid.query(this._engine.worldToScreenCoordinates(pos.worldPos));
```
eonarheim:
```suggestion
      const graphics = this._graphicsHashGrid.query(pos.screenPos);
```

PR #3787, opened 2026-06-19
feat: update playground layout and add console output

chrisk-7777:
There isn't an issue for this one, but discussed in Discord. 

## Changes

This PR overhauls the playground to:

 * Have a stronger webapp look and feel, similar to codepen, etc
 * Display logs within the playground
 * Introduce a framework to minimise subtle state bugs

## Tech choice
As discussed in Discord, Erik mentioned using what makes me "happy and productive" - which is React. Even for non-React devs, I hope it is relatively easy to trace the data flow - there is no over the top state management like redux, etc. 

## UI 

The look and feel is heavily inspired by @jyoung4242 IDE work. 

The zinc and purple color scales are from Tailwind v4

### Editor

CleanShot 2026-06-19 at 21 38 28@2x

### Assets

I'm not 100% on this; it's better than current (which is non-existent) but I will do another pass on this later.

CleanShot 2026-06-19 at 21 38 59@2x

### Console

Supports 3x colors via `console.log`, `console.info`, and `console.warn`. 

CleanShot 2026-06-19 at 21 40 01@2x

## Embedded

The same embedded search params are applied. It will look like the following when embedded in the docs:

CleanShot 2026-06-17 at 23 08 27@2x

---

## Deployments

The following `.env` will need to part of the deployment of the playground:

```properties
VITE_PLAYGROUND_PATH='./playground'
VITE_HIJACK_LOGS=true
```

**Note:** I couldn't find the actual deployment steps in `.github` for the playground. And I don't think its built during a PR?

---

## Basic accessibility

The monaco editor traps focus (possibly a future fix?), but all other elements have focus state and the template modal has proper focus trap (thanks to baseui)

### Toolbar button focus
CleanShot 2026-06-19 at 21 48 28@2x

### Toolbar switch focus
CleanShot 2026-06-19 at 21 48 38@2x

### Modal button focus (3rd button)
CleanShot 2026-06-19 at 21 48 53@2x

---

## Misc

 * I've kept light mode, but it's not great - if anyone has a good light theme we could slot in that matches the overall feel let me know
 * The dark theme is based on mookie's ide editor
 * There are more `useEffect`'s then I would normally want, however, a lot of this is imperative in nature

---

## Disclaimer

No AI clankers used for this PR _except_ for converting the dark mode json to the light mode json.


playground/src/components/button/button.module.css
```diff
@@ -0,0 +1,49 @@
...
+  --foreground-color: var(--color-zinc-100);
+  --border-color: var(--color-zinc-700);
+
+  :global(html[data-theme='light']) & {
+    --background-color: transparent;
+    --background-color-hover: var(--color-zinc-200);
+    --foreground-color: var(--color-zinc-900);
+    --border-color: var(--color-zinc-600);
+  }
+}
```
chrisk-7777, line 20: Sidenote: I've been really digging this css variable pattern recently, especially when there are a bunch of variants like with buttons. 

I feel like it gives a clear matrix of colors at the top and then the rest of the file becomes super shallow.

playground/src/components/editor/editor.tsx
chrisk-7777, line 1: This is probably the most complex component of the lot. Lots of imperative juggling to play nice between the code, the compiled code, and the editor.

The nice thing though is we no longer need the retry seen here: https://github.com/excaliburjs/Excalibur/blob/main/playground/src/main.ts#L137

Ultimately I think that was just a race condition, and by splitting things up, its easier to control the flow

playground/src/templates/default.ts
```diff
@@ -6,7 +6,8 @@ const game = new ex.Engine({
     canvasElementId: 'preview-canvas',
     displayMode: ex.DisplayMode.FitContainer,
     width: 600,
-    height: 400
+    height: 400,
+    suppressConsoleBootMessage: true,
```
chrisk-7777, line 10: I've disabled the console log because the log ui can't really handle the formatted "Powered by"

Image

Am I cheating by suppressing the console boot message? Yes
Am I prepared to write a console that can handle formatted message right now? No 😅

playground/src/components/header/header.tsx
```diff
@@ -0,0 +1,65 @@
+import { Button } from '../button/button';
+import { Moon } from '../icons/moon';
```
eonarheim, line 2: @chrisk-7777 I think we're missing these icons

playground/src/components/button/button.module.css
eonarheim: I'm a fan!

playground/src/components/header/header.tsx
chrisk-7777: Oh! The root .gitignore `Icon?` is hiding the icons here 🤦 

Will fix shortly

chrisk-7777:
Icons fixed! 

I tweaked the asset panel slightly, and also added a "credit" link? Does that gel ok? Feels like the right approach for the assets from Kenney

CleanShot 2026-06-20 at 22 33 05@2x

My only outstanding question / note is about deployment:

I couldn't see where the site/playground is deployed within the github actions. I'll need to add a few lines to write the `.env` file. Could you point me in the right direction?

eonarheim:
> My only outstanding question / note is about deployment:

> I couldn't see where the site/playground is deployed within the github actions. I'll need to add a few lines to write the .env file. Could you point me in the right direction?
 
@chrisk-7777  Uff da! I don't remember myself, let me poke around I think it's hidden in our cloudflare build. I should be able to tweak that if you don't already have access

eonarheim:
Also I love the credit link 👍

eonarheim (approved):

eonarheim:
In theory this is right, I think I understand where they're used so I can debug as necessary
image

playground/src/components/editor/editor.tsx
jyoung4242: is this something i can drop into a lit component?

chrisk-7777: 👍 discussing in discord

